PR #18, opened 2022-11-11
Fix/login page header footer layout

nmialik:


src/App.tsx
```diff
@@ -17,6 +18,7 @@ function App() {
           <BrowserRouter>
             <AppRoutes />
           </BrowserRouter>
+          <Footer />
```
IhorNesterenko99, line 21: In dashboard it will be overlapped by NavPanel. Maybe it's better to insert it separately for auth and base?

src/components/Icon/France.tsx
```diff
@@ -0,0 +1,33 @@
+export const France = ({ width = 32, height = 24 }): JSX.Element => (
```
IhorNesterenko99, line 1: Maybe it will be better to also add SVGProps<SVGElement> TS props definition as in other svgs?

src/components/Language/Language.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  border-radius: 50%;
+  margin-right: ${theme.spacing(0.4)};
+
+  svg {
+    transform: translate(-0.4rem);
+  }
+`;
+
+export const StyledLabel = styled(StyledSmallText)`
+  color: ${variables.palette.shades80};
```
IhorNesterenko99, line 28: You could change color through color property of StyledSmallText component.

PR #31, opened 2022-11-18
feature/CMIML-203-applet-table-api

kdanilenka32:


src/components/AppletsTable/AppletsTable.tsx
```diff
@@ -1,20 +1,48 @@
...
+        },
+      };
+    });
+
+    formattedApplets && setRows(formattedApplets);
+  }, [accData?.account?.applets]);
 
   const handleSearch = (value: string) => {
-    const filteredRows = [...rowsCells].filter((row) =>
+    const filteredRows = rows?.filter((row) =>
```
nmialik, line 42: In the case of search, probably we should search not by "rows", but by some kind of global array with applets, which contains all elements, and not just those filtered.

.eslintrc
```diff
@@ -1,53 +1,29 @@
...
-            "parent",
-            "sibling"
-          ]
-        ]
+        "groups": [["builtin", "external"], ["internal"], ["index", "parent", "sibling"]]
       }
     ],
     "import/no-cycle": "error",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
```
IhorNesterenko99, line 50: Maybe its better to leave this warning and to write eslint-ignore exhaustive deps if it is needed?

src/components/AppletsTable/AppletsTable.tsx
kdanilenka32: fixed

IhorNesterenko99 (approved):

PR #35, opened 2022-11-23
Feature/CMIML-208 create respondents table and connect to the api

IhorNesterenko99:


src/components/RespondentsTable/RespondentsTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import { RespondentsTableHeader } from './RespondentsTable.styles';
+import { headCells } from './RespondentsTable.const';
+
+export const RespondentsTable = (): JSX.Element => {
+  const dispatch = useAppDispatch();
+  const { t } = useTranslation('app');
+  const timeAgo = useTimeAgo();
+  const usersData = users.useUserData();
+  const [searchValue, setSearchValue] = useState('');
```
kdanilenka32, line 22: Let move this search logic to custom hook

IhorNesterenko99: Ok, in separate task

PR #39, opened 2022-11-25
Feature/folders applets api

nmialik:


src/redux/modules/Account/Account.thunk.ts
```diff
@@ -26,3 +33,33 @@ export const updateAlertStatus = createAsyncThunk(
...
 );
+
+export const getAppletsForFolders = createAsyncThunk(
+  'account/getAppletsInFolders',
+  async ({ account }: { account: Account }, { dispatch }) => {
+    const foldersData = await Promise.allSettled(
+      account.folders.map(async (folder) => await dispatch(getAppletsForFolder({ folder }))),
+    );
+
+    const result = [
```
IhorNesterenko99, line 44: Could you please make return here?

src/redux/modules/Account/Account.utils.ts
```diff
@@ -0,0 +1,100 @@
+import { Folder } from 'redux/modules';
+
+export type AppletResponse = {
```
IhorNesterenko99, line 3: Pls move types to the Account.schema.ts or to the Account.types.ts

src/redux/modules/Account/Account.utils.ts
```diff
@@ -0,0 +1,100 @@
...
+        isExpanded: false,
+        isFolder: false,
+        isChild: true,
+      };
+    });
+
+    loadedFolder.items = items;
+    removeDuplicateApplets(loadedFolder);
+  }
+  return loadedFolder;
```
IhorNesterenko99, line 99: Pls add empty line before return statement

src/redux/modules/Account/Account.reducer.ts
```diff
@@ -31,4 +31,34 @@ export const extraReducers = (builder: ActionReducerMapBuilder<AccountSchema>):
       switchAccount.error = error.response?.data as AxiosError<ErrorResponse>;
     }
   });
+
+  builder.addCase(getAppletsForFolders.pending, ({ accountFoldersApplets }, action) => {
+    if (accountFoldersApplets.status !== 'loading') {
```
kdanilenka32, line 36: Would be great to move  these functions, we have similar functions in 2 places

src/redux/modules/Account/Account.thunk.ts
IhorNesterenko99: I mean without result variable: return [...

IhorNesterenko99 (approved):

PR #41, opened 2022-11-28
feature/CMIML-216-actions-btns

kdanilenka32:


src/components/AppletsTable/Table/AppletItem/AppletItem.const.tsx
```diff
@@ -0,0 +1,33 @@
+import { Svg } from 'components/Svg';
+import { Row } from 'components/AppletsTable/Table/Table.types';
```
IhorNesterenko99, line 2: Could you pls use relative path here (for Row type)?

src/components/AppletsTable/Table/AppletItem/AppletItem.tsx
```diff
@@ -2,11 +2,18 @@ import { TableCell, TableRow } from '@mui/material';
 
 import { Row } from 'components/AppletsTable/Table/Table.types';
 import { AppletImage } from 'components/AppletsTable/Table/AppletImage';
```
IhorNesterenko99, line 4: Could you pls use relative paths here as well?

src/components/AppletsTable/Table/AppletItem/AppletItem.styles.ts
```diff
@@ -1,6 +1,49 @@
...
+  right: 4rem;
+  transform: translateY(-50%);
+`;
+
+export const StyledActionButton = styled(Button)`
+  width: 4rem;
+  height: 4rem;
+  min-width: 4rem;
+  padding: 0;
+  border-radius: 100%;
```
nmialik, line 35: Maybe it makes sense to move it to variables, since `border-radius: 100%;` already occurs several times in the code?

nmialik (approved):

PR #44, opened 2022-11-29
Feature/ML-172 Create managers table for the exact applet and connect to the API

IhorNesterenko99:
…d to the API


src/components/RespondentsTable/RespondentsTable.utils.ts
```diff
@@ -1,23 +1,21 @@
 import { UserData, UsersItems } from 'redux/modules';
 
-export const prepareAppletUsersData = (id: string, items?: UsersItems['items']) =>
-  items &&
-  items.reduce((acc: UserData[] | null, currentValue) => {
-    if (currentValue[id] && acc) {
-      return acc.concat(currentValue[id]);
-    }
+export const prepareUsersData = (items?: UsersItems['items'], id?: string) =>
+  id && items
```
kdanilenka32, line 4: is this different from  prepareManagersData?

IhorNesterenko99: it has different types

kdanilenka32 (approved):

PR #60, opened 2022-12-06
Feature/add remove renaming folder logic  connect to api

nmialik:


src/redux/modules/Folders/Folders.reducer.ts
```diff
@@ -0,0 +1,102 @@
...
+      foldersApplets.requestId = initialState.foldersApplets.requestId;
+      foldersApplets.status = 'success';
+      state.flattenFoldersApplets = action.payload;
+    }
+  });
+
+  builder.addCase(updateFolder.pending, (state, action) => {
+    const { foldersApplets } = state;
+    if (foldersApplets.status !== 'loading') {
+      foldersApplets.requestId = action.meta.requestId;
```
kdanilenka32, line 89: function duplication, pls move these functions

IhorNesterenko99 (approved):

PR #67, opened 2022-12-12
feat: ML-223 - Refactored typography, added state colors to buttons, …

IhorNesterenko99:
…added select dropdown styles, and set active notification view according to the new design. Fixed console warnings.

### DIFF
--- a/src/components/TopBar/Notifications/Notification/Notification.styles.ts
+++ b/src/components/TopBar/Notifications/Notification/Notification.styles.ts
@@ -9,6 +9,7 @@ import {
   StyledTitleMedium,
 } from 'styles/styledComponents/Typography';
 import { StyledFlexTopCenter } from 'styles/styledComponents/Flex';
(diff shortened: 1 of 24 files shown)


src/components/TopBar/Notifications/Notification/Notification.styles.ts
```diff
@@ -9,28 +9,27 @@ import {
...
-  border-radius: ${variables.borderRadius.lg};
-  text-align: left;
+export const StyledNotification = styled(Box, shouldForwardProp)`
+  padding: ${theme.spacing(1.6)};
   display: flex;
   flex-wrap: wrap;
   position: relative;
+  cursor: pointer;
+  border-bottom: ${variables.borderWidth.md} solid ${variables.palette.surface_variant};
+  transition: background-color 0.3s;
```
wildkiomi, line 27: maybe it's better to create variables for animations?

PR #76, opened 2022-12-19
ML2-246: Toggle button group component

nmialik:


src/svgSprite.ts
```diff
@@ -6,6 +6,10 @@ export const svgSprite = (): string => `
...
   </symbol>
 
+  <symbol viewBox="0 0 19 18" id="svg-check">
```
IhorNesterenko99, line 9: Why don't you use 24 / 24 view box and dimensions here?
And I've already added the svg-check component in my PR. I think they have to be the same - from the ML1 Design system.
image

nmialik: because there was an icon with only such a container in Figma
Снимок экрана 2022-12-20 в 13 08 45
but ok, we can use your icon

IhorNesterenko99: pls take a look here
in this place where are all icons
https://www.figma.com/file/LXl9BAJSwtgWuzkxqOqvX8/ML1-Design-System?node-id=53015%3A36793&t=YK8qavNGtWL28Gjw-0

IhorNesterenko99 (approved):

PR #85, opened 2022-12-29
M2-306 Share to library setting

wildkiomi:


src/components/AppletSettings/AppletSettings.styles.ts
```diff
@@ -1,10 +1,15 @@
 import { styled } from '@mui/system';
```
kdanilenka32, line 1: pls move AppletSettings to features

kdanilenka32 (approved):

PR #99, opened 2023-01-12
Feature/m2 351 edit access modal

kdanilenka32:


src/resources/roles.ts
```diff
@@ -0,0 +1,7 @@
+export const Roles = {
```
IhorNesterenko99, line 1: Let's move this variable to utils/constants.ts

src/styles/styledComponents/Typography.tsx
```diff
@@ -112,6 +112,10 @@ export const StyledBodyMedium = styled(Typography)`
   letter-spacing: ${variables.letterSpacing.lg};
 `;
 
+export const StyledBodyMediumMedium = styled(StyledBodyMedium)`
```
wildkiomi, line 115: Please, remove. In design there is a mistake (Andrei will fix it). For Atkinson font there is only two types of font-weight: 400 and 700. You saw an old design with noto-sans font and old font-weight.

src/components/Popups/EditAccessPopup/Applet/Applet.const.ts
```diff
@@ -0,0 +1,29 @@
+import { MenuItem } from 'components/Menu/Menu.types';
+import { Roles } from 'resources';
+
+export const getMenuItems = (addRole: (title: string) => void): MenuItem[] => [
+  {
+    title: Roles.editor,
+    action: (title?: string) => {
```
wildkiomi, line 7: maybe you can create a const for action and use it for every item?

src/svgSprite.ts
```diff
@@ -542,5 +542,17 @@ export const svgSprite = (): string => `
       </clipPath>
     </defs>
   </symbol>
+
+  <symbol viewBox="0 0 18 18" id="svg-editor">
```
IhorNesterenko99, line 546: Could you use '0 0 24 24' box for all svgs (to make all the icons have the standard view box)?

src/styles/variables/font.ts
```diff
@@ -10,6 +10,7 @@ export const font = {
   },
   weight: {
     bold: 700,
+    medium: 500,
```
wildkiomi, line 13: also remove pls

src/components/Popups/EditAccessPopup/Applet/Applet.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+import { StyledFlexTopCenter } from 'styles/styledComponents/Flex';
+import { variables } from 'styles/variables';
+
+export const StyledApplet = styled(Box)`
+  padding: ${theme.spacing(1.6)};
+  border-radius: ${variables.borderRadius.lg2};
+  margin-bottom: ${theme.spacing(1.2)};
+  background-color: ${variables.palette.surface5};
+
+  .MuiChip-root > svg {
```
nmialik, line 14: maybe it makes sense to move these styles to the Chip component styles

src/components/Popups/EditAccessPopup/Applet/Applet.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+    });
+    setAnchorEl(null);
+  };
+
+  return (
+    <StyledApplet>
+      <StyledRow>
+        <StyledFlexTopCenter>
+          {img && <img src={img} alt={title} />}
```
nmialik, line 48: any styles for the image? what if the image will be larger than the mocked one? - I suppose it should have a fixed width and height at least

src/svgSprite.ts
nmialik: https://www.figma.com/file/LXl9BAJSwtgWuzkxqOqvX8/ML1-Design-System?node-id=53015%3A36793&t=SAqJtQPlTsvKXOIb-0

nmialik (approved):

src/consts.ts
```diff
@@ -12,3 +12,11 @@ export const APPLET_PAGES = {
 };
```
wildkiomi, line 13: I think, you shouldn't rename and move this file

kdanilenka32: utils is not the right place for constants

kdanilenka32: and I renamed because with constants some problem with import when constants.ts

wildkiomi (approved):

PR #102, opened 2023-01-12
AppletsSmallTable component

nmialik:


kdanilenka32 (approved):

wildkiomi (approved):

src/components/Tables/AppletsSmallTable/AppletsSmallTable.const.ts
```diff
@@ -2,7 +2,7 @@ import { t } from 'i18next';
 
 import { HeadCell } from 'types/table';
 
-export const headCells: HeadCell[] = [
+export const getHeadCells = (): HeadCell[] => [
```
IhorNesterenko99, line 5: Do we need a function here instead of variable?

PR #109, opened 2023-01-18
Feature/M2-398 calendar export popup

IhorNesterenko99:
image


src/features/Applet/Schedule/ExportSchedulePopup/ExportSchedulePopup.const.tsx
```diff
@@ -0,0 +1,34 @@
+import { t } from 'i18next';
```
kdanilenka32, line 1: better use import from i18n.ts

IhorNesterenko99: It seems it's not working

kdanilenka32 (approved):

wildkiomi (approved):

PR #168, opened 2023-03-01
M2-743: Library - Remove applet popup

nmialik:


src/resources/app-fr.json
```diff
@@ -310,6 +310,9 @@
   "applet_other": "Applets",
   "viewDetails": "Afficher les détails",
   "addToCart": "Ajouter au panier",
+  "removeApplet": "Remove applet",
```
IhorNesterenko99, line 313: maybe it's better to translate this for now to keep this file in one state?

IhorNesterenko99 (approved):

PR #194, opened 2023-03-15
feature/M2-852-delete-transfer-applet

kdanilenka32:


src/modules/Dashboard/state/Applets/Applets.reducer.ts
```diff
@@ -7,6 +7,14 @@ import { AppletsSchema } from './Applets.schema';
 import { getApplets } from './Applets.thunk';
 import { state as initialState } from './Applets.state';
 
+export const reducers = {
+  deleteApplet: ({ applets }: AppletsSchema, action: PayloadAction<{ id: string }>): void => {
+    if (applets?.data) {
```
alex-i-e, line 12: please use early return, it's a good practice to avoid nesting

src/modules/Dashboard/features/Applet/Popups/TransferOwnershipPopup/TransferOwnershipPopup.tsx
```diff
@@ -11,8 +11,8 @@ export const TransferOwnershipPopup = () => {
   const { t } = useTranslation('app');
   const dispatch = useAppDispatch();
   const { transferOwnershipPopupVisible, appletId } = popups.useData();
-  const accountData = account.useData();
-  const applet = accountData?.account?.applets?.find((el) => el.id === appletId);
+  const appletstData = applets.useData();
```
muhurov, line 14: looks like there is excess symbol 't'

src/modules/Dashboard/features/Applets/Applets.tsx
```diff
@@ -40,7 +40,7 @@ export const Applets = () => {
       if (ownerId) {
         const ordering = `${order === 'asc' ? '+' : '-'}${orderBy}`;
         const { getApplets } = applets.thunk;
-        const result = await dispatch(
+        await dispatch(
```
IhorNesterenko99, line 43: I think we can remove async await from here

src/modules/Dashboard/features/Applets/Table/Table.types.ts
```diff
@@ -1,13 +1,13 @@
...
-import { FolderApplet } from 'redux/modules';
+import { Applet, FolderApplet } from 'redux/modules';
 import { HeadCell, Order } from 'shared/types/table';
 
 import { OrderBy } from '../Applets.types';
 
 export type TableProps = {
   columns: HeadCell[];
-  rows: FolderApplet[] | null;
+  rows: FolderApplet[] | Applet[] | null | undefined;
```
IhorNesterenko99, line 10: could we use rows?: FolderApplet[] | Applet[] | null;

src/modules/Dashboard/state/Applets/Applets.reducer.ts
kdanilenka32: no need here))

PR #201, opened 2023-03-19
reduce circular dependencies

nmialik:


src/shared/styles/styledComponents/IconButton.ts
```diff
@@ -1,6 +1,7 @@
 import { styled } from '@mui/material';
 
-import { variables } from '../variables';
+import { variables } from 'shared/styles/variables';
```
IhorNesterenko99, line 3: could we use just shared/styles?

src/shared/styles/styledComponents/Library.ts
```diff
@@ -3,7 +3,8 @@ import { Box } from '@mui/material';
 
 import theme from 'shared/styles/theme';
```
IhorNesterenko99, line 4: could we use just shared/styles for theme and variables?

src/shared/styles/styledComponents/QuantityCircle.ts
```diff
@@ -1,6 +1,7 @@
 import { styled } from '@mui/system';
 
 import { variables } from 'shared/styles/variables';
```
IhorNesterenko99, line 3: also here from shared/styles

kdanilenka32 (approved):

PR #259, opened 2023-04-07
M2-1181 - Builder - add validation errors and according popups on Save Changes flow on applet creation/edition

muhurov:


src/modules/Builder/features/Activities/Activities.styles.ts
```diff
@@ -0,0 +1,10 @@
+import { styled } from '@mui/material';
+
+import { StyledBuilderBtn } from 'shared/styles';
+
+export const StyledAddActivityBtn = styled(StyledBuilderBtn)`
+  position: absolute;
+  top: 2.4rem;
+  right: 6.4rem;
+  z-index: 1100;
```
IhorNesterenko99, line 9: pls use z-index variable from MUI

src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
```diff
@@ -78,12 +77,25 @@ export const BuilderApplet = () => {
     [],
   );
 
+  const { errors } = useFormState({
+    control: methods.control,
```
IhorNesterenko99, line 81: why not to extract all variables from methods?

muhurov: useFormState accepts only control and name (and another 2 flags don't extracted from methods returned from useForm())

src/modules/Builder/features/Activities/Activities.styles.ts
muhurov: done (button was moved to the ActivitiesHeader, excess styles were removed)

src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
IhorNesterenko99: for example you can do like this: 
const { control, watch, setValue, getValues, register, unregister, clearErrors, reset } = methods;

IhorNesterenko99 (approved):

src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
muhurov: i missed your point, fixed

src/modules/Builder/features/Activities/Activities.tsx
```diff
@@ -12,17 +12,18 @@ import { getNewActivity } from 'modules/Builder/pages/BuilderApplet/BuilderApple
...
+import { getActions } from './Activities.const';
 
 export const Activities = () => {
   const { t } = useTranslation('app');
-  const { control, watch } = useFormContext();
+  const { control, watch, getFieldState } = useFormContext();
   const navigate = useNavigate();
   const { appletId } = useParams();
-  const [deletionModalVisible, setDeletionModalVisible] = useState(false);
+  const [activityToDelete, setActivityToDelete] = useState<string | undefined>('');
```
alex-i-e, line 23: why did you use `undefined` ?
```suggestion
  const [activityToDelete, setActivityToDelete] = useState<string>('');
```

muhurov: fixed

src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.tsx
```diff
@@ -1,37 +1,18 @@
...
 
-  const handleAddActivity = () => {
-    const newActivity = getNewActivity();
-
-    appendActivity(newActivity);
-    navigate(
-      generatePath(page.builderAppletActivityAbout, { appletId, activityId: newActivity.key }),
-    );
-  };
+export const ActivitiesHeader = ({ isSticky, children, headerProps }: ActivitiesHeaderProps) => {
```
alex-i-e, line 9: please avoid usage the headerProps as object with single property `onAddActivity`

src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
```diff
@@ -202,6 +203,12 @@ export const useSaveAndPublishSetup = (hasPrompt: boolean) => {
       return;
     }
 
+    const isValid = await trigger();
+
+    if (isValid) {
```
alex-i-e, line 208: why did add the password popup on form validity? please check `sendRequestWithPasswordCheck` we will do it there later AND only in case we have no password in secrurity storage. Per my understanding we should add popup with errors in case of form invalidate : https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/169476176/AUS-7.1.0+-+Save+Publish
image
image
image

alex-i-e (approved):

PR #276, opened 2023-04-17
Feat: M2-1378 - Builder - Duplicate Activity, Duplicate Item (same as Duplicate Activity Flow)

muhurov:


alex-i-e (approved):

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -13,21 +14,50 @@ import {
...
+  id: undefined,
+  key: uuidv4(),
+  ...((item?.responseType === ItemResponseType.SingleSelection ||
+    item?.responseType === ItemResponseType.MultipleSelection) && {
+    responseValues: {
+      ...item.responseValues,
+      options: (item.responseValues as SingleAndMultipleSelectItemResponseValues)?.options?.map(
+        (option: SingleAndMultipleSelectionOption) => ({
+          ...option,
+          id: uuidv4(),
```
IhorNesterenko99, line 39: How do you think, maybe it's better to use key for new option?
Because on update we need to send id in payload.
image

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -13,21 +14,50 @@ import {
...
+        }),
+      ),
+    },
+  }),
+  ...(item?.responseType === ItemResponseType.Slider && {
+    responseValues: {
+      ...item.responseValues,
+      id: uuidv4(),
```
IhorNesterenko99, line 47: The same here about id and key

IhorNesterenko99 (approved):

PR #316, opened 2023-04-25
M2-1579: Dataviz - User nickname is shown instead of the First Name and Last Name Fix

nmialik:


muhurov (approved):

IhorNesterenko99 (commented):
pls add version increment for fix in package.json

PR #332, opened 2023-04-27
Fix/m2 1593 error while flow duplication

alex-i-e:


src/modules/Builder/features/ActivityFlow/ActivityFlow.utils.tsx
```diff
@@ -32,3 +34,20 @@ export const getFlowsItemActions = ({
...
 
 export const getActivityFlowKey = (flow: ActivityFlowFormValues) => flow.id ?? flow.key ?? '';
+
+export const getDuplicatedActivityFlow = (flow: ActivityFlowFormValues, name: string) => {
+  const duplicatedItems =
+    flow.items?.map((item) => ({
+      ...item,
+      id: undefined,
+      key: uuidv4(),
+    })) ?? [];
```
kdanilenka32, line 44: Why  ?? not ||

IhorNesterenko99 (approved):

muhurov (approved):

src/modules/Builder/features/ActivityFlow/ActivityFlow.utils.tsx
alex-i-e: `??` is more strict for check. That why I point that only `undefined` or `null` can be there and I used `[]` as fallback? In case of `||` I suppose that result can be any of values: 0, ``, null, undefined, but I didn't use it because it cannot be '' or 0

PR #336, opened 2023-04-28
Fix: M2-1639 - Builder - Breadcrumbs: Navigation is not possible after clicking on activity name or flow name

muhurov:


src/shared/hooks/useBreadcrumbs.tsx
```diff
@@ -23,39 +29,55 @@ export const useBreadcrumbs = (restCrumbs?: Breadcrumb[]) => {
   const { result: appletData } = applet.useAppletData() ?? {};
   const isNewApplet = useCheckIfNewApplet();
   const appletLabel = (isNewApplet ? t('newApplet') : appletData?.displayName) ?? '';
+  const activityLabel =
+    appletData?.activities?.find((activity) => getEntityKey(activity) === activityId)?.name ??
```
alex-i-e, line 33: could you check the case when we create an activity in local storage but not save it? My concern is we cannot rely on appletData here and couple of lines below

IhorNesterenko99 (approved):

alex-i-e (approved):

PR #363, opened 2023-05-03
Feat/M2-1699 Notifications reminder validation

IhorNesterenko99:


src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.utils.tsx
```diff
@@ -14,19 +16,131 @@ import {
...
+    content: <NotificationsTab />,
+    hasError: hasNotificationsErrors,
+  },
+];
+
+const getStartEndComparisonResult = (startTime: string, endTime: string) => {
+  const startDate = new Date(`2000-01-01T${startTime}:00`);
+  const endDate = new Date(`2000-01-01T${endTime}:00`);
+
+  return startDate < endDate;
```
muhurov, line 66: are you sure about format 'hh:mm'? if '1:30' is being compared instead of '01:30', so '1:30' < '12:30' === false

IhorNesterenko99: yes, the value has additional zero before the number

alex-i-e (approved):

PR #366, opened 2023-05-03
Fix/m2 1467 fix description for progress popup

alex-i-e:


muhurov (approved):

src/shared/state/Applet/Applet.thunk.ts
```diff
@@ -12,6 +12,11 @@ import {
   getAppletWithItemsApi,
 } from 'api';
 
+export const enum AppletThunkTypePrefix {
+  create = 'applet/createApplet',
```
IhorNesterenko99, line 16: pls use capitalized Create, Update

PR #368, opened 2023-05-04
M2-1697 Remove access api

wildkiomi:


muhurov (approved):

src/modules/Dashboard/api/api.ts
```diff
@@ -154,6 +156,34 @@ export const revokeAppletUserApi = (
     signal,
   });
 
+export const removeManagerAccess = (
```
IhorNesterenko99, line 159: Could you pls also delete unnecessary old api functions and types regarding access if they exist?

IhorNesterenko99 (approved):

PR #404, opened 2023-05-12
M2-1656-1733 activity and item bugs

wildkiomi:


IhorNesterenko99 (approved):

alex-i-e (approved):

muhurov (approved):

src/modules/Builder/components/Item/Item.tsx
```diff
@@ -64,7 +64,7 @@ export const Item = ({
             <StyledActivityDescription>{description}</StyledActivityDescription>
           </>
         )}
-        {count && (
+        {count !== undefined && (
```
alex-i-e, line 67: is this check really need? Per my understanding, count is a number and we should always show the next component, right?

PR #408, opened 2023-05-12
[Awaiting for M2-1843, M2-1846, M2-1844] Feature/m2 1782 add decryption logic for dataviz responses

alex-i-e:


src/shared/utils/encryption/encryption.ts
```diff
@@ -61,3 +71,43 @@ export const getAppletEncryptionInfo = ({
...
+  const secretKey = key.computeSecret(Buffer.from(userPublicKey));
+
+  return crypto.createHash('sha256').update(secretKey).digest();
+};
+
+export const decryptData = ({ text, key }: { text: string; key: number[] }) => {
+  const textParts = text.split(':');
+  const iv = Buffer.from(textParts.shift()!, 'hex');
+  const encryptedText = Buffer.from(textParts.join(':'), 'hex');
+  const decipher = crypto.createDecipheriv('aes-256-cbc', Buffer.from(key), iv);
```
nmialik, line 93: maybe it makes sense to move the algorithm('aes-256-cbc') to a const variable?

nmialik (approved):

src/shared/utils/encryption/encryption.ts
```diff
@@ -61,3 +71,43 @@ export const getAppletEncryptionInfo = ({
...
+  key.setPrivateKey(Buffer.from(appletPrivateKey));
+
+  const secretKey = key.computeSecret(Buffer.from(userPublicKey));
+
+  return crypto.createHash('sha256').update(secretKey).digest();
+};
+
+export const decryptData = ({ text, key }: { text: string; key: number[] }) => {
+  const textParts = text.split(':');
+  const iv = Buffer.from(textParts.shift()!, 'hex');
```
muhurov, line 91: please avoid using such short variable names

alex-i-e: I coped/pasted ecrypt/decrypt function from previous application as is, but it makes sense to update namings

muhurov (approved):

PR #409, opened 2023-05-13
Fix/m2 1540 set encryption after transfer ownership

alex-i-e:


src/modules/Dashboard/features/Applets/Table/AppletItem/AppletItem.tsx
```diff
@@ -49,9 +55,7 @@ export const AppletItem = ({ item }: { item: FolderApplet }) => {
   };
 
   const checkAppletEncryption = (callback: () => void) =>
-    item?.roles?.includes('owner') && !item?.encryption
-      ? setPasswordPopupVisible(true)
-      : callback();
+    item.role?.includes('owner') && !item.encryption ? setPasswordPopupVisible(true) : callback();
```
muhurov, line 58: may be it does make sense to make a constant for 'owner'? or may be to make an utility function to check for 'owner' role?

muhurov (approved):

src/modules/Dashboard/features/Applets/Table/AppletItem/AppletItem.types.ts
```diff
@@ -13,3 +13,12 @@ export type Actions = {
   };
   item: FolderApplet;
 };
+
+export const enum Roles {
```
wildkiomi, line 17: we already have it in shared/consts.tsx

wildkiomi (approved):

PR #414, opened 2023-05-17
Fix/M2-1875 Individual Schedule: calendar icon is not displaying from the left side of the respondent's name with the scheduled event

IhorNesterenko99:
…m the left side of the respondent's name with the scheduled event


src/modules/Dashboard/state/Users/Users.thunk.ts
```diff
@@ -15,6 +15,17 @@ export const getWorkspaceRespondents = createAsyncThunk(
   },
 );
 
+export const getAllWorkspaceRespondents = createAsyncThunk(
```
kdanilenka32, line 18: Why need the thunk?

IhorNesterenko99: To divide respondents with the limit of 20 from all respondents. In the schedule I need to receive all respondents.

kdanilenka32: the do not have limit by default on back?
image

src/modules/Dashboard/features/Applet/Schedule/Legend/Legend.tsx
```diff
@@ -34,7 +34,7 @@ export const Legend = ({ legendEvents, appletName, appletId }: LegendProps) => {
   const { t } = useTranslation('app');
   const { respondentId } = useParams();
   const navigate = useNavigate();
-  const { result: respondentsData } = users.useRespondentsData() || {};
+  const { result: respondentsData } = users.useAllRespondentsData() || {};
```
kdanilenka32, line 37: and in that case do y really need redux?

src/modules/Dashboard/state/Users/Users.thunk.ts
IhorNesterenko99: thanks for clarifying, I thought there is no limit by default

IhorNesterenko99: added limit 10000

kdanilenka32 (approved):

PR #452, opened 2023-05-25
Fix: M2-690 - Builder - Applet About Page: The image is removed with no confirmation pop-up after clicking the Remove on the uploaded

muhurov:


alex-i-e (approved):

src/shared/components/Uploader/Uploader.tsx
```diff
@@ -194,6 +213,11 @@ export const Uploader = ({
           image={image}
         />
       )}
+      <RemoveImagePopup
+        open={!!isRemovePopupOpen}
```
IhorNesterenko99, line 217: why double exclamation ('!!') but not just isRemovePopupOpen?

muhurov: fixed

IhorNesterenko99 (approved):

PR #459, opened 2023-05-26
Feature/m2 1626

kdanilenka32:


src/modules/Dashboard/features/Respondents/Respondents.tsx
```diff
@@ -141,28 +168,33 @@ export const Respondents = () => {
...
+          chosenRespondentsItems[key],
           setChosenAppletData,
-          chosenRespondentsItems?.id,
+          respondentKey,
           ownerId,
         )
       : undefined;
 
+  const viewableAppletsSmallTableRows = getAppletsSmallTable('viewable');
+  const editableAppletsSmallTableRows = getAppletsSmallTable('editaable');
```
IhorNesterenko99, line 195: 'editable'

src/shared/hooks/usePermissions.tsx
```diff
@@ -0,0 +1,34 @@
...
+    (async () => {
+      try {
+        const { payload } = await asyncFunc();
+
+        if (payload?.response?.status === 403 || payload?.status === 403) {
+          return setIsForbidden(true);
+        }
+        setIsForbidden(false);
+      } catch (e) {
+        getErrorMessage(e);
```
IhorNesterenko99, line 25: in case if there is an error isForbidden will be false
is it right?
and maybe it make sense to return and show an error message in case of error?

src/modules/Dashboard/features/Applet/Schedule/Schedule.tsx
```diff
@@ -48,6 +51,8 @@ export const Schedule = () => {
     };
   }, [appletId, respondentId, ownerId]);
```
IhorNesterenko99, line 52: it seems there is no need in ownerId in dependencies anymore

src/modules/Dashboard/features/Applet/Schedule/Schedule.tsx
```diff
@@ -18,9 +18,18 @@ export const Schedule = () => {
...
   const { result: appletData } = applet.useAppletData() ?? {};
   const { ownerId } = workspaces.useData() || {};
+  const { getAllWorkspaceRespondents } = users.thunk;
   const appletId = appletData?.id || '';
   const preparedEvents = usePreparedEvents(appletData);
 
+  const { isForbidden, noPermissionsComponent } = usePermissions(() =>
+    dispatch(
+      getAllWorkspaceRespondents({
+        params: { ownerId, appletId },
```
IhorNesterenko99, line 28: what if ownerId here is undefined?

kdanilenka32: inside hook we have return

src/modules/Dashboard/features/Applet/Schedule/Schedule.tsx
kdanilenka32: done

src/shared/hooks/usePermissions.tsx
kdanilenka32: the error doesn't even seem to be caught there

src/modules/Dashboard/features/Respondents/Respondents.tsx
kdanilenka32: fixed

nmialik (approved):

PR #487, opened 2023-06-07
feature/M2-2221-report-csv

kdanilenka32:


.eslintrc
```diff
@@ -52,7 +52,7 @@
     "no-undef-init": "error",
     "no-nested-ternary": "error",
     "no-unneeded-ternary": "error",
-    "camelcase": "error",
```
alex-i-e, line 55: maybe it's better to add an exception in a particular file `/* eslint-disable camelcase */` ?

alex-i-e (approved):

.eslintrc
kdanilenka32: we already had fe files,  I bel not important rule

PR #499, opened 2023-06-12
Fix/m2 1288 fix applet password saving

alex-i-e:


src/shared/components/Password/CreateAppletPassword/CreateAppletPassword.hooks.ts
```diff
@@ -0,0 +1,26 @@
+import { Encryption, getAppletEncryptionInfo, getParsedEncryptionFromServer } from 'shared/utils';
```
kdanilenka32, line 1: mb it is better to move the hook globally

kdanilenka32 (approved):

PR #502, opened 2023-06-13
Fix/m2 2252 hover interaction for table row

alex-i-e:


src/shared/components/Table/Table.types.ts
```diff
@@ -3,9 +3,10 @@ import { ReactNode } from 'react';
 import { Cell, HeadCell } from 'shared/types/table';
 
 export type RowContent = Cell & {
-  content: (item?: Row) => ReactNode | string;
+  content: (item?: Row, hasVisibleActions?: boolean) => ReactNode | string;
```
kdanilenka32, line 6: It seams no need string here

src/modules/Dashboard/features/Managers/Managers.tsx
```diff
@@ -121,17 +121,23 @@ export const Managers = () => {
...
+                  context={filteredManager}
+                  visibleByDefault={hasVisibleActions}
+                />
+              );
             },
             value: '',
             width: '20%',
           },
-        };
+        } as Row;
```
kdanilenka32, line 140: Why need  as Row here?

alex-i-e: moved type in top level

kdanilenka32 (approved):

PR #518, opened 2023-06-14
Fix: M2-2251 - Builder - Correct all issues on Scores & Reports page

muhurov:


src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.tsx
```diff
@@ -115,33 +116,35 @@ export const ScoresAndReports = () => {
           </StyledFlexTopCenter>
         }
       />
-      {sections?.map((section, index) => {
-        const sectionName = `${sectionsName}.${index}`;
-        const title = t('sectionHeader', {
-          index: index + 1,
-          name: section?.name,
-        });
+      <StyledFlexColumn sx={{ mt: theme.spacing(2.4) }}>
```
IhorNesterenko99, line 119: what if there are no sections, but this StyledFlexColumn is visible without sections?

alex-i-e (approved):

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.tsx
muhurov: so if the are no sections, it works like a spacer (just gives necessary margin for all other elements below)

IhorNesterenko99 (approved):

PR #555, opened 2023-06-23
Feature/m2 2361 add validations for variables in wisywyg

alex-i-e:


src/modules/Builder/pages/BuilderApplet/BuilderApplet.const.ts
```diff
@@ -8,3 +8,15 @@ export const CONDITION_TYPES_TO_HAVE_OPTION_ID = [
 ];
 
 export const defaultFlankerBtnObj = { name: '', image: '' };
+
+export const RESTRICTED_TYPES_IN_VARIABLES = [
```
alex-i-e, line 12: we be updated with `allowed` types

muhurov (approved):

wildkiomi (approved):

PR #650, opened 2023-07-13
Fix/m2 2543 activity user journey responses should be filled in sequentially

alex-i-e:


src/shared/utils/exportData/parseResponseValue.ts
```diff
@@ -11,18 +12,27 @@ import {
...
 import { getAnswerValue } from '../getAnswerValue';
 
-export const parseResponseValue = (answer: AnswerDTO, activityItem: Item, id: string) => {
+export const parseResponseValue = <
+  T extends DecryptedAnswerData<ExtendedExportAnswerWithoutEncryption>,
+>(
+  item: T,
+  isEvent = false,
+) => {
+  let answer: AnswerDTO | undefined = item.answer;
```
muhurov, line 30: what about
const answer = isEvent ? (item ...).response : item.answer
to prevent using let?

muhurov (approved):

IhorNesterenko99 (approved):

muhurov (approved):

PR #657, opened 2023-07-17
M2-2595 Export abtrails

wildkiomi:


src/shared/utils/exportData/parseResponseValue.ts
```diff
@@ -62,6 +63,8 @@ export const parseResponseValue = <
       })`;
     case ItemResponseType.Drawing:
       return getMediaFileName(item, 'svg');
+    case ItemResponseType.ABTrails:
+      return value as DecryptedABTrailsAnswer['value'];
```
alex-i-e, line 67: here should be a csv file name which will be exported

alex-i-e (approved):

nmialik (approved):

PR #662, opened 2023-07-18
M2-2676 Min score range

wildkiomi:


src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.utils.ts
```diff
@@ -34,25 +34,25 @@ const getItemScoreRange = (item: Item) => {
...
-        if (!option.isHidden && option.score) {
-          return [...result, option.score];
+        if (!option.isHidden) {
+          return [...result, option.score as number];
         }
 
         return result;
       },
       [],
     ) as unknown as number[];
```
alex-i-e, line 44:
```suggestion
    ) as number[];
```

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.utils.ts
```diff
@@ -34,25 +34,25 @@ const getItemScoreRange = (item: Item) => {
   ) {
     scores = (item as SingleSelectItem | MultiSelectItem).responseValues.options?.reduce(
       (result: number[], option: SingleAndMultipleSelectionOption) => {
-        if (!option.isHidden && option.score) {
-          return [...result, option.score];
+        if (!option.isHidden) {
+          return [...result, option.score as number];
```
alex-i-e, line 38: could you check if `option.score` should have only `number` type? Why do you need to cast?

alex-i-e (approved):

PR #685, opened 2023-07-24
Fix: M2-2323 - Admin Panel - Log out after closing the tab

muhurov:


src/shared/utils/authStorage.ts
```diff
@@ -0,0 +1,27 @@
...
+  try {
+    return data ? JSON.parse(data) : null;
+  } catch {
+    return data;
+  }
+};
+const remove = (key: string) => sessionStorage.removeItem(key);
+
+export const authStorage = {
+  getRefreshToken: () => get(LocalStorageKeys.RefreshToken),
```
alex-i-e, line 18: Looks incorrect to use LocalStorage keys in session storage, maybe devide them into two separate enums?

muhurov: fixed

alex-i-e (approved):

kdanilenka32 (approved):

PR #701, opened 2023-07-26
Fix: M2-2774 - Data export: report, activity user journey.csv: activity_start_time, activity_end_time are not parsed correctly

muhurov:


alex-i-e (approved):

alex-i-e:
Btw, these issues should be finished on Backend side as well

kdanilenka32 (approved):

PR #702, opened 2023-07-26
M2-2628 Editor scale image issue

wildkiomi:


src/shared/styles/theme.ts
```diff
@@ -29,6 +29,20 @@ export const theme = createTheme({
           height: '100vh',
           overflowY: 'hidden',
         },
+        '.medium-zoom-image--opened': {
+          visibility: 'hidden',
+          zIndex: 0,
```
alex-i-e, line 34: do we have a chance to use the internal enum for material ui zIndex like `theme.zIndex.appBar` and other ?

nmialik (approved):

alex-i-e (approved):

PR #710, opened 2023-07-26
Fix/m2 1288 remove password check when save and publish applet

alex-i-e:


src/modules/Builder/hooks/useAppletPrivateKeySetter.ts
```diff
@@ -1,8 +1,8 @@
 import { Encryption, getAppletEncryptionInfo, getParsedEncryptionFromServer } from 'shared/utils';
-import { useEncryptionCheckFromStorage } from 'shared/hooks';
+import { useEncryptionStorage } from 'shared/hooks/useEncryptionStorage';
```
muhurov, line 2: do you really need additional '/useEncryptionStorage'? if not, please remove in all places

muhurov (approved):

src/modules/Builder/hooks/useAppletPrivateKeySetter.ts
alex-i-e: fixed

muhurov (approved):

kdanilenka32 (approved):

PR #744, opened 2023-08-07
M2-2870 Score report common fields validation

wildkiomi:


src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -785,6 +785,20 @@ export const getAppletTabs = ({
   },
 ];
 
+export const testIsReportCommonFieldsRequired = (
+  isScoreReport: boolean,
+  value: boolean,
+  context: unknown,
+) => {
+  if (isScoreReport) {
+    const conditionalLogic = get(context, 'from.0.value.conditionalLogic')?.length;
```
IhorNesterenko99, line 794:
```suggestion
    const hasConditionalLogic = get(context, 'from.0.value.conditionalLogic')?.length;
```

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -785,6 +785,20 @@ export const getAppletTabs = ({
...
+
+    return !!conditionalLogic || value;
```
IhorNesterenko99, line 796: is the double exclamation mark needed here?
as I see above, conditional logic can be undefined, 0, or other number

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -785,6 +785,20 @@ export const getAppletTabs = ({
   },
 ];
 
+export const testIsReportCommonFieldsRequired = (
+  isScoreReport: boolean,
+  value: boolean,
```
IhorNesterenko99, line 790: could you pls give more understandable name for this parameter?

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
wildkiomi: this function should return boolean. so !! is required

IhorNesterenko99 (approved):

muhurov (approved):

PR #818, opened 2023-08-21
fix: M2-3039 Coordinator/Reviewer is able to 'Add to Builder' to the shared workspace

IhorNesterenko99:


src/modules/Library/features/Popups/AddToBuilderPopup/AddToBuilderPopup.tsx
```diff
@@ -32,21 +33,21 @@ export const AddToBuilderPopup = ({
...
   const [applets, setApplets] = useState<Applet[]>([]);
-  const { result: workspacesData = [] } = workspaces.useWorkspacesData() || {};
-  const currentWorkspace = workspaces.useData();
+  const workspacesWithRoles = workspacesState.useWorkspacesRolesData();
+  const workspaces =
+    workspacesWithRoles?.filter(
+      (workspace) =>
+        Object.keys(workspace.workspaceRoles).length === 0 ||
+        Object.values(workspace.workspaceRoles).some(
+          (roles) => roles.includes(Roles.Owner) || roles.includes(Roles.Manager),
```
kdanilenka32, line 45: we have utils ownreormanger check

kdanilenka32: and as I remember owner and manager always on 0 index

src/modules/Library/features/Popups/AddToBuilderPopup/AddToBuilderPopup.utils.tsx
```diff
@@ -265,8 +267,8 @@ export const getSteps = ({
 };
 
 export const getArrayFromApplets = (applets: FullApplet[]) =>
-  applets.reduce((acc: Applet[], { id, type, displayName, image }) => {
-    if (type === DashboardAppletType.Applet) {
+  applets.reduce((acc: Applet[], { id, type, displayName, image, role }) => {
+    if (type === DashboardAppletType.Applet && (role === Roles.Owner || role === Roles.Manager)) {
```
kdanilenka32, line 271: same here

IhorNesterenko99: as I understand in applets list there is only one main role which is a string
image

src/shared/components/Spinner/Spinner.styles.ts
```diff
@@ -1,12 +1,13 @@
...
+import { shouldForwardProp } from 'shared/utils';
 
-export const StyledSpinner = styled(StyledFlexAllCenter)`
+export const StyledSpinner = styled(StyledFlexAllCenter, shouldForwardProp)`
   position: absolute;
   width: 100%;
   height: 100%;
-  background-color: rgb(255 255 255 / 50%);
+  background-color: ${({ noBackground }: { noBackground?: boolean }) =>
+    noBackground ? 'transparent' : 'rgb(255 255 255 50%)'};
```
nmialik, line 11: 'rgb(255 255 255 50%)' - in this format the color is displayed? no need for a slash 'rgb(255 255 255 / 50%)' ?
and maybe it makes sense to move it to variables?

kdanilenka32 (approved):

nmialik (approved):

PR #839, opened 2023-08-25
Fix: M2-3086 - Admin Panel - Add auth/logout API call on logout

muhurov:


src/modules/Auth/api/api.ts
```diff
@@ -13,6 +13,8 @@ export const signInApi = ({ email, password }: SignIn, signal?: AbortSignal) =>
 
 export const logOutApi = () => authApiClientWithoutRefresh.post('auth/logout');
```
IhorNesterenko99, line 14: just for convenience and to make the code more understandable, could you please name deleteAccessTokenApi and deleteRefreshTokenApi?

muhurov: done

IhorNesterenko99 (approved):

kdanilenka32 (approved):

PR #840, opened 2023-08-28
Fix/m2 3046  copying is not failed if user tries to copy the applet from library in offline

nmialik:


src/shared/hooks/useNetwork.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  useEffect(() => {
+    window.addEventListener('offline', updateNetwork);
+    window.addEventListener('online', updateNetwork);
+
+    return () => {
+      window.removeEventListener('offline', updateNetwork);
+      window.removeEventListener('online', updateNetwork);
+    };
+  });
```
kdanilenka32, line 18: It should be every render?

src/shared/hooks/useNetwork.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState } from 'react';
+
+export const useNetwork = () => {
+  const [isOnline, setNetwork] = useState(window.navigator.onLine);
```
IhorNesterenko99, line 4:
```suggestion
  const [isOnline, setIsOnline] = useState(window.navigator.onLine);
```

kdanilenka32 (approved):

IhorNesterenko99 (approved):

PR #849, opened 2023-08-29
M2-3133: Dataviz: Add encrypted identifiers

nmialik:


src/modules/Dashboard/hooks/useDecryptedIdentifiers.ts
```diff
@@ -20,6 +20,14 @@ export const useDecryptedIdentifiers = () => {
 
   return (identifiers: IdentifierResponse[]): Identifier[] =>
     identifiers.map(({ identifier, userPublicKey }) => {
+      // workaround for decrypted identifier data
```
alex-i-e, line 23: does it depend on a pending task on BE? When it should works properly?

IhorNesterenko99 (approved):

kdanilenka32 (approved):

src/modules/Dashboard/hooks/useDecryptedIdentifiers.ts
nmialik: identifiers were not encrypted in legacy 
so we need to do an additional check on the client

PR #863, opened 2023-08-31
fix/3185-alerts-values-fix

kdanilenka32:


nmialik (approved):

src/shared/utils/exportData/parseDate.ts
```diff
@@ -0,0 +1 @@
+export const parseDate = (date: string | number) => `${Math.floor(+date / 1000)}`;
```
IhorNesterenko99, line 1: maybe give more specific name, for example: convertDateStampToMilliseconds?

muhurov (approved):

IhorNesterenko99 (approved):

PR #876, opened 2023-09-04
fix: M2-3288 Report Configuration. Subject Preview is not changing if Item Value selected.

IhorNesterenko99:


src/shared/features/AppletSettings/ReportConfigSetting/ReportConfigSetting.tsx
```diff
@@ -350,12 +363,12 @@ export const ReportConfigSetting = ({
 
   useEffect(() => {
     reset(defaultValues);
-    setSubjectData(setValue, includeRespondentId);
+    setSubjectData(subjectDataProps);
```
alex-i-e, line 366: why not remove this function here and add a dependency `appletData` in useEffect below ?

alex-i-e (approved):

PR #900, opened 2023-09-07
M2-3325 - Settings - Save changes popup doesn't appear in settings

muhurov:


src/shared/layouts/BaseLayout/components/TopBar/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -36,7 +36,7 @@ export const Breadcrumbs = () => {
         ) : (
           <StyledPlaceholder>
             <StyledLabelSmall color={variables.palette.on_surface}>
-              {label.substring(0, 1).toUpperCase()}
+              {label?.substring(0, 1).toUpperCase()}
```
IhorNesterenko99, line 39: maybe add || '' or smth like this to not show undefined?
or it's not possible that label will be undefined?

alex-i-e (approved):

src/shared/layouts/BaseLayout/components/TopBar/Breadcrumbs/Breadcrumbs.tsx
muhurov: fixed

src/shared/layouts/BaseLayout/components/TopBar/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -29,6 +29,8 @@ export const Breadcrumbs = () => {
       <Svg id={icon} width={BREADCRUMB_ICON_SIZE} height={BREADCRUMB_ICON_SIZE} />
     );
 
+    if (!icon && !label) return null;
```
IhorNesterenko99, line 32:
```suggestion
    if (!icon || !label) return null;
```

IhorNesterenko99 (approved):

PR #904, opened 2023-09-07
M2-3079: Applet is not expanded to the level of the result

nmialik:


src/modules/Library/features/Applet/Applet.utils.ts
```diff
@@ -1,4 +1,9 @@
-import { PublishedApplet } from 'modules/Library/state';
+import { Dispatch, SetStateAction } from 'react';
+
+import { ExpandedActivity } from 'modules/Library/features/Applet/Applet.types';
```
IhorNesterenko99, line 3:
```suggestion
import { ExpandedActivity } from './Applet.types';
```

src/modules/Library/features/Applet/Applet.utils.ts
```diff
@@ -1,4 +1,9 @@
-import { PublishedApplet } from 'modules/Library/state';
+import { Dispatch, SetStateAction } from 'react';
+
+import { PublishedActivity, PublishedApplet } from 'modules/Library/state';
+import { ItemResponseType } from 'shared/consts';
+import { SingleAndMultipleSelectItemResponseValues } from 'shared/state';
+import { ExpandedActivity } from './Applet.types';
```
IhorNesterenko99, line 6: pls add empty row before

alex-i-e (approved):

kdanilenka32 (approved):

PR #916, opened 2023-09-12
feat: M2-3376 Sanitize fields from Cross-site scripting (XSS) vulnerabilities and HTML injection. Prohibit to use of links as applet names.

IhorNesterenko99:


src/shared/components/FormComponents/EditorController/EditorController.tsx
```diff
@@ -62,6 +62,7 @@ export const EditorController = <T extends FieldValues>({
         render={({ field: { onChange, value }, fieldState: { error } }) => (
           <StyledFlexColumn sx={{ position: 'relative' }} data-testid={dataTestid}>
             <StyledMdEditor
+              sanitize={(content: string) => getSanitizedContent(content)}
```
alex-i-e, line 65: could you check this suggestion, doesn't it throw TS error/warning?
```suggestion
              sanitize={getSanitizedContent}
```

IhorNesterenko99: yes it throws because of other getSanitizedContent function arguments
that's why, I'm using this approach

PR #929, opened 2023-09-14
M2-3333, M2-3346: Fix translations

nmialik:


src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.const.ts
```diff
@@ -3,14 +3,14 @@ import { CalculationType } from 'shared/consts';
 
 const { t } = i18n;
 
-export const scoreItemsColumns = [
+export const getScoreItemsColumns = () => [
```
IhorNesterenko99, line 6: сould you pls move functions to ...utils.ts?
also in other places where constants were replaced with functions

IhorNesterenko99 (approved):

alex-i-e (approved):

PR #933, opened 2023-09-14
Fix: M2-3338 - Builder - Replace 'react-hook-form' watch for arrays with fields from useFieldArray hook, if it's possible, to increase components performance.

muhurov:


src/modules/Builder/consts.ts
```diff
@@ -0,0 +1,2 @@
+//TODO: remove in all places where it was used after the next major react-hook-form update
```
alex-i-e, line 1: maybe add more details for exact version:
```suggestion
//TODO: remove in all places where it was used after release the v.8.0.0 react-hook-form
```

muhurov: done

alex-i-e (approved):

PR #936, opened 2023-09-15
M2-2234 Activity item route

wildkiomi:


alex-i-e (approved):

kdanilenka32 (approved):

muhurov:
I can't find where the item becomes active from the route param after refreshing the page (or after changing active item via LeftBar)?

wildkiomi:
@muhurov for second part of your question there was a useEffect which handle navigation base on active item.
for now I made some changes. check pls

PR #940, opened 2023-09-18
fix/M2-2094-uncount-special-symbols

kdanilenka32:


nmialik (approved):

alex-i-e (approved):

src/shared/components/FormComponents/EditorController/EditorController.utils.ts
```diff
@@ -0,0 +1,12 @@
+const SPECIAL_SYMBOLS_LENGTH = 4;
+
+export const getValueLength = (value = '') => {
```
IhorNesterenko99, line 3: In my opinion, this function needs to be improved to match not only one special case with "**", but also other different cases in the editor, where special characters are used.
For example:
image

PR #946, opened 2023-09-18
fix: M2-906 Applet Page: Applet Settings: Data retention tab: It is possible to enter dot in the period

IhorNesterenko99:


muhurov (approved):

src/shared/features/AppletSettings/DataRetentionSetting/DataRetention.tsx
```diff
@@ -93,6 +93,13 @@ export const DataRetention = ({ isDashboard }: { isDashboard?: boolean }) => {
     cancelNavigation();
   };
 
+  const handlePeriodKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+    // Prevent the dot key from being entered
```
nmialik, line 97: how about "," б should we prevent this char as well?

nmialik (approved):

alex-i-e (approved):

src/shared/features/AppletSettings/DataRetentionSetting/DataRetention.tsx
IhorNesterenko99: it's prevented by default in the Input type number

PR #963, opened 2023-09-20
fix: M2-3518 Fix activity flow navigation issue

IhorNesterenko99:


src/modules/Builder/hooks/useActivityFlowsRedirection.ts
```diff
@@ -14,16 +14,17 @@ export const useActivityFlowsRedirection = () => {
...
 
-  const activities = watch('activities');
-  const activityFlows = watch('activityFlows');
-  const activityFlow = activityFlows?.find(
-    (flow: ActivityFlowFormValues) => activityFlowId === getEntityKey(flow),
-  );
-  const shouldRedirect = (isNewApplet || activities?.length > 0) && activityFlowId && !activityFlow;
-
   useEffect(() => {
+    const activities = watch('activities');
```
kdanilenka32, line 18: Is it ok watch inside useEffect?

IhorNesterenko99: there are no errors, it seems it's just a function

kdanilenka32: if use getValues instead of create subscriptions

alex-i-e: please don't unsubscribe then 
  // Callback version of watch.  It's your responsibility to unsubscribe when done.
  React.useEffect(() => {
    const subscription = watch((value, { name, type }) => console.log(value, name, type));
    return () => subscription.unsubscribe();
  }, [watch]);

alex-i-e (changes requested):
please don't forget to unsubscribe then 
https://www.react-hook-form.com/api/useform/watch/
  // Callback version of watch.  It's your responsibility to unsubscribe when done.
  React.useEffect(() => {
    const subscription = watch((value, { name, type }) => console.log(value, name, type));
    return () => subscription.unsubscribe();
  }, [watch]);

src/modules/Builder/hooks/useActivitiesRedirection.ts
```diff
@@ -12,15 +12,14 @@ export const useActivitiesRedirection = () => {
...
   const { activity } = useCurrentActivity();
   const isNewApplet = useCheckIfNewApplet();
-
   const { watch } = useFormContext();
 
-  const activities = watch('activities');
-  const shouldRedirect = (isNewApplet || activities?.length > 0) && activityId && !activity;
-
   useEffect(() => {
+    const activities = watch('activities');
```
alex-i-e, line 18: please don't forget to unsubscribe then
https://www.react-hook-form.com/api/useform/watch/

alex-i-e (approved):

kdanilenka32 (approved):

PR #970, opened 2023-09-21
M2-3507 Show spinner instead of white page on applet

wildkiomi:


alex-i-e (changes requested):
do we need to show spinner in case of an error?

muhurov (approved):

alex-i-e (approved):

PR #985, opened 2023-09-25
fix: M2-3554 TypeError is displayed after setting scores and reports in duplicated applet

IhorNesterenko99:


muhurov (approved):

nmialik (approved):

src/modules/Builder/components/ConditionRow/ConditionRow.utils.ts
```diff
@@ -53,7 +53,7 @@ export const getItemOptions = (items: ItemFormValues[], conditionRowType: Condit
     }
 
     return optionList;
-  }, []);
+  }, []) || [];
```
alex-i-e, line 56: Is not it enough to have initialValue in `reduce`?

IhorNesterenko99: if items undefined, then items?.reduce(...)  reduce is not launching, and function returns undefined

alex-i-e (approved):

PR #1003, opened 2023-09-27
fix/m2-3513-time-resp-fix

kdanilenka32:


IhorNesterenko99 (approved):

src/shared/utils/exportData/parseResponseValue.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-case-declarations */
```
nmialik, line 1: u can use just { ... } for `case` and remove this line

```
case ItemResponseType.Time: {
....
}
```

kdanilenka32: fixed

PR #1004, opened 2023-09-27
M2-3618: Library: items should not be dispalying if search is done for Performance tasks

nmialik:


src/modules/Library/features/Applet/Applet.utils.ts
```diff
@@ -34,6 +38,16 @@ export const getActivities = (
       isActivitiesExpanded = true;
     }
 
+    const isPerformanceTask = checkIfPerformanceTask(activity.items);
+    if (isPerformanceTask) {
```
IhorNesterenko99, line 42: maybe !isPerformanceTask, or I'm wrong

nmialik: no

nmialik: because if ! isPerformanceTask we need to check items (what is done below)

alex-i-e (approved):

IhorNesterenko99 (approved):

PR #1005, opened 2023-09-27
M2-3592: Header image has wrong ratio

nmialik:


IhorNesterenko99 (approved):

src/shared/components/MarkDownEditor/MarkDownEditor.styles.ts
```diff
@@ -29,10 +29,6 @@ const commonStyles = `
     text-align: right;
   }
   
-  .md-editor-preview-wrapper img {
-    max-width: 50%;
```
IhorNesterenko99, line 33: maybe better 100%, or it's by default?
because image width may be bigger than preview section width

nmialik: 100% it's default for img

IhorNesterenko99 (approved):

alex-i-e (approved):

IhorNesterenko99 (approved):

PR #1008, opened 2023-09-28
M2-3340 Update npm packages and fix vulnerabilities

wildkiomi:
**xlsx had a prototype pollution in an old version** and also it had been moved to an another repo and it will not publish its updates in npm anymore. so the only option to upgrade to a new version without prototype pollution - from cdn
**react-scripts is an abandoned package.** the latest version _doesn't support typescript ^5._ it's better to replace react-scripts with vite later
yup and yup-resolvers have now more strict type checks and new syntax for yup.when

Before there were **17 vulnerabilites (6 moderate, 11 high)**
Now there are only **6 high**, and they couldn't be fixed actually (they could be fixed by downgrading version of react-scripts from 5 to 2, but if i did it, there would be like 82 high vulnerabilities - that's not an option)


IhorNesterenko99 (approved):

package.json
```diff
@@ -38,13 +39,12 @@
...
     "lodash.uniqby": "^4.7.0",
     "lodash.uniqueid": "^4.0.1",
-    "md-editor-rt": "^4.7.0",
     "markdown-it-abbr": "^1.0.4",
     "markdown-it-container": "^3.0.0",
     "markdown-it-deflist": "^2.1.0",
     "markdown-it-emoji": "^2.0.2",
     "markdown-it-footnote": "^3.0.3",
-    "markdown-it-html5-embed": "^1.0.0",
+    "markdown-it-html5-embed": "^0.3.3",
```
alex-i-e, line 47: why it was downgraded?

wildkiomi: because new version 1.0.0 has a vulnerability. when it will be fixed and will be released a new version, when we can upgrade it

src/shared/utils/exportData/getFlankerRecords.ts
```diff
@@ -121,7 +121,7 @@ const getResponseObj = ({
     responseTouchTimestamp =
       'response_touch_timestamp' in response
         ? response.response_touch_timestamp || DEFAULT_VALUE
-        : videoDisplayRequestTimestamp + response.duration;
+        : videoDisplayRequestTimestamp + (<DecryptedFlankerAnswerItemValue>response).duration;
```
alex-i-e, line 124: it has type declaration in row 85, why it does not work?

src/shared/utils/urlGenerator.ts
```diff
@@ -133,8 +131,8 @@ export const checkCurrentActivityPage = (url: string) => ({
 
 export const checkCurrentActivityFlowPage = (url: string) => ({
   isAbout: new RegExp(`${getAppletActivityFlowPageRegexp(Path.About)}`).test(url),
-  isBuilder: new RegExp(`${getAppletActivityFlowPageRegexp(Path.FlowBuilder)}`).test(url),
-  isSettings: new RegExp(`${getAppletActivityFlowPageRegexp(Path.FlowSettings)}`).test(url),
+  isBuilder: new RegExp(`${getAppletActivityFlowPageRegexp(Path.Builder)}`).test(url),
```
alex-i-e, line 134: should not FlowBuilder|FlowSettings be removed from enum as well?

src/shared/utils/exportData/getFlankerRecords.ts
wildkiomi: I get such an error "Property 'duration' does not exist on type 'never'"

src/shared/utils/urlGenerator.ts
wildkiomi: I removed it

alex-i-e (approved):

PR #1014, opened 2023-09-28
fix: M2-3593 Flanker_360: Fixation ‘-----’ is missing before the stimulus screen, fix performance tasks validation

IhorNesterenko99:


src/modules/Builder/pages/BuilderApplet/BuilderApplet.schema.ts
```diff
@@ -375,36 +346,76 @@ export const ItemSchema = () =>
...
+      }),
+    })
+    .required();
+
+export const FlankerSchema = () =>
+  yup
+    .object({
+      question: yup.string().when('order', (order, schema) => {
+        if (order === 1) {
+          return schema.required(getIsRequiredValidateMessage('overviewInstruction'));
```
alex-i-e, line 368: could you please provide a map which is `order`s match the `instruction`s?

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -257,11 +247,12 @@ const defaultMessageConfig = {
   timer: null,
 };
 
-const getMessageItem = (name: string, question: string) => ({
+const getMessageItem = (name: string, question: string, order?: number) => ({
```
alex-i-e, line 250: please use an object of parameters

kdanilenka32 (approved):

alex-i-e (approved):

PR #1031, opened 2023-10-02
M2-2945: Summary tab: wrong calculation of Total Average value if lookup table is added to Subscale

nmialik:


IhorNesterenko99 (approved):

muhurov (approved):

src/shared/utils/isSystemItem.ts
```diff
@@ -0,0 +1,4 @@
+import { LookupTableItems } from 'shared/consts';
+
+export const isSystemItem = (name: string) =>
```
alex-i-e, line 3: could please remove `isSystemItem` from src/modules/Builder/features/ActivitySettings/SubscalesConfiguration/SubscalesConfiguration.utils.tsx
and use only from one place?

alex-i-e (changes requested):
could please remove isSystemItem from src/modules/Builder/features/ActivitySettings/SubscalesConfiguration/SubscalesConfiguration.utils.tsx
and use only from one place?

src/shared/utils/isSystemItem.ts
nmialik: it has been already removed

alex-i-e: yeap, I missed it, apparently my eyes are tired at the end of the working day :)

PR #1050, opened 2023-10-05
fix/M2-3476-legacyProfileId

kdanilenka32:


src/shared/utils/exportData/getReportCSVObject.ts
```diff
@@ -67,5 +68,6 @@ export const getReportCSVObject = <T>({
     version,
     rawScore: getRawScores(responseValues) || '',
     reviewing_id: reviewedAnswerId,
+    ...(legacyProfileId && { legacy_user_id: legacyProfileId }),
```
alex-i-e, line 71: just to confirm, should not the `legacy_user_id` column be the next with `userId` column ?

alex-i-e (approved):

src/shared/utils/exportData/getReportCSVObject.ts
kdanilenka32: Valdimir sad in the end

IhorNesterenko99 (approved):

muhurov (approved):

PR #1052, opened 2023-10-06
fix: M2-3696 Write unit tests for some shared functions (array.ts, getDateTime.ts, getAnswerValue.ts)

IhorNesterenko99:


src/shared/utils/array.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  test('creates an empty array when length is 0', () => {
+    const result = createArray(0, () => 'item');
+    const expected: string[] = [];
+
+    expect(result).toEqual(expected);
+  });
+});
+
+describe('groupBy', () => {
+  test('groups an array of objects by a key', () => {
```
kdanilenka32, line 20: Mb by  key?

muhurov (approved):

src/shared/utils/array.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    ];
+
+    const result = pluck(array, 'person.name');
+    const expected = ['John', 'Jane', 'Bob'];
+
+    expect(result).toEqual(expected);
+  });
+});
+
+describe('createArrayFromMinToMax', () => {
```
alex-i-e, line 94: I would propose to use test templates to test such utitilites:
```
  test.each`
    min   | max    | result             | description
    ${3}  | ${7}   | ${[3, 4, 5, 6, 7]} | ${'returns range of values'}
    ${5}  | ${5}   | ${[5]}             | ${'returns one item in array for same boundaries'}
    ${10} | ${[5]} | ${[]}              | ${'returns empty array when the borders change sides'}
  `('$description', ({ min, max, result }) => {
    expect(createArrayFromMinToMax(min, max)).toEqual(result);
  });
```
It's much simpler maintenance and blazzy fast enhance test coverage with new test cases

kdanilenka32 (approved):

alex-i-e (approved):

PR #1060, opened 2023-10-10
tests: M2-3727 useTable and useTimeAgo hook tests

wildkiomi:


src/shared/hooks/useTimeAgo.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+  }),
+  initReactI18next: {
+    type: '3rdParty',
+    init: () => {},
+  },
+}));
+
+describe('useTimeAgo hook tests', () => {
+  test('useTimeAgo should format date', () => {
```
IhorNesterenko99, line 20: maybe also try some false value, incorrect data, future data, and minutes ago and just now?

src/shared/hooks/useTable.test.ts
```diff
@@ -0,0 +1,66 @@
...
+jest.mock('redux/modules', () => ({
+  workspaces: {
+    useData: () => ({ ownerId: 'mockOwnerId' }),
+  },
+}));
+
+describe('useTable hook tests', () => {
+  const mockAsyncFn = jest.fn();
+
+  test('useTable should return initial searchValue, order and page', () => {
```
alex-i-e, line 14: Each test runner logs 'description' of parent block and children blocks, so it's redundant to keep name of the parent block in each test, please remove the duplications `useTable` here and in other places.

src/shared/hooks/useTimeAgo.test.ts
```diff
@@ -0,0 +1,30 @@
+import { add } from 'date-fns';
+import { renderHook } from '@testing-library/react';
+
+import { useTimeAgo } from './useTimeAgo';
+
+jest.mock('react-i18next', () => ({
```
alex-i-e, line 6: per my understanding, it's better to move shared logic for test in src/__mocks__ to use it for all the tests
```
const reactI18Next = {
  useTranslation: () => ({
    i18n: {
      changeLanguage: () => new Promise(() => {}),
      language: 'en',
    },
  }),
  initReactI18next: {
    type: '3rdParty',
    init: () => {},
  },
};
```

src/shared/hooks/useTable.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(result.current.page).toBe(1);
+  });
+
+  test('useTable should reload', () => {
+    const { result } = renderHook(() => useTable(mockAsyncFn));
+    act(() => {
+      result.current.handleReload();
+    });
+
+    expect(mockAsyncFn).toBeCalled();
```
alex-i-e, line 29: it's redundant assertion to check that function was called before `expect(mockAsyncFn).toBeCalledTimes(1);`
please remove here and below.

wildkiomi: i don't think it's redundant
if some tests fall, it's better to know why - because function wasn't called at all or because it was called more than 1 time.

src/shared/hooks/useTimeAgo.test.ts
```diff
@@ -0,0 +1,21 @@
+import { add } from 'date-fns';
+import { renderHook } from '@testing-library/react';
+
+import { mockI18Next } from '__mocks__/i18next';
+
+import { useTimeAgo } from './useTimeAgo';
+
+jest.mock('react-i18next', () => mockI18Next);
```
alex-i-e, line 8: It will be automatically mocked. There's no need to explicitly call jest.mock('module_name').

, https://jestjs.io/docs/manual-mocks#mocking-node-modules

alex-i-e (approved):

PR #1077, opened 2023-10-18
Fix/m2 3805 conditional logic disappear after save and publish

alex-i-e:


src/modules/Builder/features/ActivityItemsFlow/ItemFlow/ItemFlow.tsx
```diff
@@ -17,15 +17,12 @@ export const ItemFlow = ({ name, index, onRemove }: ItemFlowProps) => {
...
-    fields: conditions,
-    append: appendCondition,
-    remove: removeCondition,
-  } = useFieldArray({
+  const { control, getFieldState, watch } = useFormContext();
+  const { append: appendCondition, remove: removeCondition } = useFieldArray({
     control,
     name: conditionsName,
   });
+  const conditions = watch(conditionsName);
```
muhurov, line 25: Do we really need it there just for passing it through props? Maybe it does make sense to move it to ItemFlowContent?

alex-i-e: good point, I will check

nmialik (approved):

muhurov (approved):

PR #1094, opened 2023-10-24
Tests: M2-3873 - Unit tests - useWindowSize, useIsTextNodeEllipsed

muhurov:


src/shared/hooks/useWindowSize.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe('useWindowSize', () => {
+  test('should return correct height and width for window', () => {
+    const { result } = renderHook(() => useWindowSize());
+
+    expect(result.current.width).toEqual(window.innerWidth);
+    expect(result.current.height).toEqual(window.innerHeight);
+  });
+
+  const windowWidth = window.innerWidth;
```
IhorNesterenko99, line 13: maybe these variables can be used also for previous test case, so it make sense to move them up?

muhurov: fixed

IhorNesterenko99 (approved):

alex-i-e (approved):

PR #1131, opened 2023-11-02
Tests/m2 3981 url generator

muhurov:


src/shared/utils/urlGenerator.ts
```diff
@@ -52,7 +53,7 @@ export const getAppletPageRegexp = (path?: string) =>
 export const getAppletActivityPageRegexp = (path?: string) =>
   path ? `${ACTIVITY_PAGE_REGEXP_STRING}\\/${path}` : ACTIVITY_PAGE_REGEXP_STRING;
 export const getAppletActivityFlowPageRegexp = (path?: string) =>
-  path ? `${ACTIVITY_FLOW_PAGE_REGEXP_STRING}\\/${path}` : ACTIVITY_PAGE_REGEXP_STRING;
+  path ? `${ACTIVITY_FLOW_PAGE_REGEXP_STRING}\\/${path}` : ACTIVITY_FLOW_PAGE_REGEXP_STRING;
```
alex-i-e, line 56: 👀

alex-i-e (approved):

src/shared/utils/urlGenerator.ts
muhurov: seemed to be a mistake but we didn't find it because this code was never reached :)

src/shared/utils/urlGenerator.test.ts
```diff
@@ -0,0 +1,203 @@
...
+} from './urlGenerator';
+
+const mockedUuid = uuidv4();
+const mockedActivityOrFlowId = uuidv4();
+const mockedItemId = uuidv4();
+const mockedPath = 'path';
+const mockedPerfTaskUndefinedPath = `${ACTIVITIES_PAGE_REGEXP_STRING}\\/${Path.PerformanceTask}\\/undefined/(${uuidRegexp})`;
+const mockedPerfTaskDefinedPath = `${ACTIVITIES_PAGE_REGEXP_STRING}\\/${Path.PerformanceTask}\\/${mockedPath}/(${uuidRegexp})`;
+
+describe('get*Regexp', () => {
```
IhorNesterenko99, line 35: why with "*" ? :)

IhorNesterenko99 (approved):

src/shared/utils/urlGenerator.test.ts
muhurov: It means ['AppletPage', 'AppletActivityPage', 'AppletActivityFlowPage', 'AppletPerformanceActivityPage']
I grouped the same utilities into describes with appropriate names and replaced all changing name parts with '*', but in every test I added full util name for every case

PR #1140, opened 2023-11-03
Feature/m2 4024 add server logger error boundary component

alex-i-e:


src/shared/api/api.ts
```diff
@@ -92,3 +99,55 @@ export const getAppletUniqueNameApi = ({ name }: AppletUniqueName, signal?: Abor
...
+      fileId: string;
+    }>
+  >(`/file/log-file/${deviceId}`, file, {
+    params: {
+      fileId,
+    },
+    signal,
+  });
+
+export const postLogFileCheck = (
```
IhorNesterenko99: do we need postLogFileCheck and getLogFile? 
maybe only leave postLogFile?

alex-i-e: probably, yes. It's more sutable for developer purpose, not in admin panel app.

muhurov (approved):

IhorNesterenko99 (approved):

PR #1184, opened 2023-11-16
Fix/m2 4016 resp nick

kdanilenka32:


src/modules/Dashboard/features/Applet/Schedule/Schedule.tsx
```diff
@@ -36,16 +36,17 @@ export const Schedule = () => {
   );
 
   useEffect(() => {
-    if (!appletId || !hasAccess) return;
+    if (!appletId || !hasAccess || !respondentId || !ownerId) return;
```
IhorNesterenko99: pls move ownerId and respondentId check after the dispatch(getEvents({ appletId, respondentId })), respondentId can be undefined inside of getEvents
calendar can be general (without respondentId) and individual (with respondentId)

kdanilenka32: fixed

IhorNesterenko99 (approved):

alex-i-e (approved):

PR #1231, opened 2023-11-27
Fix/m2 1255 builder  applet not found

muhurov:


src/shared/components/AppletNotFoundPopup/AppletNotFoundPopup.tsx
```diff
@@ -1,26 +1,34 @@
...
+  const { pathname } = useLocation();
 
-  return (
+  const isBuilder = checkIfAppletUrlPassed(pathname);
+
+  return appletNotFoundPopupVisible ? (
     <Modal
       open={appletNotFoundPopupVisible}
-      onClose={handleClose}
+      onClose={isBuilder ? falseReturnFunc : handleClose}
```
IhorNesterenko99, line 22: what will happen if the user clicks outside the modal and initiates on close?

src/shared/components/Modal/Modal.types.ts
```diff
@@ -40,5 +40,6 @@ export type ModalProps = {
   hasActions?: boolean;
   submitBtnTooltip?: string;
   onTransitionEntered?: (node: HTMLElement, isAppearing: boolean) => void;
+  hasCross?: boolean;
```
IhorNesterenko99:
```suggestion
  hasCloseIcon?: boolean;
```

src/shared/components/AppletNotFoundPopup/AppletNotFoundPopup.tsx
muhurov: nothing, because onClose func does nothing, so the modal will still be visible

src/shared/components/Modal/Modal.types.ts
muhurov: done

IhorNesterenko99 (approved):

src/shared/components/AppletNotFoundPopup/AppletNotFoundPopup.tsx
```diff
@@ -1,26 +1,34 @@
...
 
 export const AppletNotFoundPopup = () => {
   const { t } = useTranslation('app');
   const { appletNotFoundPopupVisible, handleClose, handleSubmit } = useNotFoundPopup();
+  const { pathname } = useLocation();
 
-  return (
+  const isBuilder = checkIfAppletUrlPassed(pathname);
+
+  return appletNotFoundPopupVisible ? (
```
alex-i-e: what about early return:
`if (!appletNotFoundPopupVisible) return null;`
?

alex-i-e (approved):

src/shared/components/AppletNotFoundPopup/AppletNotFoundPopup.tsx
muhurov: fixed

PR #1244, opened 2023-11-29
Fix/M2-4242-media-items-are-not-exported

alex-i-e:


src/shared/types/answer.ts
```diff
@@ -141,7 +141,15 @@ export type DecryptedNumberSelectionAnswer = AdditionalTextType & {
 };
 
 export type DecryptedMediaAnswer = AdditionalTextType & {
-  value: string;
+  value:
+    | string
+    | {
```
IhorNesterenko99: maybe it's good idea to create separate type for this for better understanding?

IhorNesterenko99 (approved):

muhurov (approved):

PR #1275, opened 2023-12-07
M2-4308: Test coverage for Schedule (Calendar) utils

IhorNesterenko99:


src/modules/Dashboard/state/CalendarEvents/CalendarEvents.utils.test.ts
```diff
@@ -1,17 +1,1124 @@
...
+    currentYear: 2023,
+    oneTimeCompletion: false,
+    accessBeforeSchedule: true,
+    timerType: TimerType.Timer,
+    timer: 60,
+    notification: null,
+  };
+  const expected1 = [
+    {
+      activityOrFlowId: '123',
```
alex-i-e: Could you please add a variable with shared properties for `expected1, ..., expected8` ? It will show exact the same the shape of differences which you did for `props1,... props8`

alex-i-e (approved):

muhurov (approved):

PR #1286, opened 2023-12-11
M2-4359: dupl-popup

kdanilenka32:


alex-i-e (changes requested):
Please add more tests to cover the cases when `Applet already exists` and useEffect logic

alex-i-e (approved):

IhorNesterenko99 (approved):

PR #1310, opened 2023-12-18
M2-4415: Review tab: pressing on a timestamp displays the end of responses at the current timestamp if the previously opened timestamp was scrolled down till the end

nmialik:


IhorNesterenko99 (approved):

alex-i-e (approved):
could you point where was the fix for the scrolling when new timestamp was changed?

PR #1321, opened 2023-12-20
M2-4457: Unit tests for LinkForm

nmialik:


src/modules/Dashboard/features/Applet/AddUser/LinkGenerator/LinkForm/LinkForm.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+const setInviteLink = jest.fn();
+const getProps = (requireLogin = false) => ({
+  setInviteLink,
+  inviteLink: {
+    link,
+    requireLogin,
+  },
+});
+
+Object.assign(navigator, {
```
alex-i-e, line 24: Please add the logic to revert `navigator` Object to its initial value after current test (beforeAll/afterAll). Otherwise it could affect other tests where `navigator` used as well.

alex-i-e (changes requested):
left the comment

src/modules/Dashboard/features/Applet/AddUser/LinkGenerator/LinkGenerator.types.ts
```diff
@@ -4,6 +4,6 @@ export type InviteLink = {
 };
 
 export type LinkGeneratorProps = {
-  setInviteLink: (val: InviteLink | null) => void;
+  setInviteLink: (link: InviteLink | null) => void;
```
IhorNesterenko99, line 7: Maybe remove possibility to invite link be null, because of this check?
image
So we don't render LinkForm component if inviteLink is undefined or null.

src/modules/Dashboard/features/Applet/AddUser/LinkGenerator/LinkGenerator.types.ts
```diff
@@ -4,6 +4,6 @@ export type InviteLink = {
...
   inviteLink: InviteLink | null;
```
IhorNesterenko99, line 8: Maybe remove possibility to get value of invite link as null, because of this check?
image
So we don't render LinkForm component if inviteLink is undefined or null.

alex-i-e (approved):

src/modules/Dashboard/features/Applet/AddUser/LinkGenerator/LinkGenerator.types.ts
nmialik: we set setInviteLink(null) in child component

IhorNesterenko99 (approved):

PR #1385, opened 2024-01-16
M2-4596: Shell Account - new "Actions" Menu

IhorNesterenko99:


src/shared/components/ActionsMenu/ActionsMenu.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+const clickOpenMenuButton = () => {
+  const button = screen.getByTestId('test-actions-dots');
+  fireEvent.click(button);
+};
+
+describe('ActionsMenu component', () => {
+  test('renders the ActionsMenu button', () => {
+    renderMenu();
+    const button = screen.getByTestId('test-actions-dots');
+    expect(button).toBeInTheDocument();
```
muhurov, line 20: It doesn't make sense to test being in the document because getBy* methods will throw an error if the corresponding element is not in the document. It can be replaced for example by 'toBeVisible'

muhurov (approved):

PR #1389, opened 2024-01-16
M2-4722: Subscales: hidden item is taken into account for average calculation

nmialik:


muhurov (approved):

src/shared/utils/exportData/getSubscales.ts
```diff
@@ -48,8 +48,17 @@ export const calcScores = <T>(
   subscalesObject: Record<string, ActivitySettingsSubscale>,
   result: { [key: string]: { score: number; optionText: string } },
 ): { [key: string]: { score: number; optionText: string } } => {
-  const sumScore = data.items.reduce((acc, item) => {
-    if (!item?.type) {
+  const subscaleItems = data.items.map((item) => {
+    const isHidden = activityItems[item.name].activityItem.isHidden;
```
IhorNesterenko99: Could you pls integrate isHidden to already available mappers to avoid adding one more mapper and to reduce amount of operations?

nmialik: this data has not been mapped before

IhorNesterenko99: not before
let's move inside these mappers:
image

image

nmialik: fixed

IhorNesterenko99 (approved):

PR #1435, opened 2024-01-25
M2-4859: Wrong item validation after deleting and re-adding an activity

IhorNesterenko99:


src/modules/Builder/features/Activities/Activities.tsx
```diff
@@ -102,6 +102,9 @@ export const Activities = () => {
     typeof index === 'number' ? insertActivity(index, newActivity) : appendActivity(newActivity);
 
     if (isNavigationBlocked) return;
+    if (activities?.length === 0) {
```
muhurov, line 105: Maybe it would be better to move this code to useEffect?

muhurov (approved):

nmialik (approved):

PR #1449, opened 2024-01-30
M2-5028: added debugMode for qa purpose

alex-i-e:


src/shared/utils/logger.ts
```diff
@@ -75,3 +77,8 @@ export const sendLogFile = async ({
     console.warn('Error while file logging:', error);
   }
 };
+
+export const checkIfShouldLogging = () =>
```
IhorNesterenko99, line 81: I think the function can be replaced with the variable 'ifShouldLogging'

IhorNesterenko99 (approved):

PR #1453, opened 2024-01-30
M2-5013: [Admin Panel] Scores&Reports: the "if" dropdown should be without the arrow in the Score Condition section because it is disabled

alex-i-e:


src/shared/components/FormComponents/SelectController/SelectController.tsx
```diff
@@ -165,8 +166,18 @@ export const SelectController = <T extends FieldValues>({
...
-          IconComponent: (props) => <Svg className={props.className} id="navigate-down" />,
+          IconComponent: shouldSkipIcon
+            ? undefined
+            : (props) => <Svg className={props.className} id="navigate-down" />,
           ...SelectProps,
+          ...(shouldSkipIcon && {
+            inputProps: {
+              sx: {
+                pr: '1.2rem !important',
+                minWidth: '94% !important',
```
muhurov, line 177: have you tried increase specificy? like
`inputProps: {
  sx: {
    '&&': { pr: ..., minWidth: ... }
  }
}`
?

IhorNesterenko99 (approved):

src/shared/components/FormComponents/SelectController/SelectController.tsx
alex-i-e: I've checked even with '&&' specified version without '!Importtant' does not work correctly with ellipsis

PR #1472, opened 2024-02-06
M2-4835: Linting rules for `mindlogger-admin`

sultanofcardio:
resolves: [M2-4835](https://mindlogger.atlassian.net/browse/M2-4835)

This PR creates a GitHub Action to run linting tasks on pull requests submitted to the repository. As part of the ticket, I took the opportunity to update the ESlint rules on the project. These updates have raised a number of linting issues, but **won't block** the merging of pull requests. While I've addressed some of the smaller items (and any error ones), I expect that these will be addressed in the normal course of feature development so that we can eventually require linting checks to pass in order to merge PRs.

Here's a summary of what changed:

## Conflicting ESLint configurations

There are currently two ESLint configurations in the project:
- An `eslintConfig` property in the `package.json` file
- A `.eslintrc` file

[ESLint documentation](https://eslint.org/docs/latest/use/configure/configuration-files) says that only one of these is read, and the `.eslintrc` takes precedence. Thus, I've removed the ignored configuration since it wasn't being used.

## Prettier Code Formatting

Prettier has been configured in the project as the code formatter of choice, but it is currently conflicting with a number of code formatting ESLint rules. The prettier ESLint plugin disables these rules if it is placed as an inherited configuration at the end of the `extends` array, so I have made this change. The prettier ESLint plugin also enables Prettier as an ESLint rule, so that code formatting issues become errors. This should be okay, since Prettier fixes them for you.

In addition, I've disabled/removed some other code formatting related rules that aren't explicitly handled by the Prettier plugin. These include:
- `padding-line-between-statements`
- `no-extra-semi`
- `jsx-quotes` - I've moved this one to the Prettier config with `jsxSingleQuote`

## Opinionated changes

There are a number of other changes I've made that I think will allow for improved code quality. These are:
- Turning on the `react-hooks/exhaustive-deps` rule
- Adding the `unused-imports` ESLint plugin
- Turning on the `@typescript-eslint/no-non-null-assertion` rule

[M2-4835]: https://mindlogger.atlassian.net/browse/M2-4835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
+name: CI
+on: [pull_request]
+
+env:
+  NODE_VERSION: 20.11.0
```
alex-i-e: it should be mentioned that all admin panel team works with nodejs v.18 . Maybe it must be specified in package.json and here as well to reduce possible errors relating to different versions of nodejs

.eslintrc
```diff
@@ -1,51 +1,51 @@
...
-                  "@(redux|routes|resources|api|i18n)",
-              "group": "internal"
-            }
-          ],
-          "pathGroupsExcludedImportTypes": ["internal"],
-          "newlines-between": "always"
+      {
+        "groups": [["external", "builtin"], "internal", ["sibling", "parent"], "index"],
+        "pathGroups": [
+          {
```
IhorNesterenko99, line 29: I believe that the empty line between the React import and other imports from libraries, as well as the alphabetized ordering of imports, may be considered excessive. Would you be able to provide some insight into the reasoning behind these choices?

.eslintrc
```diff
@@ -1,51 +1,51 @@
...
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true
         }
+      }
     ],
     "import/no-cycle": "error",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "off",
+    "react-hooks/exhaustive-deps": "warn",
```
IhorNesterenko99, line 46: In numerous cases, not all arguments are necessary to be included in the dependencies of the useEffect hook. Implementing this rule might result in an abundance of unexpected warnings, which could be perceived as excessive.

.prettierrc
```diff
@@ -1,6 +1,13 @@
 {
-  "semi": true,
-  "trailingComma": "all",
+  "arrowParens": "avoid",
```
IhorNesterenko99: My recommendation is to utilize "arrowParens": "always".
Here's the rationale from the Prettier documentation:
"At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option."

.eslintrc
```diff
@@ -72,26 +72,20 @@
...
-      {
-        "ignoreRestSiblings": true
-      }
+      { "vars": "all", "varsIgnorePattern": "^_", "args": "after-used", "argsIgnorePattern": "^_" }
     ],
-    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-non-null-assertion": "warn",
     "@typescript-eslint/no-explicit-any": "warn",
     "@typescript-eslint/no-duplicate-enum-values": "warn",
-    "padding-line-between-statements": [
```
IhorNesterenko99, line 92: To enhance code readability, it's commonly suggested to include a blank line between your code and a return statement. Implementing this practice can contribute to better code understanding.

.prettierrc
```diff
@@ -1,6 +1,13 @@
...
+  "bracketSameLine": true,
```
IhorNesterenko99: Could we consider keeping "bracketSameLine" as false to maintain better code readability?

.prettierrc
```diff
@@ -1,6 +1,13 @@
...
+  "bracketSpacing": true,
   "singleQuote": true,
-  "printWidth": 100
-}
+  "jsxSingleQuote": false,
+  "trailingComma": "all",
+  "semi": true,
+  "printWidth": 120,
```
alex-i-e: Could you explain why did you extend width of line? It leads to changes almost all the files and leads to investigate more time to figure out who did the initial changes.

.prettierrc
```diff
@@ -1,6 +1,13 @@
...
+  "tabWidth": 2,
+  "useTabs": false,
+  "endOfLine": "lf"
```
alex-i-e, line 12: `endOfLine` uses `lf` by default , it seems no need to add this rule as well as `tabWidth`, `useTabs`, `semi`, `trailingComma`

.eslintrc
alex-i-e: It leads to changes almost all the files and leads to investigate more time to figure out who did the initial changes and why.

.eslintrc
alex-i-e: It will lead to add exceptional cases when we don't need to add dependencies as deps hooks because of existed business logic. For example, we need to add some useEffect on mount / unmount and to highlight it we don't need to add dependencies there.

.github/workflows/ci.yml
sultanofcardio: Thanks for clearing that up. I'll revise the node version

.eslintrc
sultanofcardio: I took these rules from the `mindlogger-app-refactor` repo.

https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/a5c8f472cfefe2c5db6ccd0d02efafe73381d8d8/.eslintrc.js#L42-L46

Part of this effort is to bring the code styles in line with each other, to lower the cognitive load of switching between repos. I don't have a preference either way because the tools will auto-format the files which means we won't have to think about it.

.eslintrc
sultanofcardio: I'm not sure I understand the comment. The cases where it is necessary to exclude dependencies from the array should be the exception, and in those cases the rule can be disabled. Not having warnings at all increases the likelihood that we miss cases where omitting dependencies isn't the intention.

I don't see the high number of warning as a problem, because we won't block PRs for warnings. It is simply a signal to the developer (and reviewer) that you should pay attention here

.eslintrc
sultanofcardio: Or maybe it was the `mindlogger-web-refactor`, I don't remember but they both use it

.prettierrc
sultanofcardio: You're right. I must've accidentally overwritten this one with a value from the app repo. I think keeping parentheses is the right move. Thanks for flagging this

.eslintrc
sultanofcardio: I removed this rule because ESLint deprecated and removed it to a separate plugin in the next version. The reason they removed it is because it has to do with opinionated code formatting, for which we are using prettier.

My personal preference is for keeping this so I'll revert for now and see what others think

.prettierrc
sultanofcardio: Sure

.prettierrc
sultanofcardio: Yes I did it to keep consistency with the backend and the web app

.prettierrc
sultanofcardio: Even though they're the default, they serve to indicate to the developer what's being used without needing to check the documentation. The web app also specifies these so it keep them consistent for folks working between repositories

.prettierrc
alex-i-e: It's ok, but could you please revert printWidth?

IhorNesterenko99:
I've noticed that the unit tests are failing after the changes made in the pull request. Could you please take a look and correct any issues with the unit tests?

Additionally, there are some conflicts that need to be resolved before this pull request can be merged. Could you please review and resolve the conflicts accordingly?

.husky/prepare-commit-msg
IhorNesterenko99, line 1: I just wanted to bring to your attention that we have some common rules for committing code changes in our project. These rules include including the name of the associated ticket in Jira and adhering to specific branch naming conventions:
- Type: feature, fix, tests, release
- Ticket name in Jira

These guidelines help maintain consistency and clarity across our codebase, reducing any discrepancies between developers when creating branches and writing commit messages.

Could you please share your reasoning behind the decision to delete this code?

sultanofcardio:
> I've noticed that the unit tests are failing after the changes made in the pull request. Could you please take a look and correct any issues with the unit tests?
>
> Additionally, there are some conflicts that need to be resolved before this pull request can be merged. Could you please review and resolve the conflicts accordingly?

@IhorNesterenko99 Yes, I will address these. I initially believed that the tests were failing because of [this commit](https://github.com/ChildMindInstitute/mindlogger-admin/commit/db0ae5d230d3c887d9b32b2d5aebe6e7c60073ec), but after rebasing again I realised they are still failing.

.husky/prepare-commit-msg
sultanofcardio: This is a mistake. I deleted these because the pre-commit hook makes rebasing very tedious, but I didn't intend to push this commit to GitHub. I will remove this commit 👍

.eslintrc
sultanofcardio: As it turns out, alphabetized imports break the test files because they currently depend on import order, so I've reverted this for now

IhorNesterenko99:
I wanted to bring to your attention that our Jenkins continuous integration is currently failing. Here's a screenshot for reference: 
image
It might be a good idea to reach out to our DevOps lead, Raman Siauko, to investigate what might have caused this issue.

.eslintrc
alex-i-e: Thank you, and could you please revert adding empty line after `react` imports (by keeping it in group of `external` imports on the top of the files)?

alex-i-e (approved):
Perfect! You've added meaningful and valuable lint rules for ongoing on project, some of them will point on places that should be fixed in future. I appreciate your input and contribution to improving quality!

IhorNesterenko99 (approved):

.eslintrc
farmerpaul: Missing dependencies can be incredibly annoying to debug, and I agree with @sultanofcardio that this rule is essential, even if enabled just as a warning for now (though if it were up to me, I'd want it to issue an error).

If you're certain you want to exclude a variable from the dependencies, rather than disable this rule entirely via a `// es-disable-…` comment, you can opt to do a more surgical approach by placing that variable in a ref, for example:

```tsx
export const MyComponent: React.FC<MyComponentProps> = ({ sideEffect }) => {
  const [count, setCount] = useState(0);
  const sideEffectRef = useRef(sideEffect); // 👈 Place unchanging variable in a ref
  
  const handleClick = useCallback(() => {
    setCount(count + 1);
    sideEffectRef.current();
  }, [count]); // 👈 Now we can preserve the exhaustive deps rule here
  
  return <h1 onClick={handleClick}>It Worked: {count)</h1>;
}；
```

If running an effect on mount/unmount is common, one way to reduce instances of `// es-disable-…` is to define a custom hook that is explicit in its purpose, for example:

```tsx
export const useOnceEffect = (callback: React.EffectCallback) => {
  useEffect(callback, []);
};
```

PR #1486, opened 2024-02-09
M2-5096: convert remaining popups to banners

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-5096](https://mindlogger.atlassian.net/browse/M2-5096) (parent ticket [M2-4823](https://mindlogger.atlassian.net/browse/M2-4823))

This PR converts the remaining popups (as selected from the [popups audit](https://mindlogger.atlassian.net/browse/M2-4968)) into banners. In doing so, added support for passing props to banners (as not all types of banners have static text). Also, some minor copy changes were made to error popups (which continue to be popups, as per audit).

The PR also introduces a utility for testing the presence of a banner (`expectBanner`) to make it easier to test for the presence of a banner (see [note below](#expect-banner)).

It will likely be easier to review this PR [commit-by-commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1486/commits). (FYI: Commit messages reference table rows by item # in the [popups audit](https://mindlogger.atlassian.net/browse/M2-4968).)

### 📸 Screenshots

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/eaa8a151-4411-4f1a-9c73-f2bbebd54d63

### 🪤 Peer Testing

Follow testing steps below to confirm behaviour:

| Testing steps | Expected banner text | Duration |
| - | - | - |
| Edit applet > Applet settings > Report configuration > Save > Save anyway | `Changes saved` | 3.5s |
| Edit applet > Edit About Applet Page > Upload Image > Select image 30+ MB image | `File is too big. Please upload a file less than 25 MB.` | 5s |
| Edit applet > Edit About Applet Page > Upload Image > Select BMP image | `Image format must be JPEG or PNG.` | 5s |
| Edit applet > Edit About Applet Page > Upload Audio > Select OGG audio | `Incorrect audio format` | 5s |
| Edit applet > Edit About Applet Page > Upload Audio > Select video 200+ MB video | `File is too big. Please upload a file less than 150 MB.` | 5s |
| Edit applet > Applet settings > Data retention > Save | `Changes saved` | 3.5s |
| Dashboard > Delete an applet > Enter password > Delete | `Applet was deleted` | 3.5s |
| Dashboard > Transfer ownership > Enter another user email > Confirm | `Your request was successfully sent to <email>. Please wait for receiver to accept your request.` | 7s |
| Dashboard > Choose "View Users" for an applet with multiple managers > Managers > Edit another manager's access > Save | `Changes saved` | 3.5s |
| Dashboard > Choose "View Users" for an applet with respondents > Respondents > Edit a respondent > Save | `Changes saved` | 3.5s |

### ✏️ Notes

<a id="testing"></a>
#### Missing ability to test specific popup
⚠️ Although I've converted the below popups to banners as well in this PR (as per requirements), I have not yet figured out **how to test these two**. The code lives in `PublishConcealAppletPopup.tsx`. If someone at ScienceSoft would be able to provide insight into this, that would be great!

| Testing steps | Expected banner text | Duration |
| - | - | - |
| ❓ | `The applet <applet name> has been concealed successfully` | Persistent |
| ❓ | `The applet <applet name> has been published successfully`* | Persistent |

* Note that this second popup is **not** the same as the `Applet <applet name> has been successfully saved and published.` popup, which has _not_ been converted to a banner as part of this PR. That one is expected to be done as part of [M2-5089](https://mindlogger.atlassian.net/browse/M2-5089).

<a id="expect-banner"></a>
#### `expectBanner` utility
In the tests, my hope was to test for the presence of banners by matching against the rendered text found within them, rather than using Redux state directly. However, banners aren't rendered in our tests unless the `<Banners />` component is included somewhere. Hypothetically, `<Banners />` could be embedded within our `renderWithProviders` and `renderHookWithProviders` utilities, however I tried that, and it caused circular dependency issues that surfaced either during Jest execution, or the app, or even both. Rather than spend an undue amount of time attempting to resolve those, I opted to stick with comparing against Redux state directly to check for presence of banners.

[M2-4823]: https://mindlogger.atlassian.net/browse/M2-4823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[M2-5089]: https://mindlogger.atlassian.net/browse/M2-5089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/shared/utils/expectBanner.ts
sabrichu, line 1: Nice, thanks for adding!

src/shared/state/Banners/Banners.schema.ts
```diff
@@ -1,13 +1,16 @@
-import { AppletWithoutChangesBanner } from 'modules/Builder/components/Banners';
-import { VersionWarningBanner } from 'shared/components/Banners/VersionWarningBanner';
+import { BannerProps } from 'shared/components/Banners/Banner/Banner.types';
 
-export const BannerComponents = {
+export enum BannerType {
```
sabrichu, line 3: Not a blocking comment, but is the expectation that with any new banner added to the site, we would add a new type here? I would expect these banner types to be much more generic (`SuccessBanner`, `WarningBanner`, `ErrorBanner`, etc) with the children supplied via `bannerProps`, and not being much more opinionated than that...

...but then again, those wrapper components also wouldn't be doing much more than pre-configuring some values for `severity` `duration`. Anyway, just food for thought. My PRs were going to reuse the `SaveSuccessBanner` for save success and duplicate success

sabrichu (commented):
Still looking through it all but wowza, great work! This was a lot to comb through/it's not easy maneuvering a mix of legacy approaches and components

Have an initial comment...

src/shared/components/Uploader/Uploader.tsx
```diff
@@ -197,14 +201,51 @@ export const Uploader = ({
 
   const placeholderImgId = isPrimaryUiType ? 'img-filled' : 'img-outlined';
   const placeholderImgSize = isPrimaryUiType ? 32 : 24;
-  const hasSizeError = error === UploadFileError.Size;
-  const hasFormatError = error === UploadFileError.Format;
-  const hasDimensionsError = error === UploadFileError.Dimensions;
+  const [hasSizeError, hasFormatError, hasDimensionsError] = useMemo(
```
sabrichu: Curious why `useMemo` is needed here

src/shared/state/Banners/Banners.schema.ts
farmerpaul: I've also been reusing `SaveSuccessBanner` for most success situations. You're right that it is basically just props that are changing (with `children` being one of those props), but I suppose the benefit of having distinct components is cases where the content is more complex and that banner is called from multiple locations, just to offer a level of reusability. (Which, as I say that, it now occurs to me I probably should have done with the "transfer ownership" banner! Maybe I'll make a change there…)

src/shared/components/Uploader/Uploader.tsx
farmerpaul: Good question. Neither of my `useMemo`s here are particularly warranted. I think I was trying to troubleshoot reactive behaviour during testing and was misguided into thinking this would solve it. I'll get rid of these.

src/shared/state/Banners/Banners.schema.ts
sabrichu: Got it, that seems reasonable!

src/shared/components/Uploader/Uploader.tsx
sabrichu: I have scattered many a `setTimeout` for the exact same reason

src/modules/Builder/features/ReportConfigSetting/ReportConfigSetting.tsx
```diff
@@ -382,10 +382,19 @@ export const ReportConfigSetting = ({
   };
 
   useEffect(() => {
```
alex-i-e, line 384: I guess we don't need local state `successBannerVisible/setSuccessBannerVisible` anymore, since you could retrieve this info from `banners` store (via check if `SaveSuccessBanner` is in the list). Moreover, if you move `dispatch(banners.actions.addBanner({key: 'SaveSuccessBanner', ...` into a callback you could get rid of such places where used the code
```
 () => {
      setSuccessBannerVisible(true);
    },
```
and at the same time no need `setSuccessBannerVisible(false);`

src/modules/Dashboard/features/Applet/Popups/DeletePopup/DeletePopup.tsx
```diff
@@ -56,6 +56,21 @@ export const DeletePopup = ({ onCloseCallback, 'data-testid': dataTestid }: Dele
     deletePopupClose();
   };
 
+  useEffect(() => {
+    if (activeModal === Modals.Confirmation) {
```
alex-i-e: let's get rid of `useEffect` that could be replaced with a callback. it's a way more intuitive and much improve readability of the code, simplify writing the tests and debugging.

alex-i-e (changes requested):
...left the comments in the code

src/modules/Dashboard/features/Applet/Popups/TransferOwnershipPopup/TransferOwnershipPopup.tsx
```diff
@@ -37,16 +33,38 @@ export const TransferOwnershipPopup = () => {
   };
 
   useEffect(() => {
-    if (!emailTransfered) return;
+    if (!emailTransferred) return;
```
alex-i-e: if it'll not take a long time let's get rid of `useEffect` that could be replaced with a callback as I guess into the places where triggered `setEmailTransferred(true)`

src/shared/components/Banners/Banners.tsx
```diff
@@ -1,21 +1,29 @@
...
-      {data.banners.map(({ key }) => {
+      {data.banners.map(({ key, bannerProps }) => {
         const BannerComponent = BannerComponents[key];
 
         return (
           <Collapse key={key}>
-            <BannerComponent onClose={() => dispatch(banners.actions.removeBanner({ key }))} />
+            <BannerComponent
+              {...bannerProps}
+              onClose={() => {
```
alex-i-e: As a common rule, we use an anonymous function into the html as a callback handler for cases when it includes one callback/action. In this case I would move the logic into a separate function. It reduces html code and improves readability. But feel free discuss it of course.

src/shared/components/Uploader/Uploader.tsx
```diff
@@ -197,14 +201,51 @@ export const Uploader = ({
...
+      error === UploadFileError.Dimensions,
+    ],
+    [error],
+  );
   const hasImageError = hasSizeError || hasFormatError || hasDimensionsError;
   const spinnerUiType = isPrimaryUiType ? SpinnerUiType.Primary : SpinnerUiType.Secondary;
 
   const fileName = imageField?.split('/').at(-1) || image?.name || '';
 
+  useEffect(() => {
```
alex-i-e: Here, as well let's get rid of `useEffect` that could be replaced with a callback into the places where `error` set to `Size` and `Format` accordingly (in row 110 and 111) , I suppose, when the condition `isPrimaryUiType === false`

src/shared/features/AppletSettings/TransferOwnershipSetting/TransferOwnershipSetting.tsx
```diff
@@ -1,39 +1,60 @@
...
-    setTransferOwnershipPopupVisible(false);
-    setEmailTransfered('');
+  const handleSuccess = () => {
+    setEmailTransferred('');
     transferOwnershipRef.current?.resetEmail();
   };
 
   useEffect(() => {
-    if (!emailTransfered) return;
+    if (!emailTransferred) return;
```
alex-i-e: if it'll not take a long time let's get rid of `useEffect` that could be replaced with a callback as I guess into the places where triggered `setEmailTransferred(true)`

alex-i-e:
Btw 👍 , nice to see good test coverage, thank you!

src/modules/Dashboard/features/Applet/Popups/DeletePopup/DeletePopup.tsx
```diff
@@ -90,18 +105,8 @@ export const DeletePopup = ({ onCloseCallback, 'data-testid': dataTestid }: Dele
         </Modal>
       );
     case Modals.Confirmation:
```
IhorNesterenko99: Could you please remove this case from switch statement and from Modals enum?

farmerpaul:
Appreciate your reviews @alex-i-e and @IhorNesterenko99! I agree with all your suggestions and have implemented them, so please take another look as soon as you have a moment.

I'm also curious if either of you know the answer to my [question above](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1486#testing) about testability of the "applet concealed/published" functionality?

src/modules/Builder/features/ReportConfigSetting/ReportConfigSetting.tsx
```diff
@@ -104,7 +103,7 @@ export const ReportConfigSetting = ({
   const { execute: postReportConfig } = useAsync(
     postReportConfigApi,
     () => {
```
alex-i-e: 💅: here and below no need an anonymous function wrapper, right?
```
() => {
   handleSuccess();
}
```
use just the reference `handleSuccess` on the function

farmerpaul: Agreed, and it's what I was going to do, but it required a little rearranging of code. Event handler now has to be defined early, separately from where the other event handlers live. I'm fine with that if you are, so pushing this change!

src/modules/Builder/components/ToggleItemContainer/ToggleItemContainer.styles.ts
```diff
@@ -28,15 +28,18 @@ export const StylesTitleWrapper = styled(StyledFlexTopCenter, shouldForwardProp)
     open,
```
IhorNesterenko99: It appears that the commits from our previous work, which have already been merged into the development branch, are now showing in your pull request. Could you please double-check?

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -76,53 +88,46 @@ export const EditRespondentPopup = ({
...
       open={popupVisible}
       onClose={onCloseHandler}
-      onSubmit={isSuccessVisible ? onCloseHandler : handleSubmit(submitForm)}
+      onSubmit={handleSubmit(submitForm)}
       disabledSubmit={isLoading}
       title={t('editRespondent')}
-      buttonText={t(isSuccessVisible ? 'ok' : 'save')}
-      hasSecondBtn={!isSuccessVisible}
+      buttonText={t('save')}
+      hasSecondBtn={true}
```
IhorNesterenko99:
```suggestion
      hasSecondBtn
```

src/shared/components/Banners/Banners.tsx
```diff
@@ -1,21 +1,32 @@
 import { Collapse } from '@mui/material';
 import { TransitionGroup } from 'react-transition-group';
+import { useCallback } from 'react';
```
IhorNesterenko99: Could you please relocate the import from React to the top, placing it before the other imports?

src/shared/components/Banners/Banners.tsx
```diff
@@ -1,21 +1,32 @@
...
+import { BannerPayload, banners } from 'shared/state/Banners';
 import { useAppDispatch } from 'redux/store';
 
+import { BannerComponents } from './Banners.const';
+
 export const Banners = () => {
   const dispatch = useAppDispatch();
   const data = banners.useData();
 
+  const handleClose = useCallback(
```
IhorNesterenko99: Would it be beneficial to use useCallback here? What advantages would this decision provide?

src/shared/components/Banners/SaveSuccessBanner/SaveSuccessBanner.tsx
```diff
@@ -0,0 +1,13 @@
+import { useTranslation } from 'react-i18next';
+
+import { Banner, BannerProps } from '../Banner';
+
+export const SaveSuccessBanner = ({ children, ...props }: BannerProps) => {
+  const { t } = useTranslation('app');
+
+  return (
+    <Banner duration={3500} severity="success" data-testid="save-success-banner" {...props}>
```
IhorNesterenko99: Could you please create a variable for the duration and place it in the SaveSuccessBanner.const.ts file?

src/shared/components/Banners/TransferOwnershipSuccessBanner/TransferOwnershipSuccessBanner.tsx
```diff
@@ -0,0 +1,22 @@
+import { Trans } from 'react-i18next';
+
+import { Banner, BannerProps } from '../Banner';
+
+export const TransferOwnershipSuccessBanner = ({ children, email, ...props }: BannerProps) => (
+  <Banner
+    duration={7000}
```
IhorNesterenko99: Could you please create a variable for the duration and place it in the TransferOwnershipSuccessBanner.const.ts file?

src/shared/components/FormComponents/EditorController/EditorController.tsx
```diff
@@ -19,53 +19,68 @@ export const EditorController = <T extends FieldValues>({
   withDebounce = false,
   'data-testid': dataTestid,
 }: EditorControllerProps<T>) => {
+  const dispatch = useAppDispatch();
   const editorRef = useRef<ExposeParam>();
-  const [fileSizeExceeded, setFileSizeExceeded] = useState<number | null>(null);
-  const [incorrectFileFormat, setIncorrectFileFormat] = useState<MediaType | null>(null);
+
+  const onFileSizeExceeded = useCallback(
```
IhorNesterenko99, line 25: I'm curious, what are the benefits of using useCallback here and in the next function?

IhorNesterenko99:
Hi @farmerpaul!
Thank you for taking our comments into consideration. 
Currently, the ability to Publish/Conceal requires the user to have a super-admin role. However, development of this functionality has been paused. Unfortunately, this means you won't be able to test it at this time.
Please take a look at the corresponding User Story:
https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/189104593/AUS-4.1.4.11+-+Applet+settings+tab+of+the+Applet+page+Publish+Conceal+Applet+for+MindLogger+System+Admin+Only

src/modules/Builder/features/ReportConfigSetting/ReportConfigSetting.tsx
alex-i-e: I'm ok with it, thank you!

alex-i-e (approved):
LGTM

src/shared/components/Banners/Banners.tsx
alex-i-e: Per my understanding, to not rerender `<BannerComponent ... />` which uses it this property. 

But, in this case, to prevent redundant renders of all `<BannerComponent...` from the list, it should have a bit another location and sequence of contexts for the property `handleClose`:

1). why we need to think about `useCallback/useMemo/memo` when we don't have a huge calculations/bad performance in descendants or in current component? I suppose it looks like overengineering here and in other similar places.

2.1). it should be outside of the component `Banners` to avoid the generating of new reference of the function each time when `const data = banners.useData();` is updated
```
const handlePureClose = (dispatch: AppDispatch, { key, bannerProps }: BannerPayload) => () => {
  dispatch(banners.actions.removeBanner({ key }));
  bannerProps?.onClose?.();
}
```
2.2). not to generate new anonymous function when banners data is updated:
```
<BannerComponent {...bannerProps} onClose={handlePureClose({ key, bannerProps })} />
```

src/shared/components/FormComponents/EditorController/EditorController.tsx
alex-i-e: I guess to not re-render the huge component `<Editor` (use the old reference in properties when parent component is re-rendered). 
I think it makes sense, if it has a performance issues. Otherwise, I would consider it as a overengineering.

src/modules/Builder/components/ToggleItemContainer/ToggleItemContainer.styles.ts
farmerpaul: How curious, yes, I'm seeing that above:

Screenshot 2024-02-15 at 10 21 00 AM

I have no idea how this happened; all I've been doing is rebasing on `develop`. I'll fix it, thanks for the callout.

IhorNesterenko99:
Additionally, I suggest reducing the code size for review by breaking down large tasks into smaller parts. This is something to consider for future pull requests.

PR #1506, opened 2024-02-20
M2-4662: Data vizualisation - make review date default to the last date when any Activity in the Applet was completed.

IhorNesterenko99:
[JIRA ticket](https://mindlogger.atlassian.net/browse/M2-4662)


nmialik (approved):

src/shared/components/DatePicker/DatePicker.tsx
```diff
@@ -150,7 +150,9 @@ export const DatePicker = <T extends FieldValues>({
                 id={id}
                 open={isOpen}
                 anchorEl={anchorEl}
-                onClose={handlePickerClose}
+                onClose={() => {
+                  handlePickerClose(getSelectedDate());
```
alex-i-e: could you please move the logic into a separate function to use it here and in row 195, for ex.:
```
onClose={handlePickerCloseWithSelectedDate}
...
onClick={handleCloseWithSelectedDate}
```

alex-i-e (approved):

PR #1507, opened 2024-02-20
M2-5309: change Banner `data-testid`s

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-5309](https://mindlogger.atlassian.net/browse/M2-5309)

To make Selenium tests easier, it was requested to standardize banners' `data-testid` prop, following this pattern:

| Severity                      | `data-testid`                                 |
| `success` | `success-banner` |
| `info` | `info-banner` |
| `warning` | `warning-banner` |
| `error` | `error-banner` |

In making this change, I had to adjust the `expectBanner` unit testing utility as well, since the customized `data-testid` prop would no longer be available to compare specific banners against. Instead, unit tests now use the banner's `key` for comparison.

### 🪤 Peer Testing

Test existing banner functionality and ensure there is no regressive behaviour.

### DIFF
--- a/src/shared/components/Banners/Banner/Banner.tsx
+++ b/src/shared/components/Banners/Banner/Banner.tsx
@@ -14,7 +14,6 @@ export const Banner = ({
   onClose,
   hasCloseButton = !!onClose,
   severity = 'success',
-  'data-testid': dataTestid = `${severity}-banner`,
 }: BannerProps) => {
   let timeoutId: NodeJS.Timeout | undefined;
   const [isHovering, setIsHovering] = useState(false);
@@ -47,7 +46,7 @@ export const Banner = ({
       onMouseEnter={() => setIsHovering(true)}
       onMouseLeave={() => setIsHovering(false)}
       severity={severity}
-      data-testid={dataTestid}
+      data-testid={`${severity}-banner`}
     >
       {children}
     </Alert>
--- a/src/shared/components/Banners/Banner/Banner.tsx
+++ b/src/shared/components/Banners/Banner/Banner.tsx
@@ -13,8 +13,8 @@ export const Banner = ({
   duration = 5000,
   onClose,
   hasCloseButton = !!onClose,
-  severity,
-  'data-testid': dataTestid,
+  severity = 'success',
+  'data-testid': dataTestid = `${severity}-banner`,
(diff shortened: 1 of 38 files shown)


src/shared/components/Banners/Banner/Banner.tsx
```diff
@@ -13,8 +13,8 @@ export const Banner = ({
   duration = 5000,
   onClose,
   hasCloseButton = !!onClose,
-  severity,
-  'data-testid': dataTestid,
+  severity = 'success',
+  'data-testid': dataTestid = `${severity}-banner`,
```
IhorNesterenko99: I'm wondering if we could simplify the component by removing the 'data-testid' prop and according type for it and using a variable dataTestid instead. What are your thoughts on this change?

alex-i-e: just in case , we may have a several `success` or `error` banners with the same `data-testid`, right?

farmerpaul: Yes, that's possible, good callout. Perhaps this wasn't aware of the author of the JIRA task, as it may be hard to target a specific banner if multiple banner of the same type are shown. I'll seek clarification.

farmerpaul: I thought that it would be prudent to leave the ability to override the default value for the test ID if there was a need to uniquely target a specific banner. Also just confirming the requirements with task author based on [this issue](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1507#discussion_r1497231018) being raised, will report back!

farmerpaul: I checked in with Oleg and he said the possibility of multiple banners with same `data-testid` won't be an issue, he's prepared to handle it (if that even comes up in practice). So I think it's fine to stick with the current implementation.

farmerpaul: Sounds like Oleg is happy to stick with restrictive `data-testid`s as per the task requirements, so I'll remove this extra prop and simplify the code a bit as you suggested.

alex-i-e (approved):

IhorNesterenko99 (approved):

PR #1509, opened 2024-02-21
M2-4897: Blank screen/Inconsistent info text for transfer ownership negative scenarios

IhorNesterenko99:
[JIRA ticket](https://mindlogger.atlassian.net/browse/M2-4897)

[Video](https://app.screencast.com/GAwWIltvwBi8T)


src/redux/store/reducers.ts
```diff
@@ -25,5 +26,6 @@ export const rootReducer = combineReducers({
   reportConfig: reportConfig.slice.reducer,
   themes: themes.slice.reducer,
   users: users.slice.reducer,
-  workspaces: workspaces.slice.reducer,
+  workspaces: workspaces?.slice.reducer,
```
alex-i-e: why did you use the optional chaining here?

src/shared/components/NoPermissionPopup/NoPermissionPopup.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      route,
+      routePath,
+      preloadedState: getPreloadedState(),
+    });
+
+    expect(getByTestId('no-permission-popup')).toBeVisible();
+    expect(screen.getByText('No access to applet')).toBeVisible();
+    expect(screen.getByText('To refresh the data, please click the button below.')).toBeVisible();
+
+    fireEvent.click(screen.getByText('Refresh'));
```
alex-i-e: please use `await userEvent.click...`  if possible here and below

src/shared/hooks/usePermissions.tsx
```diff
@@ -22,7 +23,8 @@ export const usePermissions = (asyncFunc: () => Promise<any> | undefined) => {
 
         if (
           payload?.response?.status === ApiResponseCodes.Forbidden ||
-          payload?.status === ApiResponseCodes.Forbidden
+          payload?.status === ApiResponseCodes.Forbidden ||
+          payload?.[0]?.type === ErrorResponseType.AccessDenied
```
alex-i-e: if the payload can be a list, it's probably better to check an assumption in the array: `typeof payload === 'array' && payload.some(data => data.type === ErrorResponseType.AccessDenied)`

src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
```diff
@@ -307,8 +313,12 @@ export const useSaveAndPublishSetup = (): SaveAndPublishSetup => {
     if (responseStatus === 'loading' && checkIfAppletBeingCreatedOrUpdatedRef.current) {
       setPublishProcessStep(SaveAndPublishSteps.BeingCreated);
     }
-    responseStatus === 'error' && setPublishProcessStep(SaveAndPublishSteps.Failed);
-  }, [responseStatus, responseTypePrefix]);
+    if (isResponseStatusError) {
+      if (hasAccessDeniedError) return;
```
alex-i-e, line 317: Didn't get where is processed the `Access Denied` error ?

IhorNesterenko99: This error is processed in the interceptor:
image

The return here is to avoid appearing two different popups.

src/redux/store/reducers.ts
IhorNesterenko99: The change was made previously due to failing unit tests. I will revert it.

alex-i-e (approved):

PR #1510, opened 2024-02-21
M2-5203: Change password form and route

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-5203](https://mindlogger.atlassian.net/browse/M2-5203)

This PR adds a password recovery page that mirrors the same page in the respondent web app. Prior to this change, users of the admin app were not able to change their password without going to the respondent web app.

### 📸 Screenshots

Here's a screen recording of me going through the password recovery process

https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/ac76ddf7-d59c-4e45-88bd-b3c7abc5ffb7

> [!WARNING]
> The email sent from the API [currently links](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1100) to a `/password-recovery` path in the admin app, but the correct path is `/auth/password-recovery`, so I had to edit the link before going through with it. We will have to update the API to fix this

#### Happy path

- Start up the API and the admin panel
- Navigate to the admin app landing page
- Click the forgot password link
- Enter your email address and click *Send me a reset link*
- Copy the link in your email (Go to http://localhost:8025/ for local testing), and **edit** it to prepend `/auth` to the path
- Enter the new password
  **Expected outcome:** You should be redirected to the login screen and see a success banner displayed for 5 seconds

#### Invalid email/key

Enter each of the following URLs in your browser with the admin app running and confirm that you see an error message saying the link is invalid
- `https://localhost:3000/auth/password-recovery`
- `https://localhost:3000/auth/password-recovery?email=someone%40example.com&key=invalid=key`

Also repeat the test while the backend is not running

image

#### Error changing password

- Repeat the steps from the happy path until you arrive at the password change form
- Turn off the backend API
- Submit the form and confirm that you're seeing an appropriate error message

image

### ✏️ Notes

Be mindful that the email sent from the API [currently links](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1100) to a `/password-recovery` path in the admin app, but the correct path is `/auth/password-recovery`. Be sure to edit the link in the email while testing to correct this


sultanofcardio:
P.S. I'm not sure yet why there are some tests failing but I'd like to get some eyes on this in the meantime

src/shared/components/Banners/Banners.const.ts
```diff
@@ -1,6 +1,7 @@
 import { ComponentType } from 'react';
 
 import { BannerType } from 'redux/modules';
+import {PasswordResetSuccessfulBanner} from 'modules/Auth/features/Login/Banners';
```
ChaconC: Missing code formatting

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (error) {
+    return (
+      <Box display="flex" flex={1} justifyContent="center" alignItems="center" textAlign="center">
+        <StyledBodyLarge variant="body1" margin="16px 0px">
+          <Box>
+            <Trans i18nKey="invalidPasswordResetLink">
+              <StyledLinkBtn
+                style={{ marginBottom: 3, fontSize: 16 }}
+                variant="text"
+                onClick={() => navigate(page.login)}
```
ChaconC: I think this should link to `page.passwordReset`, right?
The text reads "Please click here to reset your password"

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const email = searchParams.get('email');
+
+  const {
+    execute: validateKey,
+    isLoading,
+    error,
+  } = useAsync(
+    async (data: { email: string | null; key: string | null }) =>
+      await recoveryLinkHealthCheckApi(data),
+  );
```
ChaconC: Granted it doesn't affect functionality, if you navigate to a used link this results in a couple of uncaught exceptions

ChaconC (commented):
Couple of minor issues from me but overall works and looks great!

src/shared/components/Banners/Banners.const.ts
sultanofcardio: Ah, thanks for catching. In a few minutes (🤞) we should be able to catch these automatically through a GitHub action

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
sultanofcardio: That's exactly right. I must have messed up and pasted over this somewhere along the way. Thanks Carlos!

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
sultanofcardio: I ended up just calling the function inside a useEffect, and I handled the errors there

ChaconC (approved):

src/modules/Auth/features/RecoverPassword/RecoverForm/RecoverForm.schema.ts
```diff
@@ -0,0 +1,27 @@
...
+    .object({
+      password: yup
+        .string()
+        .required(passwordRequired)
+        .min(ACCOUNT_PASSWORD_MIN_LENGTH, passwordMinLength)
+        .matches(/^(\S+$)/, passwordBlankSpaces),
+      passwordConfirmation: yup
+        .string()
+        .required(passwordConfirmationRequired)
+        .oneOf([yup.ref('password')], passwordsMustMatch),
```
alex-i-e: it's rare case, but anyway an user can be confused the cases when input message should be 'Please confirm your password' instead of 'Your passwords do not match':
image
image

 I would expect to receive the message when Confirm Password is empty: 
image
image
```suggestion
        .oneOf([yup.ref('password'), ''], passwordsMustMatch),
```

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const RecoverPassword = () => {
+  const [searchParams] = useSearchParams();
+
+  const navigate = useNavigate();
+
+  const key = searchParams.get('key');
+  const email = searchParams.get('email');
+
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
alex-i-e, line 20: please consider to use `useAsync`, I believe this will get rid of the local use of `loading` and `error` states

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Box display="flex" flex={1} justifyContent="center" alignItems="center" textAlign="center">
+        <Spinner />
+      </Box>
+    );
+  }
+
+  if (error) {
+    return (
+      <Box display="flex" flex={1} justifyContent="center" alignItems="center" textAlign="center">
+        <StyledBodyLarge variant="body1" margin="16px 0px">
```
alex-i-e: please consider to use `theme.spacing`

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+      <Box display="flex" flex={1} justifyContent="center" alignItems="center" textAlign="center">
+        <Spinner />
+      </Box>
+    );
+  }
+
+  if (error) {
+    return (
+      <Box display="flex" flex={1} justifyContent="center" alignItems="center" textAlign="center">
```
alex-i-e, line 43: Maybe It's time to discuss the styling of components, but so far we stick the usage of `Styled[ComponentName]` as a common approach like you applied in [RecoverForm.styles.ts](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1510/files#diff-285908779dbf07e580c21b701a9afedcceb52cdb7c6d1c809e5c54cdb31cd4ca)
@IhorNesterenko99, please add your thoughts.

src/modules/Auth/features/RecoverPassword/RecoverForm/RecoverForm.schema.ts
sultanofcardio: Good catch Aleksandr. Thanks for the suggestion

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
sultanofcardio: I started out with `useAsync`, but I would still end up having to call the `execute` function inside a `useEffect`. I moved away from it because this approach is simpler

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
sultanofcardio: I opted to use inline styles here because the `RecoverPassword` component is a page, and so far I've only seen the styled component approach used for components in the `features` directory. However, I agree that these styles could be abstracted. Do you have an opinion on where that file could go?

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
sultanofcardio: Thanks for the suggestion. I ended up just removing the margin because it wasn't necessary. I copied this component from the respondent web app repo and modified it

src/modules/Auth/pages/RecoverPassword/RecoverPassword.tsx
alex-i-e: Yeap, make sense in this particular case.

alex-i-e (approved):

PR #1540, opened 2024-03-05
M2-5327: [Admin Panel] Website becomes unresponsive after opening the 'Scores'Reports' tab of PMHSA11+ applet

alex-i-e:
### 📝 Description

🔗 [Jira Ticket M2-5327](https://mindlogger.atlassian.net/browse/M2-5327)

### ✏️ Notes

To resolve the long first rendering, it was introduced skeleton components for section, score and score conditional components. 

The skeleton mode for **sections or scores** (in terms of reports, they are the same entities) is triggered when number of reports exceeds **3**.
The skeleton mode for **score’s conditionals** is triggered when number of conditionals exceeds **2**.

Also to make the inputs more responsive, it was added the debounce for them (in previous task https://mindlogger.atlassian.net/browse/M2-4655 )


src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreCondition/StaticScoreConditionals/StaticScoreConditionals.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ScoreConditionalLogic } from 'redux/modules';
+import { StyledTitleMedium, theme } from 'shared/styles';
+import { getEntityKey } from 'shared/utils';
+
+import { StaticScoreCondition } from '../StaticScoreCondition/StaticScoreCondition';
+
+export const StaticScoreConditionals = ({
+  scoreConditionals,
+}: {
+  scoreConditionals: ScoreConditionalLogic[];
```
IhorNesterenko99: Please move to types.

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/StaticScoreContent/StaticScoreContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { StyledFlexTopStart, StyledTitleMedium, StyledTitleSmall, theme } from 'shared/styles';
+import { ScoreConditionalLogic } from 'redux/modules';
+
+import { StaticScoreConditionals } from '../ScoreCondition/StaticScoreConditionals/StaticScoreConditionals';
+
+export const StaticScoreContent = ({
+  scoreConditionals,
+}: {
+  scoreConditionals: ScoreConditionalLogic[];
```
IhorNesterenko99: Please move to types (StaticScoreContent.types.ts).

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.tsx
```diff
@@ -67,6 +67,7 @@ export const ScoresAndReports = () => {
     name: reportsName,
     keyName: REACT_HOOK_FORM_KEY_NAME,
   });
+  const reportsSize = reports?.length ?? 0;
```
IhorNesterenko99, line 70:
```suggestion
  const reportsSize = reports?.length ?? null;
```

src/modules/Builder/features/ActivitySettings/ScoresAndReports/SectionContent/StaticSectionContent/index.ts
```diff
@@ -0,0 +1,3 @@
+import { StaticSectionContent } from './StaticSectionContent';
+
+export { StaticSectionContent };
```
IhorNesterenko99: Why not export * from './StaticSectionContent';? Also in other places.

src/shared/hooks/useStaticContent.ts
```diff
@@ -0,0 +1,28 @@
...
+import { useIntersectionObserver } from 'shared/hooks/useIntersectionObserver';
+
+export const useStaticContent = ({
+  targetSelector,
+  isStaticActive = false,
+}: {
+  targetSelector: string;
+  isStaticActive?: boolean;
+}) => {
+  const [isStatic, setStatic] = useState(isStaticActive);
```
IhorNesterenko99, line 12: What do you think about using useState(false) here? Since isStaticActive can be undefined, we can set an initial state of false to handle this. Additionally, in the useEffect hook, we can add the following code: if (isStaticActive !== undefined) setStatic(isStaticActive);. This will update the static state only if isStaticActive is not undefined.

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.tsx
```diff
@@ -96,6 +107,7 @@ export const ScoreContent = ({
     name: scoreConditionalsName,
     keyName: REACT_HOOK_FORM_KEY_NAME,
   });
+  const conditionalsSize = scoreConditionals?.length ?? 0;
```
IhorNesterenko99, line 110:
```suggestion
  const conditionalsSize = scoreConditionals?.length ?? null;
```

alex-i-e: No need, since we use this value in comparison later

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.tsx
alex-i-e: No need, since we use this value in comparison later

src/shared/hooks/useStaticContent.ts
alex-i-e: Yes, exactly. Please check the latest commits, I set this variable to `false` by default

IhorNesterenko99 (approved):

PR #1554, opened 2024-03-07
M2-4663: [Admin Panel] Review Tab: Review Date is not disabled if Activities in the Applet have never been completed by selected Respondent

alex-i-e:

### 📝 Description

🔗 [Jira Ticket M2-4663](https://mindlogger.atlassian.net/browse/M2-4663)

### 📸 Screenshots

Before (Optional) | After 
--- | --- 
image | image


src/modules/Dashboard/features/RespondentData/RespondentDataReview/ReviewMenu/ReviewMenu.tsx
```diff
@@ -42,7 +42,7 @@ export const ReviewMenu = ({
           minDate={null}
           includeDates={responseDates}
           onMonthChange={onMonthChange}
-          disabled={false}
+          disabled={!responseDates?.length}
```
IhorNesterenko99: I think it could be dangerous to disable the date picker this way, because the response days are set by this call and fromDate and toDate are equal to month:
image
In my opinion, it's better to use the lastSeen date to disable the date picker if its value is null or undefined (there is no answer at all).
image

alex-i-e: Good catch! updated

IhorNesterenko99 (approved):

PR #1564, opened 2024-03-13
M2-2217: GLOBAL - Table lines shall never be hoverable and highlightable

IhorNesterenko99:

### 📝 Description

🔗 [Jira Ticket M2-2217](https://mindlogger.atlassian.net/browse/M2-2217)

Changes include:

- Fixed: table rows might be hoverable only in case if they are clickable
- Fixed: tables head background color in modals
- Fixed: tables head sticky position without opacity 
- Added indent between buttons in ``ActivityFlowBuilderHeader``

### 📸 Screenshots

| image | image |


src/shared/components/Table/TableHead/TableHead.styles.ts
```diff
@@ -9,28 +9,31 @@ import { UiType } from '../Table.types';
...
+      ${({ uiType, tableHeadBg }) => {
+        if (uiType === UiType.Secondary || uiType === UiType.Tertiary) {
+          return `background-color: ${tableHeadBg || variables.palette.surface3}`;
+        }
 
-      if (uiType === UiType.Quaternary) {
-        return `background-color: ${variables.palette.surface1}`;
-      }
+        if (uiType === UiType.Quaternary) {
+          return `background-color: ${tableHeadBg || variables.palette.surface1}`;
```
alex-i-e, line 31: Could you please add a couple of tests to cover this condition (the same in row 27, 34)?

IhorNesterenko99: Cannot test properly because there is an issue when trying to define variables inside the styled component in the test.

alex-i-e (approved):

PR #1571, opened 2024-03-14
feat: experimental feature flags shared const

jgdev:
### 📝 Description

Create fake experimental constant to include all LD feature flags.

### DIFF
--- a/src/shared/consts.tsx
+++ b/src/shared/consts.tsx
@@ -334,7 +334,7 @@ export const NON_UNIQUE_VALUE_MESSAGE = 'Non-unique value.';
 
 export const NULL_ANSWER = 'value: null';
 
-// TODO: temporally implementation of LaunchDarkly.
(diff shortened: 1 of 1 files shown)


src/shared/consts.tsx
```diff
@@ -333,3 +333,8 @@ export const JEST_TEST_TIMEOUT = 10000;
 export const NON_UNIQUE_VALUE_MESSAGE = 'Non-unique value.';
 
 export const NULL_ANSWER = 'value: null';
+
+// TODO: temporally implementation of LaunchDarkly.
```
farmerpaul:
```suggestion
// TODO: temporary implementation of LaunchDarkly. Will be removed when LD integration is added
```
and perhaps it would be good to also include a link to the JIRA ticket for the LD integration task?

farmerpaul (approved):

PR #1577, opened 2024-03-18
M2-5835 Add LaunchDarkly for feature flagging

ChaconC:
### 📝 Description

🔗 [Jira Ticket M2-5835](https://mindlogger.atlassian.net/browse/M2-5835)

Adds the LaunchDarkly SDK and a custom hook for accessing LaunchDarkly in a limited manner, to avoid passing additional user info which may expose PII.

### 🪤 Peer Testing

**Requires `npm install`**

- Update `.env` following `.env.example`
- Add a conditional component somewhere in the app, for example in `App.tsx`
```
...
import { useLaunchDarkly } from 'shared/hooks/useLaunchDarkly';

...
  const { flags } = useLaunchDarkly();  

  return (
    <Provider store={store}>
...
          <p>{flags.multiInformantFlag ? <b>Flag on</b> : <b>Flag off</b>}</p>
...
    </Provider>
  );
};

```
- Create an account with a `@metalab.com` email (configurable via `ENV`)
- Log in to the app to join the Multi Informant segment
- Logging off will clear the flag

### 📸 Screenshots

| Flag Off                      | Flag On                                 |
| CleanShot 2024-03-18 at 00 58 51 | CleanShot 2024-03-18 at 01 00 10 |

### ✏️ Notes

`LAUNCHDARKLY_CLIENT_ID` is a public value, i.e. not a secret, it just needs to be updated depending on the environment :) 

* Waiting on confirmation the environment variable has been deployed to Jenkins


alex-i-e (dismissed):

src/shared/hooks/useLaunchDarkly.ts
```diff
@@ -0,0 +1,56 @@
+import { LDFlagValue, useFlags, useLDClient } from 'launchdarkly-react-client-sdk';
+
+import { FeatureSegments, FeatureFlags, FeatureFlagsKeys } from 'shared/types/featureFlags';
+
+const omitKeys = ['firstName', 'lastName', 'email'];
```
farmerpaul: What do you think about calling this constant `PROHIBITED_PII_KEYS` to make its purpose very explicit? Also, perhaps moving it into `useLaunchDarkly.const.ts`?

.env.example
```diff
@@ -1,3 +1,6 @@
 REACT_APP_API_DOMAIN=http://localhost:8000
 REACT_APP_ENV='dev'
 REACT_APP_DEVELOP_BUILD_VERSION=
+REACT_APP_LAUNCHDARKLY_CLIENT_ID=65df9142fbf14e10954ac55d
```
farmerpaul, line 4: Should this be added to the `README`? (Omitting the one below, since it's a temporary workaround)

ChaconC: Yes, I didn't have time to write up documentation when I posted this draft but I'll be sure to include this new env

ChaconC: Will leave LD documentation on Confluence, no need to replicate here. Just added the env to README

src/shared/hooks/useLaunchDarkly.ts
```diff
@@ -0,0 +1,56 @@
...
+  const identifyLDContext = (
+    context: {
+      userId: string;
+      workspaceId?: string;
+      appletId?: string;
+      featureTests?: FeatureSegments[];
+    },
+    onDone?: ((err: Error | null) => void) | undefined,
+  ) => {
+    if (PROHIBITED_PII_KEYS.some((val) => Object.keys(context).includes(val))) {
```
sabrichu, line 23: It might make more sense to have an allowlist (`workspaceId`, `appletId`, `featureTests`) instead of blocklist. It seems unlikely we'd be able to predict the keys that someone might pass that could be problematic

ChaconC: I think we still don't know _what_ we're going to be sending to properly identify a user's context...and we may want to pass additional data at some point

sabrichu: Yes that's very true! So all TBD...

IhorNesterenko99 (approved):

alex-i-e (approved):

PR #1580, opened 2024-03-18
M2-5579: [Participant Overview] List + Sort

sultanofcardio:
### 📝 Description

🔗 [ Jira Ticket M2-5579](https://mindlogger.atlassian.net/browse/M2-5579)
🔗 [ Figma link](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=808%3A126146&mode=design&t=NoVlk8x5mTM773pW-1)

This PR introduces the `/dashboard/<applet-id>/participants` route with the new design for the participants overview page. Continuing the theme of copy over modify, I decided to introduce a new `<Participants>` feature component for this change.

#### Before

image

#### After

image

### 🪤 Peer Testing

> [!NOTE]
> You'll need at least one applet for this test

1. Add this variable to your `.env` file
  ```shell
  REACT_APP_FEATURE_FLAG_APPLET_MULTI_INFORMANT=true
  ```
2. Start the backend API (from the [feature-multiinformant](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/tree/feature-multiinformant) branch!)
3. Start the Admin API and navigate to the Applet Participants overview tab

https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/af054c8c-f521-4de4-99b5-ed704d6d32cd

4. Add participants if necessary

Here are some criteria to keep in mind when testing this change:
- If I access the participants tab, a list of participants loads for viewing
- This list should include essential details such as: ID, NickName, Latest Activity, and Schedule
- If study information is missing, it's marked with a "–".
- Clicking the …. on the list exposes the ability to:
  - edit participant, updated to full account, when applicable 
  - export data
  - remove participant
  - assign activity
- When I am viewing a table with multiple columns and I click on the header of a column the table should sort the rows based on the values in that column in ascending order.
  - If I click the column header again, the order should toggle to descending.
- When a column in a table is sorted, a visual indicator arrow should appear next to the column header.
- The direction of the arrow indicates the sort order: an up arrow (↑) for ascending order and a down arrow (↓) for descending order.

#### Table Headers

I opted to leave the bottom border of the table header visible because the table has two header rows (as opposed to the design which only has one), and I wasn't able to find a combination of header borders that didn't feel weird besides what was already there.


src/resources/app-fr.json
```diff
@@ -674,6 +676,7 @@
   "lastName": "Nom de Famille",
   "lastNameRequired": "Le nom de famille est requis",
   "latestActive": "Dernière activité",
+  "latestActivity": "Dernère Activité",
```
farmerpaul:
```suggestion
  "latestActivity": "Dernière Activité",
```

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -0,0 +1,564 @@
...
+              </StyledCheckBox>
+            }
+            data-testid="dashboard-participants-checkbox"
+          />
+        ),
+        value: '',
+        width: ParticipantsColumnsWidth.Pin,
+      },
+      pin: {
+        content: () => <Pin isPinned={isPinned} data-testid="dashboard-respondents-pin" />,
```
farmerpaul: Just making a note of this here (though you don't need to change this line):

The `<Pin>`'s default (unpinned) colour is out of sync with Figma:
- It should be `outline_variant` when unpinned (instead of `surface_variant`) – see `Pin.styles.tsx`
- The value for the `outline_variant` token is also not in sync with what's in Figma; it should be `#c2c7cf` (instead of `#c4c9d0`) – see `palette.ts`

Might seem minor but I couldn't help but clock its lack of contrast with the background when you're hovering over a table row (my a11y-attuned eyes I guess 👀).

src/svgSprite.ts
```diff
@@ -293,6 +293,12 @@ export const svgSprite = (): string => `
       fill=${variables.palette.primary50 || 'currentColor'} fillOpacity="0.87" />
     </symbol>
 
+    <symbol viewBox="0 0 24 24" id="svg-dropdown-down-outlined">
+      <path d="M16.9997 9.1697C16.8123 8.98345 16.5589 8.87891 16.2947 8.87891C16.0305 8.87891 15.7771 8.98345 15.5897 9.1697L11.9997 12.7097L8.4597 9.1697C8.27234 8.98345 8.01889 8.87891 7.7547 8.87891C7.49052 8.87891 7.23707 8.98345 7.0497 9.1697C6.95598 9.26266 6.88158 9.37326 6.83081 9.49512C6.78004 9.61698 6.75391 9.74769 6.75391 9.8797C6.75391 10.0117 6.78004 10.1424 6.83081 10.2643C6.88158 10.3861 6.95598 10.4967 7.0497 10.5897L11.2897 14.8297C11.3827 14.9234 11.4933 14.9978 11.6151 15.0486C11.737 15.0994 11.8677 15.1255 11.9997 15.1255C12.1317 15.1255 12.2624 15.0994 12.3843 15.0486C12.5061 14.9978 12.6167 14.9234 12.7097 14.8297L16.9997 10.5897C17.0934 10.4967 17.1678 10.3861 17.2186 10.2643C17.2694 10.1424 17.2955 10.0117 17.2955 9.8797C17.2955 9.74769 17.2694 9.61698 17.2186 9.49512C17.1678 9.37326 17.0934 9.26266 16.9997 9.1697Z"
+      fill=${variables.palette.on_surface_variant || 'currentColor'} />
```
sabrichu: I realize this is a pattern in the rest of this file, but I feel like we should always be using `currentColor` in these sprites and leave it up to the consumers of the `<Svg />` component to decide the fill, rather than having the sprite have a default color that would need to be overwritten

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -0,0 +1,564 @@
...
+  );
+
+  const { execute: updateSubjectsPin } = useAsync(
+    updateSubjectsPinApi,
+    handleReload,
+    undefined,
+    () => setIsLoading(false),
+  );
+
+  const handlePinClick = ({ respondentId, subjectId }: HandlePinClick) => {
```
farmerpaul, line 239: Placing this comment here, as it's related to the pin interaction.

I get that optimistic updates are probably not in scope for now (even though that would be ideal 😏). For now, just wanted to clock that the loading state doesn't seem to be fully aligned with the toggling of the pin icon – the loader disappears, then there's a short delay, and then the UI updates:

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/8e55428a-333c-4553-80e8-01363d1e39e8

I don't know if it's easy to fix this, but just wanted to flag.

src/svgSprite.ts
farmerpaul: +1

farmerpaul (commented):
Looks great @sultanofcardio! I just noticed a few pretty minor things, and I'm curious what your thoughts are on the pinning action loading state. Nice work!

src/resources/app-fr.json
sultanofcardio: I'm a bit confounded with this one. When I translate "Latest Activity" on Google Translate, I get Dernère (which was apparently marked as correct by contributors):

Google Translate translating Latest Activity in English to Dernère Activité in French

However, when I try to translate it back it tries to correct me as you did from Dernère to Dernière:

Google Translate translating Dernère Activité in French to Last Activity in English, suggesting that I actually meant Dernière Activité (which I didn't)

Given that there's an existing entry for `"latestActive": "Dernière activité"`, I wonder which one of these is correct 🤔

src/svgSprite.ts
sultanofcardio: Sure, I think the result would be the same so I'm happy to break the pattern here

src/modules/Dashboard/features/Participants/Participants.tsx
sultanofcardio: Great catch Paul! I hesitate to make this change as part of this ticket since the `Pin` component is used in a few places, and `outline_variant` is used in even more places. I'd rather make that change as part of a ticket where I can investigate the usages

What do you think about overriding the unpinned style for this specific usage?

src/modules/Dashboard/features/Participants/Participants.tsx
sultanofcardio: Looks like this is an easy one. I'll go for the quick win for now, as doing an optimistic pin update isn't as straightforward as it seems

src/modules/Dashboard/features/Participants/Participants.styles.ts
```diff
@@ -0,0 +1,86 @@
...
+  min-width: 137px;
+`;
+
+export const StyledCheckBox = styled(StyledIcon)`
+  svg {
+    fill: ${variables.palette.outline_variant};
+  }
+`;
+
+export const MuiCheckbox = styled(Checkbox)``;
```
IhorNesterenko99: It seems this row can be removed.

src/modules/Dashboard/features/Participants/Participants.utils.tsx
IhorNesterenko99, line 1: Could you please add test coverage?

src/modules/Dashboard/features/Participants/Participants.tsx
IhorNesterenko99, line 1: Could you please add test coverage?

src/modules/Dashboard/features/Participants/Participants.styles.ts
sultanofcardio: Yup, thanks for pointing this out 👍

src/resources/app-fr.json
farmerpaul: I guess I don't trust Google translate's reviewers here. Googling "dernère vs dernière" doesn't return any results suggesting "dernère" is a valid spelling of the word, though some people have indeed misspelled it online. In my 8+ years of French immersion I never encountered a version of that word that omits the "i".

As for the other key, it's possible that the translation there isn't quite right (maybe something like "Dernier actif" could work), or that they should both be the same even though in English there's meaningful distinction (French can be weird that way).

Having said all that, IMHO it shouldn't be the responsibility of non-bilingual engineers to be making these decisions… given that Google Translate has been the go-to till now, probably many Google-translated keys aren't correct and need to pass through a proper translation service.

src/modules/Dashboard/features/Participants/Participants.tsx
farmerpaul: Hmm… I think it's probably best to leave this to a separate ticket rather than fragment this use-case, which I think will introduce unneeded complexity.

src/modules/Dashboard/features/Participants/Participants.utils.tsx
sultanofcardio: Done

src/modules/Dashboard/features/Participants/Participants.tsx
sultanofcardio: Done

src/resources/app-fr.json
sultanofcardio: That's fair. I will make the change as I trust you're much better with French that I am 😅

src/modules/Dashboard/features/Participants/Participants.tsx
sultanofcardio: Ok no prob, I'll create the Jira ticket. I ran into a hiccup and asked about it in Slack but I'll follow up with its creation once we iron that out

IhorNesterenko99 (approved):

src/modules/Dashboard/features/Participants/Participants.styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  td:nth-child(1) {
+    padding-left: 14px;
+  }
+
+  td,
+  th {
+    min-width: 130px;
+  }
```
farmerpaul, line 73: Just noting that many lengths in this file are specified in `px`, but if you search the code, `rem` is primarily used throughout the app. Whether or not it will currently impact anything is uncertain, but the beauty of `rem` is that you can use it to scale the whole site via a root-level `font-size` change (if that's ever desired for a given breakpoint, perhaps in the future – handy for widescreen breakpoints, for instance). The conversion to `rem` is easy since `10px == 1rem`.

farmerpaul (approved):
Nice work on the loading state. I just had one last comment regarding CSS length units, but preapproving! 🚀

src/modules/Dashboard/features/Participants/Participants.styles.ts
sultanofcardio: Good call. I think that's a worthwhile change. Thanks for the pointer

PR #1566, opened 2024-03-13
M2-5431: Introduce "feature flagged" AppletMultiInformant with updates to Applet Secondary Navigation

sabrichu:
### 📝 Description

🔗 [Jira Ticket M2-5431](https://mindlogger.atlassian.net/browse/M2-5431)
🔗 [Jira Ticket M2-5763](https://mindlogger.atlassian.net/browse/M2-5763)
🔗 [Jira Ticket M2-5764](https://mindlogger.atlassian.net/browse/M2-5764)

This updates our dashboard route to render an `AppletMultiInformant` copy of `Applet` if an `ENABLE_MULTI_INFORMANT_APPLET_MANAGEMENT` "flag" is true. **This will be replaced with a real feature flag when https://mindlogger.atlassian.net/browse/M2-5835 lands.**

`AppletMultiInformant` is a copy of `Applet` that changes what navigation we are exposing to the user via the Applet secondary navigation to align with the new designs: https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=808-126146&mode=design&t=3XPIWkDuGcsRN1nF-4. Specifically:

- Adds two blank placeholder pages for "Overview" and "Activities"
- Updates "Respondents" to "Participants" (**tab label only, url remains "respondents"**)
- Updates "Managers" to "Team" (**tab label only, url remains "respondents"**)
- Adds new SVGs to the SVG sprite
- Left-aligns the tabs

Currently the `Tabs` component has styles to auto-center itself/give itself a specific width, and it is used all throughout the site. To avoid messing with other pages, I added a `isCentered` prop that the Applet turns off for its use case.

**Ideally**, `Tabs` would be ignorant of its placement on the page (i.e., always been 100% width with no opinions on left or center placement, and parents of the component would control its layout on a case-by-case basis. However, to reduce the footprint of this change, this prop will do for now.

### 📸 Screenshots

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1165936/c27ee89f-56b3-48a9-946f-98b67185c7c7

| Before (Optional)                      | After                                 |
| -------------------------------------- | ------------------------------------- |
| Screenshot 2024-03-13 at 8 07 12 AM | Screenshot 2024-03-13 at 8 07 18 AM |

PR #1588, opened 2024-03-20
M2-5951: Use feature flags in Applet menu

ChaconC:
### 📝 Description

🔗 [Jira Ticket M2-5951](https://mindlogger.atlassian.net/browse/M2-5951)

Changes include:

- Replaces temporary workaround with LD FeatureFlags

### 🪤 Peer Testing

- Log in with a `multi-informant` flagged account, e.g. `mail@metalab.com`
- Navigate to an Applet details screen
- It should show the new tab bar with Overview and Activities from #1566

### DIFF
--- a/src/routes/index.tsx
+++ b/src/routes/index.tsx
@@ -12,6 +12,7 @@ import { auth } from 'redux/modules';
 import { AppletNotFoundPopup } from 'shared/components';
 import { NoPermissionPopup } from 'shared/components/NoPermissionPopup';
 import { useSessionBanners } from 'shared/hooks/useSessionBanners';
+import { useLaunchDarkly } from 'shared/hooks/useLaunchDarkly';
 
 import history from './history';
 
@@ -22,6 +23,7 @@ export const AppRoutes = () => {
   const token = authStorage.getAccessToken();
   const dispatch = useAppDispatch();
   const isAuthorized = auth.useAuthorized();
+  const { flags } = useLaunchDarkly();
 
   const status = auth.useStatus();
   const loaded = !token || status === 'error' || status === 'success';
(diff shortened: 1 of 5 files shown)


ChaconC:
trying to trigger preview env build, will re-open

src/shared/consts.tsx
```diff
@@ -333,9 +333,3 @@ export const JEST_TEST_TIMEOUT = 10000;
 export const NON_UNIQUE_VALUE_MESSAGE = 'Non-unique value.';
 
 export const NULL_ANSWER = 'value: null';
-
-// TODO: temporary implementation of LaunchDarkly. Will be removed when LD integration is added
```
sabrichu, line 337: 🥹👋 SEE YA NEVER

sabrichu (approved):
That temporary flag didn't last a week and that's the way it should be <3

src/routes/index.tsx
```diff
@@ -22,6 +23,7 @@ export const AppRoutes = () => {
   const token = authStorage.getAccessToken();
   const dispatch = useAppDispatch();
   const isAuthorized = auth.useAuthorized();
+  const { flags } = useLaunchDarkly();
```
sabrichu, line 26: Not blocking/know this is because of how the files are currently written, but would prefer to update `dashboardRoutes` to be `<DashboardRoutes />` so that we can do this call where the flag is being used, instead of passing it through here. But then we'd want to update `builderRoutes` too, and `libraryRoutes`...a whole thing. Anyway just speaking my piece haha

ChaconC: I initially set it up as `<DashboardRoutes />` however...we're doing conditional rendering in `<AppRoutes />` and react complains about wrong use of hooks

sabrichu: Why can't we have nice things!!!!!

PR #1593, opened 2024-03-21
M2-5439: Update export/settings UI in applet header

sabrichu:
### 📝 Description

🔗 [Jira Ticket M2-5439](https://mindlogger.atlassian.net/browse/M2-5439)

This updates the location of the Export Data button and settings navigation to the top right corner.

The addition of the `defaultToFirstTab` and `animateTabIndicator` props to the `LinkedTabs` component accounts for the fact that the indicator for `LinkedTabs` is controlled by the path in the URL, and clicking the settings gear will navigate the user to `/dashboard/[applet_id]/settings`, causing the indicator to either animate strangely (zooming off the page) or selecting the first tab ("Overview") even though it's the settings page.

In the future, the settings page will exist in a modal-like slideover that _shouldn't_ update the url, so this won't be a problem (https://mindlogger.atlassian.net/browse/M2-5987). Until then, this will have to do.

### 📸 Screenshots

Screenshot 2024-03-21 at 10 16 44 AM
Screenshot 2024-03-21 at 10 16 51 AM
Screenshot 2024-03-21 at 10 16 57 AM


src/modules/Dashboard/pages/Applet/AppletMultiInformant.styles.ts
```diff
@@ -0,0 +1,17 @@
+import { styled, Button } from '@mui/material';
+import { Link } from 'react-router-dom';
+
+import { StyledFlexTopCenter, theme, variables } from 'shared/styles';
+
+export const StyledActions = styled(StyledFlexTopCenter)`
+  gap: ${theme.spacing(1)};
+`;
+
+export const StyledExportButton = styled(Button)`
```
sabrichu: This feels like such minimal styles that it feels excessive to add a new Styled component, let alone a new file. Is there anything I can use instead to get rid of this file?

jgdev: @sabrichu I think the use of the [`sx`](https://mui.com/system/getting-started/the-sx-prop/) property could work, it's being used in other components as convention and it was suggested by @IhorNesterenko99 in previous PR reviews.

```tsx
<Button sx={{ gap: theme.spacing(1), color: variables.palette.on_surface_variant }} />
```

IhorNesterenko99: Usually, we use sx instead of styled-components if the number of styles is no more than two.

sabrichu: Sounds good! I'll do the same with `StyledSettingsButton`

sabrichu: @IhorNesterenko99 Love it! I make the cut!!

src/modules/Dashboard/pages/Applet/AppletMultiInformant.tsx
```diff
@@ -35,37 +53,85 @@ export const AppletMultiInformant = () => {
...
+            </StyledPanel>
+          )}
         </>
       )}
+      <ExportDataSetting
+        isExportSettingsOpen={isExportOpen}
+        onExportSettingsClose={() => {
+          setIsExportOpen(false);
+        }}
+      />
```
sultanofcardio: This component uses the applet data as the source of data to export, but because it doesn't accept it as a prop, rendering it before the appletData is available leads to weird behavior if the user selects "All time" as the date range

https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/ac546a55-14c0-4429-8384-c514fe968263

That said, using "All time" seems to be borked anyway on the BE, but that's a different issue

```suggestion
      {appletData && (
        <ExportDataSetting
          isExportSettingsOpen={isExportOpen}
          onExportSettingsClose={() => {
            setIsExportOpen(false);
          }}
        />
      )}
```

sabrichu: Nice catch!

sultanofcardio (approved):
This otherwise looks good to me. Feel free to merge after making that change I mentioned!

src/modules/Dashboard/pages/Applet/Applet.styles.tsx
```diff
@@ -0,0 +1,8 @@
+import { styled } from '@mui/material';
+
+import { variables } from 'shared/styles';
+
+export const StyledAppletLogo = styled('img')({
+  width: '48px',
```
IhorNesterenko99:
```suggestion
  width: '4.8rem',
```

IhorNesterenko99 (approved):

PR #1613, opened 2024-04-01
M2-6090: Rename multi-informant feature flag

sabrichu:
### 📝 Description

🔗 [Jira Ticket M2-6090](https://mindlogger.atlassian.net/browse/M2-6090)

Minor update to rename the current multi-informant feature flag to `enable-multi-informant`.

After this lands, the old flag (`multi-informant-feature-flag`) will be archived.

### 📸 Screenshots

|  |  |

### DIFF
--- a/src/shared/types/featureFlags.ts
+++ b/src/shared/types/featureFlags.ts
@@ -1,9 +1,9 @@
 import { LDFlagValue } from 'launchdarkly-react-client-sdk';
 
 // These keys use the camelCase representation of the feature flag value
-// e.g. multi-informant-feature-flag in LaunchDarky becomes multiInformantFeatureFlag
+// e.g. enable-multi-informant in LaunchDarky becomes enableMultiInformant
 export const FeatureFlagsKeys = {
-  multiInformantFlag: 'multiInformantFeatureFlag',
+  multiInformantFlag: 'enableMultiInformant',
 };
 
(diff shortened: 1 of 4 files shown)


src/shared/types/featureFlags.ts
```diff
@@ -1,9 +1,9 @@
 import { LDFlagValue } from 'launchdarkly-react-client-sdk';
 
 // These keys use the camelCase representation of the feature flag value
-// e.g. multi-informant-feature-flag in LaunchDarky becomes multiInformantFeatureFlag
+// e.g. enable-multi-informant in LaunchDarky becomes enableMultiInformant
 export const FeatureFlagsKeys = {
-  multiInformantFlag: 'multiInformantFeatureFlag',
+  multiInformantFlag: 'enableMultiInformant',
```
farmerpaul: Any reason not to align the key names with the values too?
```suggestion
  enableMultiInformant: 'enableMultiInformant',
```

farmerpaul (approved):
Just one comment, but looks fine, preapproving!

src/shared/types/featureFlags.ts
sabrichu: Mostly wanted to keep my change minimal...but you're right that we might as well have them be the same instead of the wild wild west. Updating!

PR #1617, opened 2024-04-02
M2-5921: Create Update Jira Tickets GitHub Actions workflow

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-5921](https://mindlogger.atlassian.net/browse/M2-5921)

This PR introduces a GitHub Actions workflow that submits a list of ticket numbers to a Jira webhook based on a number of factors.

The workflow is triggered by a GitHub release, and it collates the list of Jira ticket numbers that were worked on since the last GitHub release that had a successful Jenkins build. It then waits for the outcome of the current build (trigged by the release) and submits that list of Jira tickets if the build is successful.

My original approach to this ticket was to modify the Jenkins scripts, but I changed that because it's much simpler to have this logic in one workflow (instead of being split across multiple CodeCommit repos), and we eventually intent to move CI/CD to GitHub Actions anyway so this is a step in that direction

### 🪤 Peer Testing

The code in this workflow is the same as that in https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1205. You may refer to the Peer Testing section of that PR's description for some test runs of these test scenarios

- **Success**: The happy path where there is a Jenkins build found for the previous release candidate, and we're able to generate and submit the Jira tickets that have been worked on since then
- **Not a release candidate**: The case where the workflow is triggered by a release with a tag that does not contain `-rc`
- **Invalid tag name**: The case where the workflow is run manually with a tag name that does not exist in the repository
- **No successful Jenkins build found**: The case where we are not able to find a successful Jenkins build for any of the previous release candidate tags in the repository
- **No Jenkins build triggered for tag/release**: The case where the release that triggered this workflow did not create a Jenkins build for some reason
- **Jenkins responds with 500 error code**: The case where Jenkins is having server errors
- **Jenkins doesn't respond**: The case where Jenkins doesn't respond to API calls

### ✏️ Notes

This PR targets main because I believe the workflow has to be on the default branch in order to take effect

This workflow is currently using Jenkins API credentials that I created from my Jenkins user. Before this is merged, we need to replace those credentials with a dedicated API user

- [ ] Replace my Jenkins API credentials with API user

### DIFF
--- a/.github/workflows/update-jira-tickets.yaml
+++ b/.github/workflows/update-jira-tickets.yaml
@@ -1,0 +1,120 @@
+name: Update Jira Tickets
+
+on:
+  release:
+    types: [created]
+  # Allows running manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      tagName:
+        description: "Tag Name"
+        required: true
+
+jobs:
+  process-release:
+    env:
+      JENKINS_USER: ${{ secrets.JENKINS_USER }}
+      JENKINS_TOKEN: ${{ secrets.JENKINS_TOKEN }}
+      JIRA_WEBHOOK_URL: ${{ secrets.JIRA_WEBHOOK_URL }}
+      JENKINS_HOST: ${{ vars.JENKINS_HOST }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Get current tag
+        id: get-tag
+        run: |
+          if [ "${{ github.event_name }}" == "release" ]; then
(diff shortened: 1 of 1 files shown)


IhorNesterenko99:
Hello! Could you please confirm if it should be merged directly into the main branch, bypassing our usual workflow of dev -> release -> main?

sultanofcardio:
> Hello! Could you please confirm if it should be merged directly into the main branch, bypassing our usual workflow of dev -> release -> main?

@IhorNesterenko99 yes this change needs to go to the default branch of the repository

rcmerlo (approved):
Please, squash commits before merge

PR #1622, opened 2024-04-03
M2-6111: [Participant Overview] Link row to participant detail page

sabrichu:
### 📝 Description

🔗 [Jira Ticket M2-6111](https://mindlogger.atlassian.net/browse/M2-6111)

Adds a default `onClick` action for cells in the Participants Table that navigates to the participant detail page.

I wanted to just add a single `onClick` for the row itself, but `DatabaseTable` (the underlying component for `ParticipantsTable`) isn't set up to do that right now and is used in other places on the site. So while not the most DRY, this solution will hopefully do for now...

Also fixes the breadcrumbs for the Participants Detail page to go to `/participants` instead of `/respondents` if the feature flag is on

### 🪤 Peer Testing

- This change is based off the `feature/multiinformant-metapod` branch and requires that the `enable-multi-informant` feature flag is **ON**
- On the participants overview page, clicking on a participant row should navigate to the participant details page for that participant
- Clicking on checkmark in row should NOT navigate to participant details page
- Clicking on pin in row should NOT navigate to participant details page
- Clicking on actions in row should NOT navigate to participant details page

### DIFF
--- a/src/modules/Dashboard/features/Participants/Participants.tsx
+++ b/src/modules/Dashboard/features/Participants/Participants.tsx
@@ -284,6 +284,15 @@ export const Participants = () => {
     const stringSecretIds = joinWihComma(secretIds, true);
     const respondentOrSubjectId = respondentId ?? details[0].subjectId;
 
+    const defaultOnClick = () => {
+      navigate(
+        generatePath(page.appletParticipantActivities, {
+          appletId,
+          participantId: respondentOrSubjectId,
+        }),
+      );
+    };
+
     return {
       checkbox: {
         content: () => (
@@ -310,16 +319,19 @@ export const Participants = () => {
         content: () => stringSecretIds,
         value: stringSecretIds,
         width: ParticipantsColumnsWidth.Default,
+        onClick: defaultOnClick,
       },
       nicknames: {
         content: () => stringNicknames,
         value: stringNicknames,
         width: ParticipantsColumnsWidth.Default,
+        onClick: defaultOnClick,
       },
       tags: {
         content: () => <>--</>,
         value: '',
         width: ParticipantsColumnsWidth.Default,
+        onClick: defaultOnClick,
       },
       status: {
         content: () => (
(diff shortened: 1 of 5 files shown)


IhorNesterenko99 (approved):

alex-i-e (approved):

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -331,17 +343,20 @@ export const Participants = () => {
         ),
         value: '',
```
farmerpaul, line 344: While you're here, I feel like this shouldn't be blank, in case the column is used for sorting:
```tsx
        value: status,
```

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -310,16 +319,19 @@ export const Participants = () => {
...
+        onClick: defaultOnClick,
       },
       nicknames: {
         content: () => stringNicknames,
         value: stringNicknames,
         width: ParticipantsColumnsWidth.Default,
+        onClick: defaultOnClick,
       },
       tags: {
         content: () => <>--</>,
```
farmerpaul, line 331: Also while you're here, I'd previously created a shared component specifically for the case to correctly style `--` when the value is falsy (it's supposed to be a lighter grey):

```suggestion
        // TODO: Replace `null` with tags component when available
        content: () => <StyledMaybeEmpty>{null}</StyledMaybeEmpty>,
```

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -331,17 +343,20 @@ export const Participants = () => {
         ),
         value: '',
         width: ParticipantsColumnsWidth.Status,
+        onClick: defaultOnClick,
       },
       lastSeen: {
         content: () => latestActive,
```
farmerpaul, line 349: Here too while you're at it, if you don't mind 😇 

```tsx
const latestActive = lastSeen && timeAgo.format(getDateInUserTimezone(lastSeen));

…


      lastSeen: {
        content: () => <StyledMaybeEmpty>{latestActive}</StyledMaybeEmpty>,
        …
      },
```

farmerpaul (commented):
Looks great! I just had a few related suggestions to tighten up this table while you're fixing things up.

sabrichu:
> Looks great! I just had a few related suggestions to tighten up this table while you're fixing things up.

Oops missed this before merging! Happy to address in a new PR

PR #1628, opened 2024-04-04
M2-0000: setup the test flag

alex-i-e:


src/index.tsx
```diff
@@ -28,11 +29,20 @@ Sentry.init({
 
 const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement);
 
+console.log({
```
IhorNesterenko99, line 32: do we need to leave console.log here?

src/modules/Dashboard/pages/Main/Main.tsx
```diff
@@ -7,6 +8,9 @@ import { Mixpanel } from 'shared/utils/mixpanel';
 import { dashboardTabs } from './Main.const';
 
 export const Main = () => {
+  const { testingFlag } = useFlags();
+  console.log({ testingFlag });
```
IhorNesterenko99, line 12: do we need to leave console.log here?
if yes, please add comments

PR #1631, opened 2024-04-05
[M2-5887] Update LaunchDarkly segment detection

ChaconC:
- [x] Related documentation has been added / updated - Need to update Confluence

### 📝 Description

🔗 [Jira Ticket M2-5887](https://mindlogger.atlassian.net/browse/M2-5887)

Updates LaunchDarkly's `identify` use to send workspace names instead of selecting the feature flag from the client.
Previously `identify` was only called when logging in through `LoginForm`, now it should be called anytime `Auth` is updated.

Changes include:

- Moves `identify` to occur after auth and any time workspaces are updated
- Remove `FeatureSegments`

### 📸 Screenshots

Flag is properly updated when switching to a workspace name that matches the `multi-informant` segment rule

https://github.com/ChildMindInstitute/mindlogger-admin/assets/2916156/b28a7f67-493a-41a4-b3c9-03bcbca37eb2

### ✏️ Notes

Segment rules use `workspaceName` to flag users as being part of a feature flag segment.
Currently, the only rule active is on the `testingFlag` (not declared in `featureFlags.ts`) where if the string `multi-infomant` is part of the workspace name it returns `true`, otherwise it returns `false`

### 🪤 Peer Testing

- Add `testingFlag` to the declared flags in `featureFlags.ts`
- Navigate to Applet > Participants > Add Participant
- Invite a new member with role `Manager` and change the workspace name to include `multi-informant`
- The value of the flag `testingFlag` should turn `true`


ChaconC:
Need to add relevant documentation on Confluence but that shouldn’t really block review for now.

src/App.tsx
```diff
@@ -23,6 +25,14 @@ const App = () => {
     })();
   }, []);
 
+  // Retrieves the LD client after initialization to pass it to our singleton
+  const ldClient = useLDClient();
+  useEffect(() => {
+    if (ldClient) {
```
alex-i-e: Could you please avoid nesting a bit:
```
if (!ldClient) return;
LaunchDarkly.init(ldClient);
```

src/shared/utils/launchDarkly.ts
```diff
@@ -0,0 +1,52 @@
...
+
+let _ldClient: LDClient;
+let _userId: string;
+
+export const LaunchDarkly = {
+  async init(client: LDClient) {
+    _ldClient = client;
+  },
+  async login(userId: string) {
+    if (_ldClient) {
```
alex-i-e: Please use early return to avoid nesting here and in methods below...

alex-i-e (approved):

PR #1642, opened 2024-04-09
M2-5585: Participant grid view

ChaconC:

### 📝 Description

🔗 [Jira Ticket M2-5585](https://mindlogger.atlassian.net/browse/M2-5585)

Changes include:

- Extracted Activities grid view into its own component
- Data fetch, `actions`, and `formatRow` is separated from grid so each view can handle it as necessary, e.g. different endpoints
- Removed search as per latest designs and following recent conversations
- Moved Participant views inside a Participant folder

### 📸 Screenshots

Video: 

https://github.com/ChildMindInstitute/mindlogger-admin/assets/2916156/006ae3c8-ca46-42bd-86d5-adbd3a6e95c9

|  CleanShot 2024-04-09 at 05 54 36 | CleanShot 2024-04-09 at 05 51 45 |

### 🪤 Peer Testing

Test using a `enable-multi-informant` enabled branch or use the preview environment.
- Navigate to any Applet that contains at least 1 Activity
- Select "Participants" tab
- Select any participant
- Activity list should be populated
- Activity list at the Applet level should also work correctly.

### ✏️ Notes

The Participant Activity list is using `getSummaryActivitiesApi` to retrieve the user's activities. This endpoint doesn't return all the related fields for the `Activity` but it does filter by activities assigned to the `Participant`. This should be replaced/updated as part of future work when the concept of assigning activities is introduced.


src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
```diff
@@ -0,0 +1,126 @@
...
+                sx={{ minWidth: '13.2rem' }}
+              >
+                {t('addActivity')}
+              </StyledButton>
+            </StyledFlexWrap>
+          </>
+        )}
+      </StyledFlexWrap>
+
+      {rows?.length && (
```
farmerpaul: Protect against `rows.length === 0`:
```suggestion
      {!!rows?.length && (
```

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+import { StyledButton } from './ActivityGrid.styles';
+
+export type ActivitiesGridProps = {
+  rows?: Row[];
+  order: Order;
+  orderBy: string;
+  isLoading: boolean;
+  'data-testid': string;
+};
```
farmerpaul: As much as I also advocate for Props types to be defined in the same file as the component, the existing pattern for the Admin panel has been to keep those definitions within respective `*.types.ts` files. 😇

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { StyledFlexTopCenter } from 'shared/styles';
+import { DatavizActivity, getSummaryActivitiesApi } from 'api';
+import { useAsync } from 'shared/hooks';
+import { DateFormats } from 'shared/consts';
+import { page } from 'resources';
+import { workspaces } from 'redux/modules';
+import { ActivityGrid } from 'modules/Dashboard/components/ActivityGrid/ActivityGrid';
+import { StyledSvg } from 'modules/Dashboard/components/ActivityGrid/ActivityGrid.styles';
+import { ActivityActionProps } from 'modules/Dashboard/components/ActivityGrid/ActivityGrid.types';
+import { getActivityActions } from 'modules/Dashboard/components/ActivityGrid/ActivityGrid.utils';
```
farmerpaul: Nit: could these be imported altogether from `modules/Dashboard/components/ActivityGrid` (needs `index.ts`)?

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
```diff
@@ -47,17 +38,19 @@ export const Activities = () => {
...
       },
     });
-  });
+
+    return () => {
+      setActivitiesData(null);
+    };
+  }, [appletId, getActivities]);
 
   const actions = useMemo(
```
farmerpaul: Would it be possible/practical to move this memo, and maybe the `formatRow` callback as well, into a shared hook that lives in the `ActivityGrid` component folder? (Like `ActivityGrid.hooks.tsx`?) Then the consuming screen could still override one of the `actions` if needed.

farmerpaul (commented):
Looks great @ChaconC, nice refactor! I just had a few minor suggestions and a suggestion for consolidating a couple things to further reduce repetition, see what you think!

Also I tried rebasing on `feature/multiinformant-metapod` again and there were no conflicts, but I'm getting a mysterious runtime TS error that might be due to a circular reference. I don't think it's your code, it might be due to the last merge from `develop`. Just a heads-up.

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
ChaconC: My bad! forgot to move it back to its own file 😞

farmerpaul (approved):
✨ :shipit: ✨

PR #1644, opened 2024-04-09
M2-5769, M2-6112: Update participant view header

JustinNusca:
### 📝 Description

🔗 [M2-5769](https://mindlogger.atlassian.net/browse/M2-5769): [Participant Detail] Export Function
🔗 [M2-6112](https://mindlogger.atlassian.net/browse/M2-6112): [Participant Detail] Open Settings

This PR updates the Participant View Header to include the "Export" and "Settings" options that are already visible on the Applet View Header.

Because these options are the same, I opted to abstract these options and their related functionality to a new component, which I've named `HeaderOptions`.

### 📸 Screenshots

| Before | After |
|-|-|
| localhost_3000_dashboard_76112bef-c9f9-4c72-8165-5bc0b64ea609_participants_8466fe1b-17df-4c56-805c-4257542de561(iPad Pro) | localhost_3000_dashboard_76112bef-c9f9-4c72-8165-5bc0b64ea609_participants_8466fe1b-17df-4c56-805c-4257542de561(iPad Pro) (1) |

### 🪤 Peer Testing

- On the Applet dashboard, the "Export" and "Settings" options continue to function as expected.
- On the Participant Details page within an applet, the "Export" and "Settings" options are visible in the header and function identically to the ones on the Applet Dashboard.

[M2-5769]: https://mindlogger.atlassian.net/browse/M2-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-6112]: https://mindlogger.atlassian.net/browse/M2-6112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
+++ b/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
@@ -1,12 +1,12 @@
 import { useState } from 'react';
-import { Box, Button } from '@mui/material';
+import { Button, IconButton } from '@mui/material';
 import { Link, generatePath, useParams } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { page } from 'resources';
 import { Svg } from 'shared/components';
 import { ExportDataSetting } from 'shared/features/AppletSettings';
-import { theme, variables } from 'shared/styles';
+import { StyledFlexTopCenter, variables } from 'shared/styles';
 import { Mixpanel } from 'shared/utils';
 
 export const HeaderOptions = () => {
@@ -25,27 +25,29 @@ export const HeaderOptions = () => {
   };
 
   return (
-    <Box sx={{ display: 'flex', gap: theme.spacing(1), placeItems: 'baseline' }}>
+    <StyledFlexTopCenter sx={{ gap: 1 }}>
       <Button
         data-testid="header-option-export-button"
         onClick={handleOpenExport}
-        startIcon={<Svg id="export" width={24} height={24} />}
+        startIcon={<Svg id="export" width={18} height={18} />}
         sx={{ color: variables.palette.on_surface_variant }}
       >
         {t('export')}
       </Button>
 
-      <Link
-        data-testid="header-option-settings-link"
+      <IconButton
+        component={Link}
+        sx={{ width: '4.8rem', height: '4.8rem' }}
         to={generatePath(page.appletSettings, { appletId })}
+        data-testid="header-option-settings-link"
       >
         <Svg id="settings" fill={isSettingsSelected ? variables.palette.primary : ''} />
-      </Link>
+      </IconButton>
 
       <ExportDataSetting
         isExportSettingsOpen={isExportOpen}
         onExportSettingsClose={handleCloseExport}
       />
-    </Box>
+    </StyledFlexTopCenter>
   );
 };
--- a/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
+++ b/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
@@ -1,12 +1,12 @@
-import { Box, Button } from '@mui/material';
 import { useState } from 'react';
+import { Box, Button } from '@mui/material';
 import { Link, generatePath, useParams } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { page } from 'resources';
 import { Svg } from 'shared/components';
 import { ExportDataSetting } from 'shared/features/AppletSettings';
-import { StyledTitleMedium, theme, variables } from 'shared/styles';
+import { theme, variables } from 'shared/styles';
 import { Mixpanel } from 'shared/utils';
 
 export const HeaderOptions = () => {
@@ -15,21 +15,24 @@ export const HeaderOptions = () => {
   const { appletId } = useParams();
   const isSettingsSelected = location.pathname.includes('settings');
 
+  const handleOpenExport = () => {
+    setIsExportOpen(true);
+    Mixpanel.track('Export Data click');
+  };
+
+  const handleCloseExport = () => {
+    setIsExportOpen(false);
+  };
+
   return (
     <Box sx={{ display: 'flex', gap: theme.spacing(1), placeItems: 'baseline' }}>
       <Button
         data-testid="header-option-export-button"
-        sx={{ color: variables.palette.on_surface_variant, gap: theme.spacing(1) }}
-        onClick={() => {
-          setIsExportOpen(true);
-          Mixpanel.track('Export Data click');
-        }}
+        onClick={handleOpenExport}
+        startIcon={<Svg id="export" width={24} height={24} />}
+        sx={{ color: variables.palette.on_surface_variant }}
       >
-        <Svg id="export" width={24} height={24} />
-
-        <StyledTitleMedium as="span" sx={{ color: variables.palette.on_surface_variant }}>
-          {t('export')}
-        </StyledTitleMedium>
+        {t('export')}
       </Button>
 
       <Link
@@ -41,9 +44,7 @@ export const HeaderOptions = () => {
 
       <ExportDataSetting
         isExportSettingsOpen={isExportOpen}
-        onExportSettingsClose={() => {
-          setIsExportOpen(false);
-        }}
+        onExportSettingsClose={handleCloseExport}
       />
     </Box>
   );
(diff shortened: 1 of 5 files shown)


src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,50 @@
+import { Box, Button } from '@mui/material';
+import { useState } from 'react';
```
IhorNesterenko99, line 1: Please move the import from React to the top.

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Mixpanel } from 'shared/utils';
+
+export const HeaderOptions = () => {
+  const [isExportOpen, setIsExportOpen] = useState(false);
+  const { t } = useTranslation('app');
+  const { appletId } = useParams();
+  const isSettingsSelected = location.pathname.includes('settings');
+
+  return (
+    <Box sx={{ display: 'flex', gap: theme.spacing(1), placeItems: 'baseline' }}>
```
IhorNesterenko99: In cases where the ``sx`` prop contains more than three properties, it would be beneficial to consider using styled components. This approach can enhance the readability and maintainability of the code. Please check other occurrences as well.

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Button
+        data-testid="header-option-export-button"
+        sx={{ color: variables.palette.on_surface_variant, gap: theme.spacing(1) }}
+        onClick={() => {
```
IhorNesterenko99: Could you please use separate handler function to improve readability?

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+          setIsExportOpen(true);
+          Mixpanel.track('Export Data click');
+        }}
+      >
+        <Svg id="export" width={24} height={24} />
+
+        <StyledTitleMedium as="span" sx={{ color: variables.palette.on_surface_variant }}>
```
IhorNesterenko99: Please consider using ``<Button startIcon={...}``. Also, does the text need to be wrapped in a ``<p>`` tag (StyledTitleMedium)?

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+      </Button>
+
+      <Link
+        data-testid="header-option-settings-link"
+        to={generatePath(page.appletSettings, { appletId })}
+      >
+        <Svg id="settings" fill={isSettingsSelected ? variables.palette.primary : ''} />
+      </Link>
+
+      <ExportDataSetting
```
IhorNesterenko99, line 42:
```suggestion
      {isExportOpen && <ExportDataSetting
```

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
mbanting: @IhorNesterenko99 agreed. Although it should be highlighted and fixed automatically with eslint and prettier. What do you think about having this added to the rules (in a separate ticket, not this one)?

mbanting (commented):
Nice work @JustinNusca 
I tested [M2-6112](https://mindlogger.atlassian.net/browse/M2-6112) and that seems to be working as expected.
However, when testing [M2-5769](https://mindlogger.atlassian.net/browse/M2-5769) and clicking the Export button, nothing happens. Is that expected? The AC says

> should behave exactly the same as clicking the "Export Data" item in the ... actions menu on the participants overview table

https://www.loom.com/share/8f7c2a1b723e4c8a8c01f2e99efc6946

JustinNusca:
@mbanting Definitely worked initially — looks like the suggested change regarding conditionally rendering the Dialog broke it, and I didn't verify the entire flow locally before pushing the changes. I amended that commit to revert this change, so that functionality should be restored now 👍

mbanting:
> @mbanting Definitely worked initially — looks like the suggested change regarding conditionally rendering the Dialog broke it, and I didn't verify the entire flow locally before pushing the changes. I amended that commit to revert this change, so that functionality should be restored now 👍

Working now!

mbanting (approved):
Nice job @JustinNusca !

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
IhorNesterenko99: @mbanting Hi! That sounds good. Let's go ahead and add this rule.

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,51 @@
...
+      >
+        {t('export')}
+      </Button>
+
+      <Link
+        data-testid="header-option-settings-link"
+        to={generatePath(page.appletSettings, { appletId })}
+      >
+        <Svg id="settings" fill={isSettingsSelected ? variables.palette.primary : ''} />
+      </Link>
```
farmerpaul: Just noticed a few little visual discrepancies here:
- I think `StyledFlexTopCenter` may be a better choice here than baseline-aligned items, even though the original code was aligned that way. Also, with `sx`, the theme spacing multiplier is automatically applied to numeric values, so you can just use `sx: {{ gap: 1 }}`
- The `startIcon` SVG should be 18x18 rather than 24x24
- The `Link`'s default `display` of `inline` is causing the SVG to be a little misaligned; I think you can get away with adding `<Link style={{ display: 'flex' }} …` there to fix that (it doesn't support `sx`)
- Bonus: It looks like the settings button is supposed to have a hover style, as it's technically an `icon-button` in Figma. We may not have this component, and not sure if our existing `Button` has icon-only support (with that particular hover style), but just wanted to clock that. May warrant a ticket in the backlog/for followup.

src/modules/Dashboard/pages/ParticipantDetails/ParticipantDetails.tsx
```diff
@@ -71,13 +70,19 @@ export const ParticipantDetails = () => {
...
               marginX: theme.spacing(2.4),
               marginBottom: theme.spacing(1.2),
+              placeContent: 'space-between',
             }}
           >
-            <StyledHeadlineLarge>{respondent?.result.secretUserId}</StyledHeadlineLarge>
-            <StyledHeadlineLarge color={palette.outline}>
-              {respondent?.result.nickname}
-            </StyledHeadlineLarge>
+            <Box sx={{ display: 'flex', gap: theme.spacing(1.6) }}>
```
farmerpaul: I might use `<StyledFlexTopCenter sx={{ gap: 1.6 }}>` here, but NBD.

farmerpaul (approved):
Just some nitpicky visual things I noticed, but otherwise great work. Happy to preapprove pending those suggestions.

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
JustinNusca: I pushed up another commit that should fix the UI issues you described.

Regarding the Figma `icon-button`: I opted to swap to using the `IconButton` from MUI for handling that link. It provides hover/active styles that admittedly aren't themed correctly currently, but it looks like we already have [a task for next sprint](https://mindlogger.atlassian.net/browse/M2-6071) dedicated to updating button variants to match the ones defined in Figma specifically. 

I figure, as you've suggested, adding appropriate theme values for the IconButton component could be a follow-up, and handled as part of that task.

farmerpaul: Perfect, works for me!

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -0,0 +1,53 @@
...
+        onClick={handleOpenExport}
+        startIcon={<Svg id="export" width={18} height={18} />}
+        sx={{ color: variables.palette.on_surface_variant }}
+      >
+        {t('export')}
+      </Button>
+
+      <IconButton
+        component={Link}
+        sx={{ width: 48, height: 48 }}
```
IhorNesterenko99:
```suggestion
        sx={{ width: '4.8rem', height: '4.8rem' }}
```
Just a small suggestion for consistency: use rem instead of px.

IhorNesterenko99 (approved):

PR #1646, opened 2024-04-10
[M2-6040][M2-5711][M2-5768]: Take Now flow

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-6040](https://mindlogger.atlassian.net/browse/M2-6040)
🔗 [Jira Ticket M2-5711](https://mindlogger.atlassian.net/browse/M2-5711)
🔗 [Jira Ticket M2-5768](https://mindlogger.atlassian.net/browse/M2-5768)

This PR introduces several new things to kick off the "Take Now" flow. They are: the `TakeNowModal` component, the `useTakeNowModal` hook, and the `enableMultiInformantTakeNow` feature flag. In this PR, I also enable launching of the modal from the Applet Activities Overview and Participant Detail Activities Tab screens.

#### The `TakeNowModal` component and `useTakeNowModal` hook

Use of the modal is restricted through the use of the use of the `useTakeNowModal` hook. I did this so that the consumer of the modal could control its presence by simply calling the `openTakeNowModal` function, which is returned by the hook. This is an intentional departure from the way modals are currently implemented, which has been to expose control of the `open` state to the parent.

The modal component itself only accepts an `onClose` prop, so that the caller may be notified of when the modal is closed. Otherwise, the modal controls its own open state by unmounting itself

The `openTakeNowModal` function accepts an activity, and optional default values for the subject and participant fields.

image

#### The `enableMultiInformantTakeNow` feature flag

This feature flag controls whether the take now feature is exposed. By default, the take now menu option is presented to users with the roles of `Admin`, `Manager`, and `Coordinator`. The option will be generally be missing for any other user type, and it will be missing for all users when this flag is turned off.

image

#### Activities Overview

https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/41eeaf84-bdc8-4fa5-8dc9-8b52e64eaf34

#### Participant Details Activities Tab

https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/c4b6285e-91c8-4f32-bf10-cd4925449268

### 🪤 Peer Testing

> [!NOTE] 
> You'll need to have both the respondent web app (`dev` branch) and this app running. 

- Sign in to the admin app with an admin user.
- Navigate to the applet activities overview tab (`/dashboard/{appletId}/activities`)
- Click the more menu (⋮) on one of the activities and note the presence of the `Take Now` option
- Click the `Take Now` option. The "Take Now" modal should now appear.
- Check that only the admin user is populated in the "Who is responding?" dropdown
- Check that all participants appear in the "Who is this activity about?" dropdown
- Click "Start Activity". You should be taken to the respondent web app at the beginning of the activity you selected earlier (you may be asked to authenticate first)
- Navigate to the participant details activities tab and repeat the steps from earlier. Once you click on the take now menu option, the modal should have that participant pre-populated in the "Who is this activity about?" dropdown

> [!WARNING]
> The participant details page currently only works for participants with full accounts

### ✏️ Notes

This PR makes use of the `REACT_APP_WEB_URI` environment variable. This variable is not new, it just wasn't documented in the `.env.example` file, so I've added it


src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
```diff
@@ -46,13 +63,15 @@ export const useActivityGrid = (dataTestId: string) => {
...
       },
       takeNow: ({ context }: MenuActionProps<ActivityActionProps>) => {
-        // TODO: Implement Take Now
-        // https://mindlogger.atlassian.net/browse/M2-5711
-        const { activityId } = context || {};
-        alert(`TODO: Take now (${activityId})`);
+        const { activityId } = context || { activityId: '' };
+        const activity = getActivityById(activityId);
+        if (activity) {
+          console.log('Opening modal with activity:', activity.id);
```
IhorNesterenko99, line 69: Would it be necessary to keep this console.log statement in the code?

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledDropdown.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box } from '@mui/material';
+import React from 'react';
```
IhorNesterenko99, line 2: Please move the import from React to the top.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import { Tooltip } from 'shared/components';
+import { SelectController, SelectUiType } from 'shared/components/FormComponents';
+import { StyledFlexColumn, StyledTitleMedium, StyledTitleTooltipIcon } from 'shared/styles';
+
+export type Option = {
```
IhorNesterenko99, line 8: Please move types to LabeledDropdown.types.ts

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.styles.ts
```diff
@@ -0,0 +1,23 @@
+import { Box, styled } from '@mui/material';
+
+import { theme } from 'shared/styles';
+
+export const StyledTakeNowModalContent = styled(Box)`
```
IhorNesterenko99, line 5:
```suggestion
export const StyledTakeNowModalContent = styled(StyledFlexColumn)`
```

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 0.8rem;
+  padding: ${theme.spacing(2.4, 3.2)};
+`;
+
+export const StyledTakeNowModalHeader = styled(Box)`
```
IhorNesterenko99, line 12:
```suggestion
export const StyledTakeNowModalHeader = styled(StyledFlexTopCenter)`
```

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  align-items: center;
+  gap: 2.4rem;
+`;
+
+export const StyledTakeNowModalBody = styled(Box)``;
```
IhorNesterenko99, line 18: Please remove.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledDropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+  options,
+  value,
+  placeholder,
+  onChange,
+}: LabeledDropdownProps) => (
+  <StyledFlexColumn sx={{ gap: 1.6, marginTop: 2.4 }}>
+    <Box sx={{ display: 'flex', gap: 0.4 }}>
+      <StyledTitleMedium fontWeight="bold">{label}</StyledTitleMedium>
+      <Tooltip tooltipTitle={tooltip}>
+        <Box sx={{ height: 24 }}>
```
IhorNesterenko99, line 36:
```suggestion
        <Box sx={{ height: "2.4rem" }}>
```

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const [participants, setParticipants] = useState<Option[]>([]);
+  const [events, setEvents] = useState<CreatedEvent[]>([]);
+  const { ownerId } = workspaces.useData() || {};
+  const userData = auth.useData();
+  const { appletId } = useParams();
+
+  const { execute: getParticipants } = useAsync(getWorkspaceRespondentsApi, (response) => {
+    if (response?.data) {
+      const data = (response.data as ParticipantsData).result;
+      console.log('Participants:');
```
IhorNesterenko99, line 34: Please remove console logs.

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
```diff
@@ -40,11 +41,6 @@ export const Activities = () => {
     };
   }, [appletId, getActivities]);
 
-  const activities = useMemo(
-    () => (activitiesData?.result ?? []).map((activity) => formatRow(activity)),
-    [activitiesData, formatRow],
-  );
-
   useEffect(() => {
```
sultanofcardio: I noticed this and am wondering if this is an unnecessary `useEffect`. If so, I'll probably remove it. @farmerpaul what do you think?

farmerpaul: Yeah, I don't understand the purpose either. I included it because Cleber had decided to keep it when he started building this page, but I assume he also just derived it from another list view. I think I was hoping someone would criticize it during review like you're doing now. 😅 (thank you!)

You can find the same thing elsewhere:
- [`Managers.tsx`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/multiinformant-metapod/src/modules/Dashboard/features/Managers/Managers.tsx#L181-L186)
- [`Respondents.tsx`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/multiinformant-metapod/src/modules/Dashboard/features/Respondents/Respondents.tsx#L390-L395)
- [`Participants.tsx`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/multiinformant-metapod/src/modules/Dashboard/features/Participants/Participants.tsx#L443-L448) (I guess we copied it)

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
```diff
@@ -2,22 +2,26 @@ import { useState } from 'react';
...
-  'data-testid': dataTestid,
+  rows,
+  TakeNowModal,
+  'data-testid': dataTestId,
 }: ActivityGridProps) => {
   const navigate = useNavigate();
   const { t } = useTranslation('app');
   const { appletId, participantId: _participantId } = useParams();
+  const workspaceRoles = workspaces.useRolesData();
+  const roles = appletId ? workspaceRoles?.data?.[appletId] : undefined;
```
ChaconC: `roles` is unused

src/shared/utils/testUtils.ts
```diff
@@ -1,9 +1,55 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
+export const mockSchema = <T extends unknown>(
+  data: T | null = null,
+  meta?: MetaSchema,
+): BaseSchema<T | null> => ({
+  status: 'success',
+  requestId: 'requestId',
+  ...meta,
+  data,
+});
```
farmerpaul: With these new test utilities defined, I'm actually not able to run the app anymore; I get the runtime error `Uncaught (in promise) Error: Do not import '@jest/globals' outside of the Jest test environment`. (Presumably there's something slightly different in my build environment that this error occurs for me but not for you?) In any case, to solve it, I'd suggest we avoid including the line `export * from 'testUtils.ts'` in `src/shared/utils/index.ts` because it ends up bundling Jest-only code into any modules that import from the `src/shared/utils` base dir. This change will require a little tweaking of `import` lines found in test suites.

.env.example
```diff
@@ -2,3 +2,4 @@ REACT_APP_API_DOMAIN=http://localhost:8000
 REACT_APP_ENV='dev'
 REACT_APP_DEVELOP_BUILD_VERSION=
 REACT_APP_LAUNCHDARKLY_CLIENT_ID=65df9142fbf14e10954ac55d
+REACT_APP_WEB_URI=http://localhost:3000
```
farmerpaul: 3000 is the admin app port; shouldn't it be this?
```suggestion
REACT_APP_WEB_URI=http://localhost:5173
```

README.md
```diff
@@ -88,13 +88,14 @@ See the section about [running tests](https://facebook.github.io/create-react-ap
...
-| ------------- | ------------- | ------------- | ------------- |
-| REACT_APP_API_DOMAIN               | yes | null | MindLogger Backend API base URL |
-| REACT_APP_ENV                      | no | null | Environment to run the app in (`prod` or `staging`)  |
-| REACT_APP_DEVELOP_BUILD_VERSION    | no | null | Footer app build number |
-| REACT_APP_MIXPANEL_TOKEN           | no | null | Mixpanel token |
-| REACT_APP_LAUNCHDARKLY_CLIENT_ID           | no | null | LaunchDarkly client key to choose target environment |
+| Key                              | Required | Default value | Description                                          |
+|----------------------------------|----------|---------------|------------------------------------------------------|
+| REACT_APP_API_DOMAIN             | yes      | null          | MindLogger Backend API base URL                      |
+| REACT_APP_WEB_URI                | yes      | null          | Base URL of the MindLogger respondent web app        |
```
farmerpaul: Just to align with the `.env.example` file…
```suggestion
| REACT_APP_WEB_URI                | yes      | http://localhost:5173 | Base URL of the MindLogger respondent web app        |
```

src/resources/app-fr.json
```diff
@@ -1173,11 +1174,19 @@
...
   "takePhoto": "Prendre une photo",
-  "takeNow": "Prendre maintenant",
+  "takeNow": "Prendre Maintenant",
+  "takeNowModalParticipantLabel": "Qui répond?",
+  "takeNowModalParticipantPlaceholder": "Sélectionnez parmi les participants et les membres de l'équipe",
+  "takeNowModalParticipantTooltip": "Il s'agit du participant qui réalisera l'activité.",
+  "takeNowModalSubjectLabel": "Qui est cette activité à propos?",
+  "takeNowModalSubjectPlaceholder": "Sélectionnez parmi les participants",
+  "takeNowModalSubjectTooltip": "C’est le sujet de l’activité assignée.",
+  "takeNowNotFound": "Pas Trouvé",
```
farmerpaul: Nit: Although I've observed that this is pretty inconsistent in the apps' language files, in French, word capitalization rules don't mirror English rules. As a general rule (though there's some nuance), you should assume that in French "title case", only the first letter of the _whole phrase_ is capitalized, whereas in English title case, the first letter of _each word_ (apart from short prepositions and conjunctions) is capitalized. So this is actually more correct:
```json
  "takeNow": "Prendre maintenant",
  "takeNowNotFound": "Pas trouvé",
```

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.types.ts
```diff
@@ -4,8 +4,12 @@ import { Roles } from 'shared/consts';
 import { DatavizActivity } from 'api';
 import { Order } from 'shared/types';
 
+import { TakeNowModalProps } from '../TakeNowModal/TakeNowModal.types';
+import { FeatureFlags } from '../../../../shared/types/featureFlags';
```
farmerpaul:
```suggestion
import { FeatureFlags } from 'shared/types/featureFlags';
```

.env.example
ChaconC:
```suggestion
REACT_APP_WEB_URI=http://localhost:5173
```

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
```diff
@@ -1,18 +1,28 @@
...
+  mockedAppletId,
+  mockedOwnerId,
+  mockedRespondent,
+  mockedRespondent2,
+  mockedUserData,
+} from 'shared/mock';
 import { getPreloadedState } from 'shared/tests/getPreloadedState';
-import { renderWithProviders } from 'shared/utils';
+import { mockSchema, mockSuccessfulHttpResponse, renderWithProviders } from 'shared/utils';
+import { mockGetRequestResponses } from 'shared/utils/testUtils';
```
ChaconC: You can import `mockGetRequestResponses` from `shared/utils`

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+        buttonText={t('startActivity')}
+        onSubmit={handleSubmit}
+        onClose={handleClose}
+        disabledSubmit={subject === null || participant === null}
+        submitBtnVariant={SubmitBtnVariant.Contained}
+        footerStyles={{ padding: '0.8rem 3.2rem 3.2rem' }}
+        data-testid={`${dataTestId}-take-now-modal`}
+      >
+        <StyledTakeNowModalContent>
+          <StyledTakeNowModalHeader>
```
farmerpaul: Nit: You could get away with avoiding a `.styles.ts` file just be making use of existing flexbox shared components (I think it's been said that up to 2 style attributes passed via `sx` is fine):

```suggestion
        <StyledFlexColumn sx={{ gap: 0.8, p: theme.spacing(2.4, 3.2) }}>
          <StyledFlexTopCenter sx={{ gap: 2.4 }}>
```

src/shared/utils/testUtils.ts
```diff
@@ -1,9 +1,55 @@
+import mockAxios, { HttpResponse } from 'jest-mock-axios';
```
ChaconC: This is giving me an error causing the app to crash (blank page on `npm run start`)

CleanShot 2024-04-15 at 14 02 24

Removing this import removes the error. Probably related to removing `src/shared/tests/axios-mock.ts` as it was an excluded path

ChaconC (commented):
I'm unable to test the Take Now flow. The web app opens and immediately redirects me back to the admin app.
Is this the expected behavior?

src/shared/components/Modal/Modal.styles.ts
```diff
@@ -15,20 +15,19 @@ export const StyledDialog = styled(Dialog)`
...
   && {
-    padding: ${theme.spacing(2, 3.2, 3.4)};
+    padding: ${theme.spacing(3.2, 3.2, 0.8)};
   }
 `;
 
 export const StyledCloseButton = styled(StyledClearedButton)`
-  position: absolute;
-  top: 1.5rem;
-  right: 3.2rem;
```
farmerpaul, line 31: Thank you 😍 

For it to align even more pixel-perfectly, I believe it should also have an additional 4px margin (Screenshot 2024-04-15 at 5 11 49 PM), which will also bump the title down a few pixels to better match Figma:
```tsx
  margin: 0.4rem;
```

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
```diff
@@ -2,22 +2,26 @@ import { useState } from 'react';
...
 import { StyledButton } from './ActivityGrid.styles';
 import { ActivityGridProps } from './ActivityGrid.types';
 
 export const ActivityGrid = ({
-  rows,
   isLoading = false,
-  'data-testid': dataTestid,
+  rows,
+  TakeNowModal,
+  'data-testid': dataTestId,
```
farmerpaul, line 17: Nit: While I get the rationale for uppercasing the `I`, FWIW, the lowercase `i` version is actually much more common in the codebase at the moment. I suppose the basis for the lowercase `i` choice is it being the result of a pure kebab-case to camelCase conversion. But I agree it's a bit odd to have this be the only identifier in the codebase where we consistently do not camelCase `…Id`, so if that's the consensus, I'm fine with uppercasing it from now on.

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (userData) {
+      // For now, the admin is the only member of this list
+      let whoIsResponding = `${userData.user.id}`;
+      if (userData.user.firstName) {
+        whoIsResponding += ` (${userData.user.firstName}`;
+        if (userData.user.lastName) {
+          whoIsResponding += ` ${userData.user.lastName[0].toUpperCase()}.`;
+        }
+        whoIsResponding += ')';
+      }
```
farmerpaul: We should probably confirm what the grey text is meant to be showing [in Figma](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=3915-232753&mode=design&t=Ujj22OLyVsESGTcK-4). First name and abbreviated last name make intuitive sense, however the existing Respondents screen, as well as the new Participants screen, currently show nickname, not first & last name. There are many screens where a summarized version of a participant is shown, which I anticipated in my PR #1652 by creating a component, `ParticipantSnippet` to use for this very purpose (though maybe I should have called it `UserSnippet`). I may have assumed wrong what props it should take, but I settled with secret user ID, nickname, and tag, since that's what we're already showing in the Respondents and Participants tables. We should at least get some agreement as to the purpose of the `nickname` property and make sure to use it as intended in the UI.

Further to this: I noticed you opted to format these dropdowns in plain text, without the rich text formatting shown in Figma:

Screenshot 2024-04-15 at 5 33 24 PM

I get that supporting this styling this could be a bit of a chore and you may have wanted to avoid it in your PR, but is there a separate ticket to take care of it?

farmerpaul (commented):
Nice work @sultanofcardio! I ran into a couple issues trying to test, but figured them out and included some suggestions. Otherwise it behaved as intended. Also added some silly nits, take 'em or leave 'em. 😛

The only thing I think we might need to get more clarity on is what is exactly going to be shown in user snippets, as seen in the dropdowns in your PR, as well as the top of the Upgrade Account modal in [my PR](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1652) (and eventually in several other screens). Look forward to your thoughts!

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: Back when I was implementing this part of the modal, the dropdown list looked like this



I hadn't updated my implementation because the design was still changing and I assumed I should work with the version that was there before. Although admittedly, I omitted the avatar icon for the same reason.

I think the question about nickname vs first and last name is a fair one. I'll ask about it in Slack, and I'll just update the component to match the current design

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
sultanofcardio: I would like to opt for uppercasing the I from now on, unless there's a good reason for sticking with the current convention. But for this PR, I'm going to change it back

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: Good call 👌

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.types.ts
sultanofcardio: I really need to fix this auto-import situation 😅

src/resources/app-fr.json
sultanofcardio: Thanks for the tip. The translation came out as you suggested, and I manually changed it. I'll keep that in mind from now on 👍

.env.example
sultanofcardio: Good catch 🙏 I made the same mistake in my `.env` file, but forgot to fix it here

sultanofcardio: Good catch 🙏 I made the same mistake in my `.env` file, but forgot to fix it here

src/shared/utils/testUtils.ts
sultanofcardio: Thanks for this suggestion, I was just trying to figure this out. I hadn't tried to run the app since I wrote the tests

src/shared/utils/testUtils.ts
sultanofcardio: Paul suggests [this change](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1646#discussion_r1566299508). I haven't fixed it yet, but hopefully this can help you resume your testing while I implement the fix

README.md
sultanofcardio: I left this one null because the default value for `REACT_APP_API_DOMAIN` is also null, even though there's a value for it in `.env.example`. I assumed this was because the way we run the app locally varies, but putting a value here makes sense

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.types.ts
farmerpaul: If you know of a trick to make VS Code always write auto-import lines as desired – I'd love to hear it! I kinda think it would require writing a custom extension though…

sultanofcardio:
> I'm unable to test the Take Now flow. The web app opens and immediately redirects me back to the admin app. Is this the expected behavior?

@ChaconC that must be a result of the wrong port being used in the value of `REACT_APP_WEB_URI` that I provided. Changing it to the value you suggested should give the right behavior. Were you able to test with that value?

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.types.ts
sultanofcardio: These particular ones were done by WebStorm. I'm using both because the git experience in WebStorm is just too good 🙈

src/shared/components/Modal/Modal.styles.ts
sultanofcardio: Those are those eagle eyes you've got there mate 🎉

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
sultanofcardio: Actually, I'm gonna do a 180 here and keep the uppercase I. I see that Carlos has already started this change, and I don't want to go backwards

src/shared/components/Modal/Modal.styles.ts
farmerpaul: Lol… I have a bit of a reputation

src/shared/utils/testUtils.ts
ChaconC: Thanks, I got the PR in a working state similar to @farmerpaul's approach 👍

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: Looks like [these are actually nicknames](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x?type=design&node-id=3915-232728&mode=dev#770527901), so I'm going to steal your `ParticipantSnippet` component

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledDropdown.tsx
```diff
@@ -0,0 +1,95 @@
+import { Autocomplete, Box, TextField } from '@mui/material';
+import React, { useCallback, useRef, useState } from 'react';
```
IhorNesterenko99:
```suggestion
import { useCallback, useRef, useState } from 'react';
```
Please move the import from React to the top.

sultanofcardio: Hey @IhorNesterenko99. I don't understand this suggestion. Why do I need to move it to the top?

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
```diff
@@ -1,5 +1,5 @@
 import { format } from 'date-fns';
-import { useMemo, useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
```
IhorNesterenko99, line 1: Could you please move the import from React to the top, and also in other places?

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
IhorNesterenko99, line 1:
```suggestion
```

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.types.ts
```diff
@@ -0,0 +1,10 @@
+import { BoxProps } from '@mui/material/Box/Box';
+import React from 'react';
+
+export type ParticipantSnippetProps<T extends React.ElementType> =
```
IhorNesterenko99: In my opinion, it's better to import elements from 'react' individually rather than importing the entire library, like this: ``import { ElementType } from 'react';`` Please verify this approach in all other places.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -0,0 +1,107 @@
+import { Autocomplete, Box, TextField } from '@mui/material';
+import React, { useCallback, useRef, useState } from 'react';
```
IhorNesterenko99: No need to import React in this component since it doesn't use any JSX that requires it. Please verify if this applies to other components as well.

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+describe('ParticipantSnippet component', () => {
+  test('should render the snippet', () => {
+    render(<ParticipantSnippet {...props} />);
+
+    expect(screen.getByText(props.secretId)).toBeInTheDocument();
+    expect(screen.getByText(props.nickname)).toBeInTheDocument();
+    expect(screen.getByText(props.tag)).toBeInTheDocument();
+  });
+});
```
IhorNesterenko99: Please add a test case for when nickname and tag are not provided.

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
IhorNesterenko99:
```suggestion
import { ElementType } from 'react';
```

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+  OpenTakeNowModalOptions,
+  TakeNowModalProps,
+  UseTakeNowModalProps,
+} from './TakeNowModal.types';
+import { StyledImageContainer, StyledImg } from '../ActivitySummaryCard/ActivitySummaryCard.styles';
+import { LabeledUserDropdown } from './LabeledDropdown/LabeledUserDropdown';
+import { BaseActivity } from '../ActivityGrid';
+import { ParticipantDropdownOption } from './LabeledDropdown/LabeledUserDropdown.types';
+
+export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
```
IhorNesterenko99, line 25: Please separate the code into the hook and the component itself. If the file name is ``SomeComponent.tsx``, I expect to import ``SomeComponent``, not the hook from it. The hook can be placed in ``SomeComponent.hooks.ts(x)``.
Alternatively the hook can be placed to the ``modules\components\hooks``, leaving the component folder only for the component itself.
Also, please ensure that the functionality is covered by unit tests.

sultanofcardio: The choice to colocate the code was deliberate. I can rename the file to `TakeNowModal.hooks.tsx`. Let me know if that works

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
```diff
@@ -9,17 +9,36 @@ import { DateFormats } from 'shared/consts';
...
   getActivityActions,
-  ActionsObject,
+  StyledSvg,
 } from 'modules/Dashboard/components/ActivityGrid';
+import { useLaunchDarkly } from 'shared/hooks/useLaunchDarkly';
+
+import { useTakeNowModal } from '../TakeNowModal/TakeNowModal';
 
-export const useActivityGrid = (dataTestId: string) => {
+export const useActivityGrid = (dataTestId: string, activitiesData: ActivitiesData | null) => {
```
IhorNesterenko99, line 23: Please cover the hook with unit tests.

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.utils.tsx
```diff
@@ -50,7 +57,7 @@ export const getActivityActions = ({
       action: takeNow,
       title: t('takeNow'),
       context: { appletId, activityId },
-      isDisplayed: true,
+      isDisplayed: canDoTakeNow,
       'data-testid': `${dataTestId}-activity-take-now`,
     },
   ];
```
IhorNesterenko99, line 65: Please cover these cases with unit tests.

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.tsx
```diff
@@ -11,14 +11,14 @@ import { StyledButton } from './ActivityGrid.styles';
 import { ActivityGridProps } from './ActivityGrid.types';
 
 export const ActivityGrid = ({
```
IhorNesterenko99, line 13: Please cover component with unit tests.

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    expect(screen.getByText(props.secretId)).toBeInTheDocument();
+    expect(screen.getByText(props.nickname)).toBeInTheDocument();
+    expect(screen.getByText(props.tag)).toBeInTheDocument();
+  });
+
+  test('should render with only the secretId', () => {
+    render(<ParticipantSnippet secretId={props.secretId} />);
+
+    expect(screen.getByText(props.secretId)).toBeInTheDocument();
```
IhorNesterenko99: I suggest adding exceptions like this: 
``expect(screen.queryByText(props.nickname)).not.toBeInTheDocument();``

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+import unionBy from 'lodash/unionBy';
+import { useTranslation } from 'react-i18next';
+
+import { Tooltip } from 'shared/components';
+import { StyledFlexColumn, StyledTitleMedium, StyledTitleTooltipIcon, theme } from 'shared/styles';
+import { ParticipantSnippet } from 'modules/Dashboard/components/ParticipantSnippet';
+
+import { LabeledUserDropdownProps, ParticipantDropdownOption } from './LabeledUserDropdown.types';
+
+export const LabeledUserDropdown = ({
```
IhorNesterenko99, line 12: Please add test coverage.

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
IhorNesterenko99: I suggest moving the JSX component itself to TakeNowModal.tsx and using it in the hook. This will make the hook code cleaner and separate from the JSX.

sultanofcardio: I understand what you're suggesting. I'm saying that the hook is meant to create and return the component when called, so I can't move it to a separate file

IhorNesterenko99: Why not, just use the component with provided props and import it to the hook.

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+        setActivity(null);
+        window.open(url.toString(), '_blank');
+      }
+    }, [subject, participant]);
+
+    if (!activity || !participants || !participantsAndTeamMembers) {
+      return null;
+    }
+
+    return (
```
IhorNesterenko99, line 116: Here you can return <TakeNowModal .... />

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledDropdown.tsx
IhorNesterenko99: Hi @sultanofcardio. If you take a look at the files in the project, the first import is always from React, then the others. It's a good idea to implement highlighting and auto-fixing with ESLint and Prettier, as was suggested by @mbanting.

farmerpaul: Hi @IhorNesterenko99. I understand the convention for having React as the 1st import of a `.tsx` file, however there's no rule in our lint configuration that dictates this specifically, so IDEs would not automatically reorder this line, nor show an error about it – at least mine doesn't, and I have both ESLint and Prettier enabled. I'm quite certain @sultanofcardio also has ESLint and Prettier enabled in his dev environment as he's the one who was tasked to unify linting rules across the repos a couple months back.

I haven't figured out how to tweak `.eslintrc`'s `import/order` rules to always ensure `react` import is first unless it's in its own group (separated by an extra newline), which seems unnecessary. But if you have any ideas for automating this I'd welcome it!

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: My intention with this hook + component combination is that the consumer of the `TakeNowModal` doesn't have to manage the state of its visibility or the what is currently being displayed. It's meant to reduce the surface area of interaction to just one function: `openTakeNowModal`. This is similar to how we show banners without needing to explicitly manage their visibility.

Separating the component from the hook would make that impossible because of all the props it would need to accept. I can't see a way to do that while preserving the interaction model. Can you suggest a way to do this?

IhorNesterenko99: I'm having difficulty understanding why we can't move the JSX to its own component. However, I'll mark this comment as resolved.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.types.ts
```diff
@@ -0,0 +1,28 @@
+import { AutocompleteProps } from '@mui/material/Autocomplete/Autocomplete';
+
+import { ParticipantSnippetProps } from 'modules/Dashboard/components/ParticipantSnippet';
+
+export type ParticipantDropdownOption = Omit<
+  ParticipantSnippetProps<'li'>,
+  'boxProps' | 'data-testid'
+> & {
+  id: string;
+};
```
farmerpaul:
```suggestion
export type ParticipantDropdownOption = ParticipantSnippetInfo & {
  id: string;
};
```

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <ParticipantSnippet<'li'>
+            {...psProps}
+            boxProps={{
+              sx: { p: theme.spacing(0.6, 1.6), cursor: 'pointer' },
+              ...props,
+            }}
+          />
+        )}
+        getOptionLabel={(value) =>
+          `${value.secretId}${value.nickname ? ` (${value.nickname})` : ''}`
```
farmerpaul, line 93: Would the effort to allow `Autocomplete` to accept a `<ParticipantSnippet>` for its currently selected value be too much for this PR? I see that there are probably some structural/UX considerations; i.e. presumably it would only show when focus is blurred because the user has to be able to access the (plain-text) search input field when focused.

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.test.tsx
farmerpaul, line 1: Thank you 🙏🏻

farmerpaul (approved):
Just had a question/brainstorm about the `LabeledUserDropdown`, but otherwise changes look great! Thanks for writing all those tests. 🔥

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: I should've mentioned this. I didn't see a straightforward way to implement it without doing something fancy like you described (especially since Matt stressed that we should try to reuse existing component as much as possible),and I didn't want to drag out the PR any more, so I fell back to just using plain text.

I'd love to get this matching the design eventually though

PR #1656, opened 2024-04-16
[M2-6246] Rename Participant "role" to "tag"

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-6246](https://mindlogger.atlassian.net/browse/M2-6246)

This PR includes some very minor changes, partly because the desired changes were already mostly done, and because tags are mostly still unimplemented:
- Where referring to a Participant's "tag" (for which BE support was recently added), use the word "Tag" instead of "Role", both in the code and the UI
  - Note: in Figma it's still incorrectly shown as "Role" – that's due to be fixed in [this task](https://mindlogger.atlassian.net/browse/M2-6271)
- Move the "Tag" column to the correct location on the Participants screen

Also, unrelated:
- Fix a TypeScript error introduced in the last merge

### 📸 Screenshots

| Screenshot 2024-04-16 at 5 10 01 PM | Screenshot 2024-04-16 at 5 09 01 PM |
| Screenshot 2024-04-16 at 5 09 48 PM | Screenshot 2024-04-16 at 5 09 24 PM |

### DIFF
--- a/src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
+++ b/src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
@@ -1,15 +1,18 @@
+import { BoxTypeMap } from '@mui/system';
+import { ElementType } from 'react';
+
 import { Chip } from 'shared/components';
 import { StyledBodyLarger, StyledFlexTopCenter, variables } from 'shared/styles';
 
 import { ParticipantSnippetProps } from './ParticipantSnippet.types';
 
-export const ParticipantSnippet = ({
+export const ParticipantSnippet = <T extends ElementType = BoxTypeMap['defaultComponent']>({
   secretId,
   nickname,
   tag,
   boxProps,
   'data-testid': dataTestId,
-}: ParticipantSnippetProps) => {
+}: ParticipantSnippetProps<T>) => {
   const { sx, ...rest } = boxProps ?? {};
 
   return (
--- a/src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
+++ b/src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
@@ -24,7 +24,7 @@ export const ParticipantSnippet = ({
         </StyledBodyLarger>
       )}
       {/*
-      TODO: add respondent tag/label (with appropriate Chip/Tag component) when available
+      TODO: add respondent tag (with appropriate Chip/Tag component) when available
       https://mindlogger.atlassian.net/browse/M2-5861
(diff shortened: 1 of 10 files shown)


src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
```diff
@@ -1,18 +1,15 @@
-import { ElementType } from 'react';
-import { BoxTypeMap } from '@mui/system';
-
 import { Chip } from 'shared/components';
 import { StyledBodyLarger, StyledFlexTopCenter, variables } from 'shared/styles';
 
 import { ParticipantSnippetProps } from './ParticipantSnippet.types';
 
-export const ParticipantSnippet = <T extends ElementType = BoxTypeMap['defaultComponent']>({
+export const ParticipantSnippet = ({
```
sultanofcardio: We have to keep the type parameter here, otherwise typescript complains about this props spread

https://github.com/ChildMindInstitute/mindlogger-admin/blob/7460397c388471a5692244087f997ce2094612c2/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx#L83-L89

I thought I'd be able to get around this by specifying the `component` prop, but the typing information doesn't seem to propagate to the `ParticipantSnippetProps`, which maintains its default value

farmerpaul: Yup, I can see why this would fail.

What do you think about just making do with `ParticipantSnippet` always rendering a `div`, and wrapping yours in an `<li>`, just to keep things a little simpler? Otherwise I think we'll have to propagate that awkward generic syntax in several contexts… maybe it's not as bad as I think, but it seemed to be having a bit of a domino effect when I was trying to resolve it earlier.

sultanofcardio: I can live with wrapping it in a `<li>`, but I think the generic syntax is only necessary if you're trying to spread props that you don't control. Like in this instance, the Autocomplete is expecting a list item element. Do you foresee this applying anywhere else?

IhorNesterenko99 (approved):

src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx
farmerpaul: I solved it by doing a minor refactor (extracted model props from the component props), so I was able to preserve the generic syntax and it shouldn't affect your code. :boom:

sultanofcardio (approved):

PR #1657, opened 2024-04-16
[M2-5609] Update DashboardTable/Team Listing View

JustinNusca:
### 📝 Description

🔗 [M2-5609](https://mindlogger.atlassian.net/browse/M2-5609): [FE][Teams Overview] Team Listing View

This PR updates the appearance of the Dashboard Table to match the specs from Figma. While the associated task is intended for the "Team Listing View", the table UI itself is reused across many of the table views in the various applet tabs, and some of the UI changes I've made impacts those screens as well.

In addition to the changes to the Dashboard Table UI, which mostly consist of border and padding updates, I've also tweaked the columns shown on the Team Listing view. This includes:

- Removing the pins column/ability to pin a row, per the Figma specs.
- Updating the "Role" column to be titled in the singular, and to render the role as a `Chip` component.
- Adding a "Checkmark" column, intended for performing bulk operations
- Removing the title of the "Actions" column
- Re-ordering columns

#### Team Listing

| Before | After |
|-|-|
| team-before | team-after |

[M2-5609]: https://mindlogger.atlassian.net/browse/M2-5609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.styles.ts
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.styles.ts
@@ -1,6 +1,7 @@
 import { styled, TableCell } from '@mui/material';
 
 import { TableHead } from 'shared/components';
+import { theme } from 'shared/styles';
 import { variables } from 'shared/styles/variables';
 import { shouldForwardProp } from 'shared/utils/shouldForwardProp';
 
@@ -23,7 +24,7 @@ export const StyledTableCell = styled(TableCell, shouldForwardProp)`
       `;
     }
 
-    return { paddingTop: 12, paddingBottom: 12 };
+    return { paddingTop: theme.spacing(1.2), paddingBottom: theme.spacing(1.2) };
   }}
 `;
 
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.styles.ts
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.styles.ts
@@ -1,13 +1,17 @@
 import { styled, TableCell } from '@mui/material';
 
+import { TableHead } from 'shared/components';
 import { variables } from 'shared/styles/variables';
 import { shouldForwardProp } from 'shared/utils/shouldForwardProp';
 
 export const StyledTableCell = styled(TableCell, shouldForwardProp)`
-  font-size: ${variables.font.size.md};
-  line-height: ${variables.font.lineHeight.md};
+  font-size: ${variables.font.size.lg};
   letter-spacing: ${variables.font.letterSpacing.sm};
-  word-break: break-all;
+  line-height: ${variables.font.lineHeight.lg};
+
+  &&:first-of-type {
+    padding-left: 2rem;
+  }
 
   ${({ hasColFixedWidth, width }: { hasColFixedWidth?: boolean; width?: string }) => {
     if (hasColFixedWidth) {
@@ -19,8 +23,11 @@ export const StyledTableCell = styled(TableCell, shouldForwardProp)`
       `;
     }
 
-    return `
-      height: 4.8rem;
-    `;
+    return { paddingTop: 12, paddingBottom: 12 };
   }}
 `;
+
+export const StyledTableHead = styled(TableHead)({
+  '&& tr:nth-of-type(2) th': { padding: '1.2rem' },
+  '&& tr th:first-of-type': { paddingLeft: '2rem' },
+});
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
@@ -1,8 +1,7 @@
-import { Table as MuiTable, TableBody, TablePagination, TableRow } from '@mui/material';
+import { Box, Table as MuiTable, TableBody, TablePagination, TableRow } from '@mui/material';
 
 import {
   EmptyState,
-  TableHead,
   UiType,
   StyledTableCellContent,
   StyledTableContainer,
@@ -10,7 +9,7 @@ import {
 import { DEFAULT_ROWS_PER_PAGE } from 'shared/consts';
 
 import { DashboardTableProps } from './DashboardTable.types';
-import { StyledTableCell } from './DashboardTable.styles';
+import { StyledTableCell, StyledTableHead } from './DashboardTable.styles';
 
 export const DashboardTable = ({
   columns,
@@ -52,10 +51,11 @@ export const DashboardTable = ({
       uiType={uiType}
       hasColFixedWidth={hasColFixedWidth}
       onScroll={onScroll}
+      sx={{ height: rows?.length ? 'auto' : '100%' }}
     >
       {!!rows?.length && (
         <MuiTable stickyHeader data-testid={dataTestid}>
-          <TableHead
+          <StyledTableHead
             headCells={columns}
             order={order}
             orderBy={orderBy}
@@ -78,9 +78,13 @@ export const DashboardTable = ({
                     sx={{ cursor: row[key].onClick ? 'pointer' : 'default' }}
                     data-testid={`${dataTestid}-${index}-cell-${key}`}
                   >
-                    {row[key].contentWithTooltip
-                      ? row[key].contentWithTooltip
-                      : row[key].content(row)}
+                    <Box
+                      sx={{ overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap' }}
+                    >
+                      {row[key].contentWithTooltip
+                        ? row[key].contentWithTooltip
+                        : row[key].content(row)}
+                    </Box>
                   </StyledTableCell>
                 ))}
               </TableRow>
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
@@ -1,4 +1,4 @@
-import { Box, Table as MuiTable, TableBody, TablePagination, TableRow } from '@mui/material';
+import { Table as MuiTable, TableBody, TablePagination, TableRow } from '@mui/material';
 
 import {
   EmptyState,
@@ -7,6 +7,7 @@ import {
   StyledTableContainer,
 } from 'shared/components';
 import { DEFAULT_ROWS_PER_PAGE } from 'shared/consts';
+import { StyledEllipsisText } from 'shared/styles';
 
 import { DashboardTableProps } from './DashboardTable.types';
 import { StyledTableCell, StyledTableHead } from './DashboardTable.styles';
@@ -78,13 +79,11 @@ export const DashboardTable = ({
                     sx={{ cursor: row[key].onClick ? 'pointer' : 'default' }}
                     data-testid={`${dataTestid}-${index}-cell-${key}`}
                   >
-                    <Box
-                      sx={{ overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap' }}
-                    >
+                    <StyledEllipsisText>
                       {row[key].contentWithTooltip
                         ? row[key].contentWithTooltip
                         : row[key].content(row)}
-                    </Box>
+                    </StyledEllipsisText>
                   </StyledTableCell>
                 ))}
               </TableRow>
--- a/src/modules/Dashboard/features/Managers/Managers.tsx
+++ b/src/modules/Dashboard/features/Managers/Managers.tsx
@@ -9,13 +9,12 @@ import {
   Avatar,
   AvatarUiType,
   Chip,
-  ChipShape,
   MenuActionProps,
   Search,
   Spinner,
-  Svg,
 } from 'shared/components';
 import { banners, workspaces } from 'redux/modules';
+import { StyledMaybeEmpty } from 'shared/styles/styledComponents/MaybeEmpty';
 import { useAsync, usePermissions, useTable } from 'shared/hooks';
 import { DashboardTable, DashboardTableProps } from 'modules/Dashboard/components';
 import { Manager } from 'modules/Dashboard/types';
@@ -26,8 +25,8 @@ import { useAppDispatch } from 'redux/store';
 
 import { ManagersRemoveAccessPopup, EditAccessPopup } from './Popups';
 import { ManagersTableHeader } from './Managers.styles';
-import { getManagerActions, getHeadCells } from './Managers.utils';
 import { ManagersData } from './Managers.types';
+import { getManagerActions, getHeadCells } from './Managers.utils';
 
 export const Managers = () => {
   const { t } = useTranslation('app');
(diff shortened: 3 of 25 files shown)


src/modules/Dashboard/components/DashboardTable/DashboardTable.styles.ts
```diff
@@ -19,8 +23,11 @@ export const StyledTableCell = styled(TableCell, shouldForwardProp)`
       `;
     }
 
-    return `
-      height: 4.8rem;
-    `;
+    return { paddingTop: 12, paddingBottom: 12 };
```
IhorNesterenko99: Please use theme.spacing for paddings and margins. Please also check other occurrences.

src/modules/Dashboard/features/Managers/Managers.tsx
```diff
@@ -116,52 +107,110 @@ export const Managers = () => {
...
               if (ownerId === id || !filteredManager?.applets?.length) {
                 return;
               }
 
               return (
-                <ActionsMenu
-                  menuItems={getManagerActions(actions, filteredManager)}
-                  data-testid="dashboard-managers-table-actions"
-                />
+                <Box display="flex" justifyContent="flex-end" alignItems="center" width="100%">
```
IhorNesterenko99: Please consider using StyledComponents for styling instead of inline properties to maintain consistency with the current codebase.

src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
```diff
@@ -78,9 +78,13 @@ export const DashboardTable = ({
                     sx={{ cursor: row[key].onClick ? 'pointer' : 'default' }}
                     data-testid={`${dataTestid}-${index}-cell-${key}`}
                   >
-                    {row[key].contentWithTooltip
-                      ? row[key].contentWithTooltip
-                      : row[key].content(row)}
+                    <Box
+                      sx={{ overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap' }}
+                    >
```
farmerpaul: Looks like there's a preexisting shared component for this very purpose: `StyledEllipsisText`

Also just a note that I've seen it suggested in past reviews to restrict inline styling (preferably via `sx`, as you've done here) to no more than 2 attributes; for more than that, it warrants its own Styled component.

src/modules/Dashboard/features/Managers/Managers.utils.tsx
```diff
@@ -17,42 +19,61 @@ export const getHeadCells = (id?: string): HeadCell[] => {
```
farmerpaul, line 13: This comment is referring to `ManagersColumnsWidth` enum above. ☝🏻
It seems you've removed any references to the enum (which I can understand being unnecessary if the values are not being repeated anywhere), but as such we can get rid of the enum definition, too.

src/modules/Dashboard/features/Managers/Managers.tsx
```diff
@@ -116,52 +107,110 @@ export const Managers = () => {
...
+          editAccessAction: ({ context: user }: MenuActionProps<Manager>) => {
+            setSelectedManager(user || null);
+            setEditAccessPopupVisible(true);
+          },
+        };
+        const emptyContent = (
+          <Box component="span" sx={{ color: variables.palette.outline_variant2 }}>
+            --
+          </Box>
+        );
```
farmerpaul: Please use the existing styled component, `StyledMaybeEmpty`, for this purpose, which you'll find is used in a couple other spots (such as [Participants table](https://github.com/ChildMindInstitute/mindlogger-admin/blob/8aee5c3f90428f1bb779ff1723ec3bf6e34c057f/src/modules/Dashboard/features/Participants/Participants.tsx#L335)).

src/modules/Dashboard/features/Managers/Managers.tsx
```diff
@@ -116,52 +107,110 @@ export const Managers = () => {
...
+        const actions = {
+          removeAccessAction: ({ context: user }: MenuActionProps<Manager>) => {
+            setSelectedManager(user || null);
+            setRemoveAccessPopupVisible(true);
+          },
+          editAccessAction: ({ context: user }: MenuActionProps<Manager>) => {
+            setSelectedManager(user || null);
+            setEditAccessPopupVisible(true);
+          },
+        };
```
farmerpaul: Could these be defined once at the `useMemo` function body level rather than for each element of the array?

src/modules/Dashboard/features/Participants/Participants.const.tsx
```diff
@@ -1,5 +1,6 @@
 export enum ParticipantsColumnsWidth {
-  Pin = '4.8rem',
+  Pin = '8rem',
+  Actions = '3.2rem',
```
farmerpaul: This newly added `Actions` enum value doesn't seem to be referenced? There's already a `Menu` enum value for use with the actions column, though it has a different width than what you have here…

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -276,9 +284,12 @@ export const Participants = () => {
...
             icon={
-              <StyledCheckBox>
-                <Svg id="checkbox-empty-outlined" height="20" width="20" />
-              </StyledCheckBox>
+              <Svg
+                fill={variables.palette.outline_variant2}
+                height="20"
+                id="checkbox-empty-outlined"
+                width="20"
+              />
```
farmerpaul: It seems we'll be standardizing on this styling of checkboxes at least in tables, and probably universally in the admin panel as a whole. Would it be worth creating a shared `Checkbox` component that defaults to this SVG for the `icon` prop?

src/shared/styles/variables/palette.ts
```diff
@@ -14,6 +14,7 @@ export const palette = {
   on_surface_variant: '#42474e',
   outline: '#72777f',
   outline_variant: '#c4c9d0',
+  outline_variant2: '#C2C7CF',
```
farmerpaul: Nit 😇 
```suggestion
  outline_variant2: '#c2c7cf',
```

src/shared/styles/theme.ts
```diff
@@ -289,7 +289,8 @@ export const theme = createTheme({
       styleOverrides: {
         root: {
           color: variables.palette.on_surface_variant,
-          padding: '0.6rem 1.2rem',
+          padding: '0.2rem 0.8rem',
+          height: 24,
```
farmerpaul:
```suggestion
          height: '2.4rem',
```

src/shared/styles/theme.ts
```diff
@@ -305,16 +306,8 @@ export const theme = createTheme({
...
-            border: `${variables.borderWidth.md} solid ${variables.palette.outline}`,
             color: variables.palette.on_surface_variant,
-            backgroundColor: 'transparent',
-            '&:hover': {
-              backgroundColor: variables.palette.on_surface_variant_alfa8,
-            },
-            a: {
-              color: variables.palette.on_surface_variant,
-            },
+            backgroundColor: variables.palette.inverse_on_surface,
```
farmerpaul: Thanks for doing this styling! You may be aware already, but a more comprehensive update of the Chip component is going to be captured in [M2-6161](https://mindlogger.atlassian.net/browse/M2-6161).

Did you confirm it's okay to remove the hover state for this color variant? Just noting that the `TagsController` looks like it might use `Chip` in a more interactive manner.

If you are confident we can remove the hover state here, then it would make sense to update this other instance of `Chip` where I manually disabled hover state: [ParticipantSnippet.tsx](https://github.com/ChildMindInstitute/mindlogger-admin/blob/8aee5c3f90428f1bb779ff1723ec3bf6e34c057f/src/modules/Dashboard/components/ParticipantSnippet/ParticipantSnippet.tsx#L35-L40)

src/shared/components/Table/TableHead/TableHead.types.ts
```diff
@@ -3,6 +3,7 @@ import { HeadCell, Order } from 'shared/types/table';
 import { UiType } from '../Table.types';
 
 export type TableHeadProps = {
+  className?: string;
```
farmerpaul, line 6: I'm not seeing where you've introduced code that uses this new `className` prop. Just curious of its purpose.

src/modules/Dashboard/features/Managers/Managers.tsx
```diff
@@ -116,52 +107,110 @@ export const Managers = () => {
...
               return (
-                <ActionsMenu
-                  menuItems={getManagerActions(actions, filteredManager)}
-                  data-testid="dashboard-managers-table-actions"
-                />
+                <Box display="flex" justifyContent="flex-end" alignItems="center" width="100%">
+                  <ActionsMenu
+                    menuItems={getManagerActions(actions, filteredManager)}
+                    data-testid="dashboard-managers-table-actions"
+                  />
```
farmerpaul, line 195: I added the ability to use the new color variant of the `...` button for `ActionsMenu` a few PRs ago via `buttonColor="secondary"`. Would you mind updating it here and in the Participants table? 🙏🏻

farmerpaul (commented):
Hey @JustinNusca, I just had a handful of small suggestions/questions, but great work!!

src/shared/components/Table/TableHead/TableHead.types.ts
JustinNusca: It's used [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1657/commits/1baa9aab8849da8cb175ba9b57c86cd18bde5c13#diff-73665bee8296711d9fe42d0b7e90eb2d25901aa022259d7acd6b7a4465c14098R30-R33), to allow the classname that the StyledComponent creates to be passed to the StyledComponent that wraps the MUI TableHead.

src/shared/styles/theme.ts
JustinNusca: I spent a bunch of time on this today, and I think I've basically taken care of [M2-6161](https://mindlogger.atlassian.net/browse/M2-6161) in doing so 😅.

In the additional commit I pushed up, I:

- Re-implemented the hover styles, but am only conditionally applying them when the Chip is interactive (more specifically, when a `tabindex=0` attribute it present, which MUI automatically adds when an `onClick` or `onDelete` prop is passed to it. This should cover the usage in `TagsController`.

- Updated my revisions to the theme so that hover states appear correct for each color (I followed existing styles where I could, but no hover states are provided for the revised styles in Figma, so I wound up improvising a little bit and did my best to stay true to the new colors).

- Altered the prop interface for Chips to allow the set of MUI chip props in order to support the "variant" prop. The old "secondary" styles used a grey border, instead of a grey background. The MUI Chip's `variant` can be either `"filled"` or `"outlined"` that is intended to control this behaviour, so opted to disambiguate the existing styles and allow for `color` and `variant` to be set separately. This was my attempt at providing some backwards compatibility with these old styles — none of the new instances of the Chip in Figma appear to _use_ an `"outlined"` variant, and I'm not using it in my implementation, but if for some reason we need to revert to the previous "secondary" Chip styling, it can be opted into on an individual basis by setting that prop.

Chips look like so, now:

Screenshot 2024-04-18 at 6 22 13 PM

src/modules/Dashboard/features/Participants/Participants.const.tsx
JustinNusca: Holdover from a previous attempt at modifying the column widths that I neglected to remove 😅

src/shared/components/Table/TableHead/TableHead.types.ts
farmerpaul: Oh interesting, didn't realize that would be needed. TIL, thx!

src/modules/Dashboard/features/Managers/Managers.utils.tsx
```diff
@@ -8,33 +8,13 @@ import { variables } from 'shared/styles';
...
-            <Svg
-              fill={variables.palette.outline_variant2}
-              height="20"
-              id="checkbox-empty-outlined"
-              width="20"
-            />
-          }
-        />
-      ),
+      label: <Checkbox aria-label={t('checkAll')} checked={false} />,
```
farmerpaul, line 17: Just wondering if it was intentional to omit `data-testid="dashboard-managers-checkbox-header"`?

src/shared/styles/utils/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  const key = isFilled ? 'backgroundColor' : 'borderColor';
+
+  switch (color) {
+    case 'default':
+    case 'secondary':
+      return {
+        color: variables.palette.on_surface_variant,
+        [key]: variables.palette.inverse_on_surface,
+
+        '&[tabindex="0"]:hover': {
```
farmerpaul: Maybe helpful to include a brief comment (maybe outside the `switch`) to explain how the tabindex trick works here (what you mentioned in the PR comment)?

src/shared/styles/utils/index.ts
farmerpaul, line 1: Nit: This is just convention, but as yet we haven't been defining functions directly in an `index.ts` in this app, and they're mostly used to export modules defined in sibling files.

I think in this case, since this function pertains directly to theming, I'd probably place this code in `src/shared/styles/theme.utils.ts` (which is an existing convention).

src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
farmerpaul: Just checking if you're still planning to use `StyledEllipsisText` here in a future commit…

src/shared/styles/theme.ts
farmerpaul: That's fantastic @JustinNusca, thanks for doing that! IMO you might as well assign yourself to M2-6161 as well as add it to this PR's description (so that it gets auto-linked to this PR).

I think the only other part of completing M2-6161 would be to make sure the chips under the Account Type column on the [Participants screen](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=1628-123599&mode=design&t=mywOdvfdkUyr1NOM-4) is fully styled to match Figma, which essentially just means changing the "Pending Invite" chip to "✉︎ Pending" and using the appropriate color, if that's something you wouldn't mind squeezing in?

farmerpaul (approved):
Just a few last minor comments, but preapproving!

src/shared/styles/utils/index.ts
JustinNusca: Added a comment, [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1657/commits/48f16b23b2d473c8eac4bc3440b286d8fdb6ddb4#diff-afff194d342e79376a7b2c5c8e2b8e8a417a62ad672a514afe6fe8e17a4a6fdaR22-R24)!

src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
JustinNusca: Whoops, I think I mixed the original comment up with `StyledMaybeEmpty` while making updates and overlooked it — [updated, here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1657/commits/48f16b23b2d473c8eac4bc3440b286d8fdb6ddb4#diff-fced307897174d1f11dc037c5464650394ae7b42ef41cd611af2c63ed4bd1e8fR82-R86)

src/shared/styles/theme.ts
JustinNusca: Oh sure! Updated:

Screenshot 2024-04-19 at 11 16 15 AM

src/modules/Dashboard/features/Managers/Managers.utils.tsx
JustinNusca: Yeah, I copied it over from the existing similar column in `Participants`, and just renamed it initially.

But this isn't used for any tests yet, because no bulk operation functionality has been implemented yet. I figured I'd remove unused attributes to reduce potential confusion, and let whoever ends up responsible for adding this feature take ownership and add it if the need it (who knows if this Checkbox will exist as-is when it comes to actually building the feature)?

PR #1680, opened 2024-04-24
[M2-5771] Add Schedule view to Participant details

JustinNusca:
### 📝 Description

🔗 [M2-5771](https://mindlogger.atlassian.net/browse/M2-5771): [Participant Detail] Schedule View for Default and Individual

This PR adds the Schedule view to the appropriate tab of the Participant details screen. If the specific participant has an individual schedule, the view will default to showing their individual schedule. Otherwise, functionality should remain the same as the standalone Schedule page for an applet.

As part of this change, I needed to update a number of the schedule components. Previously, individual schedule components all defined their own handlers for particular bits of scheduling-related functionality, which all operated on the assumption that these components would be used on the existing `/schedule` route. 

- To address this, I swapped from using parameters in the route (eg `/schedule/:userId`) to query parameters (eg `?user=:userId`, in order to allow for individual schedule selection functionality to work irrespective of the current route.
- In some other cases, I added props for callback handlers in order to allow for some functionality to be handled at a higher level (ie the screen component), rather than lower level ones that are now used in multiple places.
- I did a bit of renaming here, as I found the ambiguity of respondent ID difficult to grapple with here: I renamed the route parameter for the participant details screen to `:subjectId`, and certain other props & variables throughout the Schedule components to specify _user_ ID. My hope was to make it clearer what records these IDs related to, which are needed for certain components/API endpoints.

### 📸 Screenshots

| Before | After |
|-|-|
| localhost_3000_dashboard_2f25b302-89ef-47ac-b4c5-b0dc31b2a86d_participants_2263efd9-8ea6-43fb-8f14-6f5e1bd3e25d_schedule_user=4477dcbb-cc37-4233-a05b-18a132e14448(iPad Pro) |  |

[M2-5771]: https://mindlogger.atlassian.net/browse/M2-5771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/features/Participant/Activities/Activities.tsx
```diff
@@ -90,14 +92,16 @@ export const Activities = () => {
...
+    <Box sx={{ padding: `${theme.spacing(3.2)} ${theme.spacing(1.6)}` }}>
+      <ActivityGrid
+        rows={activities}
+        TakeNowModal={TakeNowModal}
+        order="desc"
+        orderBy=""
+        isLoading={isLoading || isSubjectLoading}
+        data-testid={dataTestId}
+      />
+    </Box>
```
sultanofcardio: Why the padding change here? This is laid out a bit different structurally than in [Figma](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=1558-147074&mode=dev), but if we're going for spacing consistency I think it should be ``<Box sx={{ padding: `${theme.spacing(3.2)}` }}>`` right?

sultanofcardio (approved):
All good with one nit! Tbh the UX of the unpopulated search popup on the participant detail schedule page is a little weird, but I understand that we don't want to change that at this time

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
JustinNusca: Ah, you're right. I removed the padding provided by the `TabPanel` so that the Schedule tab could have the correct amount (ie, none — that tab's padding is provided by it's child content, to allow borders to extend to the edges of the tab), and re-added it here for the Activities tab. 

The padding here is equivalent to the spacing on the Schedule tab, and I assumed it was uniform across tabs, but apparently not. It should be updated in the [additional commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1680/commits/1b7bf99f4265bce8aa6aecfcd757c12d809b2736) I pushed up.

JustinNusca: This _does_ make me realize that the horizontal padding for the header, on both the Dashboard and Participant Details screens are off. 

In Figma, it looks like it should be a uniform 32px, except for the breadcrumbs, while in implementation it varies between 24 and 34px.

PR #1681, opened 2024-04-24
[M2-5582] Remove checkbox columns from Participant & Manager tables

JustinNusca:
### 📝 Description

🔗 [M2-5582](https://mindlogger.atlassian.net/browse/M2-5582): [Participant Overview] Pinning

The linked task is for adding pinning to the Participants table in the applet Dashboard. This already exists and appears to work as expected. It also, however, specifies that the functionality should match the pinning feature in the Managers table.

This previously existed, but I recently removed it in https://github.com/ChildMindInstitute/mindlogger-admin/pull/1657, because Pins were neither included in the linked designs from that task, _or_ the acceptance criteria, so I assumed it was deprecated functionality and removed it in order to match the new design specs. 😅

Based on the wording of M2-5582, I think this was accidental, so I restored the existing functionality in this PR.

[M2-5582]: https://mindlogger.atlassian.net/browse/M2-5582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Participants/Participants.utils.tsx
+++ b/src/modules/Dashboard/features/Participants/Participants.utils.tsx
@@ -1,5 +1,4 @@
 import { Dispatch, SetStateAction } from 'react';
-import { Checkbox } from '@mui/material';
 import { t } from 'i18next';
 
 import { Svg } from 'shared/components/Svg';
@@ -132,15 +131,8 @@ export const getHeadCells = (id?: string): HeadCell[] => {
 
   return [
     {
-      id: 'checkbox',
-      label: <Checkbox aria-label={t('checkAll')} checked={false} />,
-      enableSort: true,
-      width: ParticipantsColumnsWidth.Pin,
-    },
-    {
       id: 'pin',
       label: '',
-      enableSort: true,
       width: ParticipantsColumnsWidth.Pin,
     },
     {
--- a/src/modules/Dashboard/features/Managers/Managers.test.tsx
+++ b/src/modules/Dashboard/features/Managers/Managers.test.tsx
@@ -9,7 +9,6 @@ import {
   mockedOwnerId,
   mockedManager,
   mockedEmail,
-  mockedManagerId,
 } from 'shared/mock';
 import { Roles } from 'shared/consts';
 import { initialStateData } from 'shared/state';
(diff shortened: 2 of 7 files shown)


sultanofcardio (approved):

JustinNusca:
@sultanofcardio Per the discussion in [Slack](https://metalab.slack.com/archives/C06EKU790TY/p1714075766398369), I pushed up an additional commit to [remove non-functional checkboxes from Participant & Manager tables](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1681/commits/534c34b105d3e47bdc82683846d7cb8b02366f96).

src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -131,16 +130,9 @@ export const getHeadCells = (id?: string): HeadCell[] => {
...
-    {
-      id: 'checkbox',
-      label: <Checkbox aria-label={t('checkAll')} checked={false} />,
-      enableSort: true,
-      width: ParticipantsColumnsWidth.Pin,
-    },
     {
       id: 'pin',
       label: '',
-      enableSort: true,
```
JustinNusca, line 143: I also removed the "Sort" button from the Pin column on the Participants table, since that appeared to defeat the purpose of pinning, and isn't included in the list of sortable columns in Figma.

sultanofcardio:
@JustinNusca I'm a bit confused now. Your summary in Slack says there shouldn't be any pinning in the managers table, but this PR still re-adds it anyway. I'm not against it, but I'm just not clear on what you're supposed to be doing. It sounds to me that this PR should just be removing the checkboxes?

JustinNusca:
Oh geez, I dunno what I was thinking, but it seems you're better at understanding my own words than I am. 😅

You're absolutely right and I was moving quickly without thinking things through.

The changeset in this PR should now only include the checkbox removal. Pinning is retained in the Participants table. The re-addition of pins to the Managers table has been removed.

sultanofcardio (approved):
LGTM! 🚀

PR #1688, opened 2024-04-29
[M2-6462] Rename LaunchDarkly to FeatureFlags

ChaconC:
### 📝 Description

Renames use of "LaunchDarkly" to more general "FeatureFlags"

🔗 [Jira Ticket M2-6462](https://mindlogger.atlassian.net/browse/M2-6462)

Changes include:

- Renamed `useLaunchDarkly` to `useFeatureFlags`
- Renamed `flags` to `featureFlags`

### 🪤 Peer Testing

- Update `.env` with `REACT_APP_LAUNCHDARKLY_CLIENT_ID` (Check [Confluence](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/487227432/Using+LaunchDarkly+with+MindLogger#Keys) for IDs)
- Add the `testing-flag` to `FeatureFlagsKeys`. This key changes from `false` to `true` after logging in
  ```javascript
  export const FeatureFlagsKeys = {
    ...
    testingFlag: 'testingFlag',
  };
  ```
- Check values change by logging the values of `featureFlags` from the `useFeatureFlags` hook


farmerpaul (approved):
Changes look good, so approving! 🚀 

The only thing is that when testing, `testingFlag` doesn't turn to `true` after logging in, on my end. The other feature flags are being set correctly. I confirmed the behaviour is the same on the `feature/multiinformant-metapod` branch. Maybe there's something different about the admin app not triggering that flag to be toggled?

ChaconC:
> Changes look good, so approving! 🚀
> 
> The only thing is that when testing, `testingFlag` doesn't turn to `true` after logging in, on my end. The other feature flags are being set correctly. I confirmed the behaviour is the same on the `feature/multiinformant-metapod` branch. Maybe there's something different about the admin app not triggering that flag to be toggled?

Yes, on the admin app we still have the rule where the workspace name needs to include `multi-informant` to get the features enabled.

PR #1687, opened 2024-04-26
[M2-6231] Add Team Member

farmerpaul:
- [x] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-6231](https://mindlogger.atlassian.net/browse/M2-6231)

Created the Add Team Member popup, which is accessed from **Applet > Team > Add Team Member**.

Related changes include:
- Added a `limitTagRows` prop to the `TagsAutocompleteController` component to assign a maximum scrollable height
- Updated input helper/error text font size to align with Figma, and allowed it to word-wrap if needed rather than be truncated with "…" ([Figma convo](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x?type=design&node-id=2185-51645&mode=design#773960311) for context)
- Simplified styling of table header on Participants tab to better align with structure used elsewhere

> [!TIP]
> Other minor changes are explained in commit messages, so review [individual commits](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1687/commits) for those descriptions.

### 📸 Screenshots

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/62e813a0-1bd2-410c-b16f-675c8b0065fe

### 🪤 Peer Testing

Please see [JIRA task's AC](https://mindlogger.atlassian.net/browse/M2-6231) for detailed expectations while testing.

> [!NOTE]
> Currently, team members that have been invited but have not yet accepted the invitation are **not** listed as "pending" on the Team tab, and will be covered in a separate task TBD.

Prerequisites:
- Have an existing applet created in your account that has several participants (10+).
- Have **no managers (team members)** added to your workspace yet

1. Navigate to that applet's **Team** tab, and press the new **Add Team Member** button
    **Expected outcome:** The Add Team Member popup should appear, with the Workspace Name field present and prepopulated with your account name.
2. Fill out fields, including leaving some empty or providing an invalid email address format.
    **Expected outcome:** Validation errors should appear accordingly, and submit button is disabled until form is valid.
3. Fill form out to add a new team member with **Manager** role, and submit.
    **Expected outcome:** A success banner should appear, however the success message will show **undefined undefined** for the first and last name due to a BE bug, which will be [addressed here](https://mindlogger.atlassian.net/browse/M2-6159).
4. Accept the invitation for the new manager you've invited, and refresh the Team page.
    **Expected outcome:** The new team member should be listed.
5. Start adding another team member
[...]

PR #1691, opened 2024-04-30
[M2-5625]: [FE][Activity Overview] Update Layout to Separate Flows Block

JustinNusca:
### 🔗 Related Issues
- [M2-5625](https://mindlogger.atlassian.net/browse/M2-5625): [FE][Activity Overview] Update Layout to Separate Flows Block

### 📝 Description

This PR updates the "Activities" screen in both the Applet & Participant dashboards to include a visually-separated list of Activity Flows, where applicable.

As part of this change, I:

- Broke up parts of the existing ActivityGrid component into isolated components, so I could re-compose them as needed to arrange the updated UI.
    - More specifically, I [moved the existing toolbar into an `ActivitiesToolbar` component](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/a3ca65b83969850e82b1a1a1155e98200c81e4ba). This component is still mostly placeholder, so it amounted to shifting some JSX around. 
    - The ActivityGrid now only renders the list of activity cards, instead of being responsible for most of the page layout.
- Added new components, including:
    -  [An `ActivitiesSectionHeader` component](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691#:~:text=M2%2D5625%3A%20Add%20ActivitiesSectionHeader%20component.). This renders a simple header that sections off the Flows & Activity grids.
    - [A `FlowSummaryCard` component](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/3e044060a28931b3ac2d1762e300b23cfc0c1179). This is similar to the existing `ActivitySummaryCard`, but for ActivityFlows, and is used in the... 
    - [FlowGrid component](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/2a995ee9b7b322100492eb2fa5da7785784a3055), which is the "Flow" version of the existing `ActivityGrid`.

I have updated both `Activity` screen components so that they fetch an applet's activityFlows (via the `applet/${appletId}` endpoint), and render the appropriate UI using the above components. As part of these changes, I also:

- [Removed enforced padding from the Applet dashboard's tab panels.](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/4835b6bc13d92468d11c6cd5a5f2df97e4453baf)
  - This is similar to a change I made on the Participants dashboard screen as well. Because we want the scrollable area of the page to be relegated to the content beneath the tabs (which is the current behaviour), the existing padding on these tab panels caused page content to be cut-off before reaching the edge of the screen when this content was scrollable. This change removes and re-adds the padding on the inner scrollable content area.
- Added [a new icon](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/afe9403647b351e46ce51e0d63b47bada5cbdc16) and [palette color](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/54f98aeea50482a8110c9edb0873434b249143ec).

### 📸 Screenshots

| Before | After |
|-|-|
| localhost_3000_dashboard_df3e3cea-a341-4db9-968c-158318863d39_activities (1) | localhost_3000_dashboard_df3e3cea-a341-4db9-968c-158318863d39_activities | 

[M2-5625]: https://mindlogger.atlassian.net/browse/M2-5625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryCard.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+export const StyledRoot = styled(Box)({
+  border: `1px solid ${variables.palette.surface_variant}`,
+  borderRadius: '1.6rem',
+  display: 'flex',
+  gap: theme.spacing(2.4),
+  overflow: 'hidden',
+  padding: theme.spacing(2.4),
+  placeContent: 'space-between',
+  placeItems: 'center',
+});
```
farmerpaul: Nit: This is quite similar to the [container styles](https://github.com/ChildMindInstitute/mindlogger-admin/blob/c00d79af536978012d63851a0f41da693e09b287/src/modules/Dashboard/components/ActivitySummaryCard/ActivitySummaryCard.styles.ts#L11-L18) defined in `ActivitySummaryCard`, though might not be worth literally sharing styling per se. Just thinking you could get flexbox attributes for free using `StyledFlexSpaceBetween`, and use more of those theme tokens (border width/radius). 🤷🏻‍♂️

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryThumbnail/FlowSummaryThumbnail.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+  borderRadius: 8,
+  display: 'flex',
+  flexShrink: 0,
+  flexWrap: 'wrap',
+  height: theme.spacing(8),
+  overflow: 'hidden',
+  placeContent: 'flex-start',
+  placeItems: 'flex-start',
+  position: 'relative',
+  width: theme.spacing(8),
```
farmerpaul: Nit: For reasons I haven't yet understood, I've seen PR review feedback requesting that only gaps & spacing to use `theme.spacing`, while width & height should be in `rem` units. 🤷🏻‍♂️ Again I don't know that this is documented anywhere (I think that's supposed to be an action item somewhere), just sharing what I've encountered.

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryThumbnail/FlowSummaryThumbnail.styles.ts
```diff
@@ -0,0 +1,43 @@
+import { Box, styled } from '@mui/material';
+
+import { theme, variables } from 'shared/styles';
+
+export const StyledRoot = styled(Box)({
+  background: variables.palette.primary_container,
+  borderRadius: 8,
```
farmerpaul:
```suggestion
  borderRadius: variables.borderRadius.md,
```

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryThumbnail/FlowSummaryThumbnail.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+  background: variables.palette.primary80,
+  borderRadius: '100%',
+  color: variables.palette.primary,
+  display: 'flex',
+  fontSize: variables.font.size.md,
+  fontWeight: variables.font.weight.bold,
+  height: '3.2rem',
+  inset: 0,
+  margin: 'auto',
+  padding: `${theme.spacing(0.8)} ${theme.spacing(1.2)}`,
```
farmerpaul:
```suggestion
  padding: theme.spacing(0.8, 1.2),
```

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryThumbnail/FlowSummaryThumbnail.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+export const StyledActivityThumbnail = styled('img')({
+  background: variables.palette.surface1,
+  width: '50%',
+  height: '50%',
+  position: 'relative',
+  objectFit: 'cover',
+  border: 'none',
+});
+
+export const StyledItemCount = styled('div')({
```
farmerpaul: Nit: It's common elsewhere to use `StyledFlexAllCenter` for center-aligned content, so you should be able to get rid of the `display`, `placeContent`, `placeItems` props (if you want).

Side note: I've noticed you you habitually use `place-content` over `justify-content`, and `place-items` over `align-items`. I'm not used to using them. I get that they're shorthands, but I suppose it doesn't occur to me to affect both axes by default – but maybe it's considered best practice these days and I'm just out of the loop! 😅 I'll switch to that if you think it's a good habit?

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
```diff
@@ -1,59 +1,93 @@
...
-      isLoading={isLoading}
-      order={'desc'}
-      orderBy={''}
-    />
+    <StyledFlexColumn sx={{ gap: 2.4, overflow: 'auto', p: 3.2 }}>
+      <ActivitiesToolbar appletId={appletId} data-testid={dataTestId} />
+
+      {isLoading ? (
+        <Spinner />
+      ) : (
```
farmerpaul: Nit: Just noting that we're either rendering the spinner, or the populated activities – however in many other screens, the spinner is just overlaid on top of the content (if any has been rendered yet).

As it is now (using either-or logic), if the activities are ever refreshed from the API after the page has first rendered (which won't happen here yet, but could later as we further build out this screen), then the existing rendered content would flicker away during refresh. NBD, but it might be nicer to avoid the flicker down the road.

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
```diff
@@ -87,16 +97,40 @@ export const Activities = () => {
...
-        isLoading={isLoading || isSubjectLoading}
-        data-testid={dataTestId}
-      />
-    </Box>
+    <StyledFlexColumn sx={{ gap: 2.4, overflow: 'auto', p: 3.2 }}>
+      <ActivitiesToolbar appletId={appletId} data-testid={dataTestId} />
+
+      {isLoading ? (
+        <Spinner />
+      ) : (
```
farmerpaul: Same as earlier comment

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
```diff
@@ -1,59 +1,93 @@
...
+        <StyledFlexColumn sx={{ gap: 4.8 }}>
+          {flows?.length && (
```
farmerpaul: So that `0` doesn't render when it's empty:
```suggestion
          {!!flows?.length && (
```

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
```diff
@@ -87,16 +97,40 @@ export const Activities = () => {
...
+        <StyledFlexColumn sx={{ gap: 4.8 }}>
+          {flows?.length && (
```
farmerpaul:
```suggestion
          {!!flows?.length && (
```

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+          icon={<Svg aria-hidden id="logic" height={18} width={18} />}
+          shape={ChipShape.Rectangular}
+          title={t('flow')}
+        />
+
+        <Chip
+          icon={<Svg aria-hidden id="checklist-filled" height={18} width={18} />}
+          shape={ChipShape.Rectangular}
+          title={t('activityCount', { count: activities?.length ?? 0 })}
+        />
```
farmerpaul, line 47: Nit: just noticing that Chips are a slightly different height in the `rectangular` variant vs. `rounded`:

Screenshot 2024-05-01 at 6 48 43 PM
Screenshot 2024-05-01 at 6 48 48 PM

I guess to make them align to Figma, you'd have to remove the static `height` and `lineHeight` as dictated in `theme.tsx`, and instead change `line-height` dynamically in `Chip.styles.ts` based on the `shape` prop (`variables.font.lineHeight.md` vs `variables.font.lineHeight.lg`).

Only thing is, if you make this change to all rectangular chips, it's going to affect the changes I made in #1687 for the Participants autocomplete field on the Add Team Member screen, where we rely on an expected height of rectangular chips to limit the height of the autocomplete field to 3 rows. Also maybe it will affect other screens unpredictably… hard to stay without some digging.

Maybe don't worry about it for now. I'm being too eagle-eyed! 🦅

farmerpaul (approved):
Great work, Justin, love all the changes! ✨ 

I identified basically just nits, a couple minor bugs but nothing major, so preapproving!

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
JustinNusca: Good catch!

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
JustinNusca: Hm, this was a bit more nuanced than it first appeared. I set up the loading states to mimic the existing first-time loading experience (eg, the toolbar and page UI appears with the spinner, and then the data renders when available), but you're right that I didn't account for the possibility of persisting data when re-fetching without leaving the page.

That said, I'm also not sure the existing implementation allowed for this, either — the `useAsync` hook we're using here [clears any existing data](https://github.com/ChildMindInstitute/mindlogger-admin/blob/60bb60127f708a6c8ea01ecade05dae91a1ae459/src/shared/hooks/useAsync.ts#L22) whenever `execute` is called.

It's also a bit more nuanced than it first appears on this page, because there's now multiple parallel network requests that will return at different times, and I'm trying to avoid some disruptive shifting UI that occurs when one resolves before the other or vice/versa.

I've handled it in this PR by adding an apollo-client/useQuery-inspired `previousData` property to the return value of `useAsync`, [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1691/commits/f88da33f6ab9d8b147b022b086572450ce2c19c2), which I'm using both to persist data for subsequent requests, and infer nuanced loading state info (ie, if it is the first or subsequent time `execute` has been called) to render the UI as you're describing, here.

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryCard.tsx
JustinNusca: Yeah, I left this alone for now, mostly to minimize potential disruption. This seems worthwhile as a standalone task.

src/modules/Dashboard/components/FlowSummaryCard/FlowSummaryThumbnail/FlowSummaryThumbnail.styles.ts
JustinNusca: I'll be honest and say that I've started using it by default not out of a need for specify both axes/block and inline elements simultaneously, and more because I appreciate the synonymous nomenclature (ie, I typically sort lists of properties and it allows for these similar styles to get sorted together for easy reference).

If there were a real need to explicitly declare justify _and_ align for content _and_ items (which I've never really run into since I sorta fell into using `placeX` by default), I'd be inclined to use the explicit properties for the things I want to set.

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
```diff
@@ -1,73 +1,72 @@
...
   });
+  const formattedActivities = useMemo(
+    () => activities.map((activity) => formatRow(activity)),
+    [activities, formatRow],
+  );
 
   const isLoading = isLoadingActivities || isLoadingFlows;
+  const showContent =
+    (isLoading && previousActivitiesResponse?.data?.result?.activitiesDetails?.length > 0) ||
+    !isLoading;
```
farmerpaul, line 53: I appreciate all the effort you went to to improve the loading state UX! Who know it would end up becoming this complicated. 😅

I'm wondering if this loading state pattern, as you've implemented in the same way for both Activities screens, might make sense to – eventually – apply more broadly in other consumers of `useAsync`? Not that I think we should update all the screens to invoke this same logic yet. More that I see an opportunity to encapsulate these bits of logic into `useAsync`, rather than repeat for each screen, which feels just a little unwieldy.

- Perhaps `response ?? prevResponse` could be returned from the hook as, say, `cachedResponse`?
- Somehow encapsulating the logic in `showContent` in a more generic way, also returnable from the hook as, say, `hasCachedData`? (Does it need to delve into previous response data structure as you've done here, or is the truthiness of `prevResponse` enough?)

Just some thoughts to improve comprehension and reusability, not sure if it's practical! I'm totally fine to keep it as you've done it here.

farmerpaul (approved):
Thanks for making those changes @JustinNusca! I had one idea I added a comment for, but feel free to take it or leave it.

Also, I noticed that "Take Now" isn't working for the flows. There's an error in the web app saying the activity doesn't exist. I guess the web app should be looking up the activity flow in case it fails to discover an activity with that ID, given that there's only one query parameter `startActivityOrFlow` being used for both scenarios (unless there's a single BE call that also accepts an ambiguous ID?). Anyway worth clocking – I haven't checked if there's already a ticket for this.

src/modules/Dashboard/features/Applet/Activities/Activities.tsx
JustinNusca: Yeah — this `useAsync` seems like a great target for building more patterns around, because I do appreciate the way it returns state directly in a way that lets us avoid hook or callback-driven approaches (eg an `onSuccess` that sets the response data locally in the component). 

I was searching around the codebase and there doesn't seem to be much use of it in the way I am here, outside of using the `isLoading` flag.

I'll say one of the first changes I'd want to make is returning the data object directly, instead of the response, to reduce the amount of property drilling going on 😅.

> Somehow encapsulating the logic in showContent in a more generic way, also returnable from the hook as, say, hasCachedData? (Does it need to delve into previous response data structure as you've done here, or is the truthiness of prevResponse enough?)?

This was something I was thinking about already — the interface of this hook looks a lot like Apollo's [useLazyQuery](https://www.apollographql.com/docs/react/api/react/hooks/#uselazyquery), specifically. They return a `called` value which is analogous to your suggestion, though that gets set as soon as the initial request starts. TanStack's [useQuery](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) includes more nuanced fields akin to what we're trying to do here, like `isInitialLoading` distinct from `isLoading`, and `isFetched`, which is essentially the check we're trying to do here (ie: show whatever content we have, so long as it's not the initial load).

I think I'm not so inclined to make additional changes yet, but I think this is a good spot to have some eyes on going forward 👀👀. I think this is worth acting on if we start using the hook more in this way.

PR #1695, opened 2024-05-02
M2-6506: Fix handling of user ID in schedule popups.

JustinNusca:
### 🔗 Related Issues

[M2-6506](https://mindlogger.atlassian.net/browse/M2-6506): [Admin Panel] Issues with modifying events from Participant Detail Schedule tab

### 📝 Description

This PR fixes an issue with a few popups on the schedule view. As part of my recent changes, I refactored a number of schedule view components to use query parameters for determining if the user is viewing a schedule for a specific user, or viewing the default schedule. This was a change from the previous approach of using the presence of a respondentId in the route path to determine this, as this would not be true when viewing the schedule from the participant details route. 

I must have missed these dialogs when making this change, as they were still reliant on react router's `useParams`, and not the expected `useSearchParams`. As a result, they always behaved as though we were modifying the _default_ schedule, even when we were expecting to modify the _individual_ schedule. Honestly, I'm not 100% on how I overlooked this, given that I created and deleted a bunch of events via this same UI while working on this feature and I'm positive I made the relevant changes in create/edit dialogs previously 😕.

### 📸 Screenshots

| popup-before | popup-after |
| create-popup-before | create-popup-after |

[M2-6506]: https://mindlogger.atlassian.net/browse/M2-6506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Schedule/ClearScheduledEventsPopup/ClearScheduledEventsPopup.test.tsx
+++ b/src/modules/Dashboard/features/Applet/Schedule/ClearScheduledEventsPopup/ClearScheduledEventsPopup.test.tsx
@@ -49,10 +49,13 @@ describe('ClearScheduledEventsPopup', () => {
     const route = `/dashboard/${mockedAppletId}/schedule/${mockedRespondentId}`;
     const routePath = page.appletScheduleIndividual;
     mockAxios.delete.mockResolvedValueOnce(null);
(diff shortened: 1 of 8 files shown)


JustinNusca:
Closing this PR in favour of https://github.com/ChildMindInstitute/mindlogger-admin/pull/1704.

PR #1707, opened 2024-05-08
fix: Update button variants (M2-6071)

JustinNusca:
### 📝 Description

🔗 [M2-6071](https://mindlogger.atlassian.net/browse/M2-6071): [FE] Update Button variants to align with Figma

This PR implements all of the button variants depicted [here in Figma](https://www.figma.com/file/LXl9BAJSwtgWuzkxqOqvX8/ML1-Design-System?type=design&node-id=57868-142461&mode=design&t=YolB7mwxGKCMjrvK-4), a number of which have not previously been added.

I added additional values to the MUI Button's `variant` prop, which map to the new variants added, and match the naming of the Figma variants as closely as I could. In addition to the already-provided variant values of `'contained'`, `'outlined'`, and `'text'`, we now also support `'elevated'`, `'tonal'`, and `'textNeutral'`.

I've updated the styles of some existing button variants where necessary, to match the properties in Figma more closely, particularly the hover/focus styles, and the text color of the `'outlined'` variant.

I've also set the default value of the Button's variant prop to `'textNeutral'`, which is the closest visual match for how all the buttons without an explicit `variant` prop value previously appeared. The `text` variant, which was the previous default, was updated to have a blue text color, in order to more closely match the appearance in Figma.

The Button's `color` prop is still available, and may continue to be used in combination with the `contained` variant to create an error button, as shown [here in Figma](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=6202-193901&mode=design&t=S0nD9IF2Tt6UkCZS-4). The color prop will likely not have an effect otherwise, as all of the other variants have explicitly-defined background & text colors.

While reviewing the various screens, I noticed a discrepancy between the implementation of the toolbar for the Participant Details' Activity screen, and its appearance in Figma: In Figma, this screen has a distinct set of buttons, while in the app currently, it shares the toolbar with the Applet-level Activity screen. I took the liberty of creating a new component for this screen with appropriate buttons. None of them are wired up to any functionality, and the toolbar in its entirety will likely end up disabled for the initial release, but I figured it would be better for it to be there for when we needed it 🤷‍♀️.

#### Update Account Dialog

| Before | After |
|-|-|
| upgrade-dialog-before | upgrade-dialog-after |

#### Participant Details' Activities Tab

| Before | After |
|-|-|
| participant-activities-before | participant-activities-after |

#### Add Subject Dialog

| Before | After |
|-|-|
| create-account-before | create-account-after |

#### Applet Participants Tab

| Before | After |
|-|-|
| applet-participants-before | applet-participants-after |

#### Applet Activities Tab

| Before | After |
|-|-|
| applet-activities-before | applet-activities-after |

### 🪤 Peer Testing

1. Navigate to the relevant screens, observe the updated button variants.
2. In general, observe that no buttons have changed erroneously.

[M2-6071]: https://mindlogger.atlassian.net/browse/M2-6071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
+++ b/src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
@@ -68,9 +68,7 @@ export const ActivitiesToolbar = ({
                   alert('TODO: Assign activity');
                 }}
                 sx={{ minWidth: theme.spacing(10) }}
-                // TODO: Replace with missing `tonal` button variant as shown in Figma
-                // https://mindlogger.atlassian.net/browse/M2-6071
-                variant="outlined"
+                variant="tonal"
               >
                 {t('assign')}
               </Button>
--- a/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
@@ -214,10 +214,8 @@ export const AddParticipantPopup = ({
           buttonText={t(isFullAccount ? 'sendInvitation' : 'create')}
           disabledSubmit={!isValid}
           hasSecondBtn
-          // TODO: Update second button variant once 'tonal' variant added
-          // https://mindlogger.atlassian.net/browse/M2-6071
           secondBtnText={t('Reset')}
-          secondBtnVariant="outlined"
+          secondBtnVariant="tonal"
           onSecondBtnSubmit={resetForm}
           disabledSecondBtn={!isDirty}
           hasLeftBtn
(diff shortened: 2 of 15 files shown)


farmerpaul (approved):
Great work @JustinNusca! Thanks for noticing the toolbar inconsistencies and taking care of them. Happy to approve.

I did notice one nitpicky spacing inconsistency, which might not actually be related to your PR, but seems like a quick win. The space between the Activities toolbar and the first Flows heading seems to be a little too high, compared to Figma:

Figma | Code
-- | --
Screenshot 2024-05-09 at 7 16 05 PM | Screenshot 2024-05-09 at 7 24 10 PM

Oh and I guess the primary CTA label should be "Assign Activity" not just "Assign" (likely a relic from an earlier design).

PR #1711, opened 2024-05-09
fix(mi): Hide unfinished UI (M2-6458)

JustinNusca:
### 📝 Description

🔗 [M2-6458](https://mindlogger.atlassian.net/browse/M2-6458): [Participants] Hide Placeholder Elements Not Intended for R1

This PR hides some unfinished UI not intended for the immediate upcoming release. This includes:

- Removing the "Filter & Sort" buttons on the Applet-level Participants screen. These weren't connected to anything and are just simple `Button` components, so I opted to just delete the relevant lines for now.
- Removing the "Add CSV" button on the "Add Participant" dialog on the Applet-level Participants screen. Again, this wasn't connected to anything and was being configured by a couple props passed to the Modal component, so I opted to just remove those prop values for now.
- Hid the "Assign Activity" menu item on the dropdown menu visible when clicking the Ellipsis button on table rows in the Applet-level Participants screen. I made the rendering of this option conditional on the value of the `enableAssignActivity` feature flag, which is enabled in development environments.

This ticket also additionally specified the desire to truncate excessively long IDs on the Participants table. I added support for a `maxWidth` property on cell data, in addition to the existing `width` property, and added appropriate values to set this accordingly. The cell should be wide enough to fit a generic UUID, and on screens that are otherwise too small to accommodate longer cell widths, the column will being to truncate longer ID values.

Finally, this addresses a small issue with vertical padding between elements on the Activities page, originally pointed out [here, by farmerpaul](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1707#pullrequestreview-2048561941).

#### Participants Table

| Before | After |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_activities (2) | participants-after |

#### Participant Actions Dropdown Menu

| Before | After |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_activities (4) | action-menu-after |

#### Add Participant Dialog

| Before | After |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_activities (3) | add-menu-after |

#### Activities Padding

| Before | After |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_activities (1) | localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_activities |

### 🪤 Peer Testing  

1. Navigate to the **Applet > Participants** page and observe that Sort & Filter buttons are absent.
2. Observe that the ID column truncates sufficiently lengthy content.
3. Press to open a Participant's dropdown menu, and observe that the "Add Activity" option is 
absent (depending on the feature flag state).
4. Press the "Add Participant" button, and observe that the "Add via CSV" button is absent.
5. Navigate to the **Applet > Activities** page and observe that the padding between the toolbar and the Flow/Activities grid is correct.

[M2-6458]: https://mindlogger.atlassian.net/browse/M2-6458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
@@ -179,9 +179,6 @@ export const AddParticipantPopup = ({
           title={t('addParticipant')}
           buttonText={t('next')}
           hasLeftBtn
-          leftBtnText={t('addViaCSV')}
-          leftBtnVariant="outlined"
-          onLeftBtnSubmit={handleAddViaCSV}
           data-testid={dataTestid}
         >
           <StyledModalWrapper>
--- a/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
@@ -62,10 +62,6 @@ export const AddParticipantPopup = ({
     onClose?.(shouldRefetch);
   };
 
-  const handleAddViaCSV = () => {
-    alert('TODO: Add via CSV');
-  };
-
   const resetForm = () => reset({ ...defaults, accountType });
 
   const {
@@ -178,7 +174,6 @@ export const AddParticipantPopup = ({
           onSubmit={handleNext}
           title={t('addParticipant')}
           buttonText={t('next')}
-          hasLeftBtn
           data-testid={dataTestid}
         >
           <StyledModalWrapper>
--- a/src/modules/Dashboard/features/Participants/Participants.const.tsx
+++ b/src/modules/Dashboard/features/Participants/Participants.const.tsx
@@ -1,6 +1,7 @@
 export enum ParticipantsColumnsWidth {
   Pin = '8rem',
   Default = '28rem',
+  Id = '34.5rem',
   Schedule = '15rem',
   AccountType = '16rem',
   Menu = '7.2rem',
--- a/src/modules/Dashboard/features/Participants/Participants.const.tsx
+++ b/src/modules/Dashboard/features/Participants/Participants.const.tsx
@@ -1,7 +1,7 @@
 export enum ParticipantsColumnsWidth {
   Pin = '8rem',
   Default = '28rem',
-  Id = '34.5rem',
+  Id = '20.8rem',
   Schedule = '15rem',
   AccountType = '16rem',
   Menu = '7.2rem',
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
@@ -72,7 +72,10 @@ export const DashboardTable = ({
                     align={row[key].align}
                     width={row[key].width}
                     hasColFixedWidth={hasColFixedWidth}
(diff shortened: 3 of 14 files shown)


src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
```diff
@@ -179,9 +179,6 @@ export const AddParticipantPopup = ({
           title={t('addParticipant')}
           buttonText={t('next')}
           hasLeftBtn
-          leftBtnText={t('addViaCSV')}
-          leftBtnVariant="outlined"
-          onLeftBtnSubmit={handleAddViaCSV}
```
farmerpaul, line 184: Could you also delete the definition of `handleAddViaCSV` earlier in this file? Its presence has introduced a new eslint warning after this change.

src/modules/Dashboard/features/Participants/Participants.const.tsx
```diff
@@ -1,6 +1,7 @@
 export enum ParticipantsColumnsWidth {
   Pin = '8rem',
   Default = '28rem',
+  Id = '34.5rem',
```
farmerpaul: Just curious where you encountered this width measurement. When I check [Figma](https://www.figma.com/file/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?type=design&node-id=1628-123598&mode=design&t=8UUqsPA2g56Tfd5w-4), it shows this cell as being `20.8rem`:

Screenshot 2024-05-14 at 2 48 29 PM

However maybe I'm misinterpreting the task requirements; the Figma designs alone don't provide any clarity on how columns are expected to behave at varying viewport widths.

src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
```diff
@@ -179,9 +179,6 @@ export const AddParticipantPopup = ({
           title={t('addParticipant')}
           buttonText={t('next')}
           hasLeftBtn
```
farmerpaul: This should also be removed, else the button is still focusable and visible on hover:

```suggestion
```

farmerpaul (approved):
Just a couple questions/comments, but glad to preapprove.

src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
JustinNusca: Good catch, we have so many warnings that its difficult to spot when they've changed 😅.

src/modules/Dashboard/features/Participants/Participants.const.tsx
JustinNusca: I went ahead and updated this to match the 20.8 measurement. 

I agree that the Figma designs don't really provide clarity — nor would I expect an exhaustive specification for how HTML tables should behave in this application. I interpreted "excessively long" to mean "longer than anything the app itself will create by default", so I picked 34.5 to accommodate the approximate length of an entire UUID.

Note that also, even with this max-width value, all of the table cells will stretch to fill space on larger viewports.

PR #1713, opened 2024-05-09
fix(mi): Hide stats on `ActivitySummaryCard`s (M2-6507)

JustinNusca:
### 📝 Description

🔗 [M2-6507](https://mindlogger.atlassian.net/browse/M2-6507): [Activities] Hide Activity Card elements Not Intended for R1

This PR updates the `ActivitySummaryCard` to hide the stats section of the card. This is handled via a new prop, `showStats`, which defaults to `false`, and must be opted into rendering by passing the appropriate value, which we are not doing anywhere at the moment.

The linked task contains a list of suggestions around adding discrete prop values for controlling the visibility of individual stats. Given the rationale of "we don't know what stats we want to show yet", I opted to defer on this change (in favour of this single flag) for a few reasons:

* I think supporting this degree of customization would be better handled by either passing the stats as an array that maps to the output UI, or even by just accepting generic children for that whole area of the UI. This would allow us to support conditional rendering, or rendering arbitrary content by just passing the specific content we want to render, and omitting that which we don't.
* Given the reasoning — that we don't know which stats we're rendering — it seems like the design itself is subject to change. Given how variable the layouts could be for different amounts of content beyond what's already planned for, I don't want to try and accommodate such UI scenarios until we know we need to.

### 📸 Screenshots

| Before | After |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_participants_65e2fef6-0e69-4575-ae4c-62e9bc73822b (1) | localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_participants_65e2fef6-0e69-4575-ae4c-62e9bc73822b |

### 🪤 Peer Testing

1. Navigate to the **Applet > Activities** page and observe the Activity Summary Cards.
2. Navigate to the **Applet > Participant Details > Activities** page and observe the Activity Summary Cards.

[M2-6507]: https://mindlogger.atlassian.net/browse/M2-6507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/ActivitySummaryCard/ActivitySummaryCard.styles.ts
+++ b/src/modules/Dashboard/components/ActivitySummaryCard/ActivitySummaryCard.styles.ts
@@ -8,14 +8,17 @@ import {
   variables,
 } from 'shared/styles';
 
-export const StyledContainer = styled(StyledFlexSpaceBetween)`
-  flex-direction: column;
-  padding: ${theme.spacing(2.4)};
-  width: 30.6rem;
-  height: 44rem;
-  border: ${variables.borderWidth.md} solid ${variables.palette.surface_variant};
-  border-radius: ${variables.borderRadius.lg2};
-`;
+export const StyledContainer = styled(StyledFlexSpaceBetween)(
+  ({ showStats = false }: { showStats?: boolean }) => ({
+    flexDirection: 'column',
+    gap: theme.spacing(3.2),
+    padding: theme.spacing(2.4),
+    width: '30.6rem',
+    height: showStats ? '44rem' : undefined,
+    border: `${variables.borderWidth.md} solid ${variables.palette.surface_variant}`,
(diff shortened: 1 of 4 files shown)


mbanting (approved):
Looks good. Tested as well. Ideally I would have preferred to have a conversation first if deviating from the the ticket (AC or engineering approach). While those are valid points and it's possible design can evolve, the question really was more about what metrics to show rather than how (ie. visually) it should be displayed. It was also an opportunity to use the extra capacity we had this sprint to make a small, calculated investment in enriching this component that could lessen the effort in a future sprint. Let's move forward with this for now.

farmerpaul (approved):

PR #1706, opened 2024-05-07
feat(mi): create/view participant tags (M2-5861)

farmerpaul:
- [x] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-5861](https://mindlogger.atlassian.net/browse/M2-5861)

Adds create/view support for Participant tags to the admin panel. Includes:
- Tag dropdown in Add Participant popup
  - updated `SelectController` to add more universal support for existing `withChecked` prop
- Tags shown in the Participants table & Upgrade Account popup
  - created `ParticipantTagChip` component to be reused everywhere a tag is rendered
- Tags shown in the Take Now participant dropdown
- Tags shown in the Add Team Member popup (when role = "Reviewer")
  - refactored `TagsAutocompleteController` to be value-type-agnostic so it can render `ParticipantSnippet`s in the dropdown

### 📸 Screenshots

| Before                      | After                                 |
| -------------------------------------- | ------------------------------------- |
| Screenshot 2024-05-07 at 4 48 24 PM | Screenshot 2024-05-07 at 4 42 31 PM |
| Screenshot 2024-05-07 at 4 48 54 PM | Screenshot 2024-05-07 at 4 42 55 PM |
| Screenshot 2024-05-07 at 4 49 17 PM | Screenshot 2024-05-07 at 4 43 22 PM |
| Screenshot 2024-05-07 at 4 49 35 PM | Screenshot 2024-05-07 at 4 43 47 PM |
| Screenshot 2024-05-07 at 4 50 25 PM | Screenshot 2024-05-07 at 4 45 16 PM |

### 🪤 Peer Testing

1. Navigate to Applet > Participants and click Add Participant.
2. Choose either Limited Account and click Next.
    **Expected outcome:** Tag selector should be available, with "No selection" selected by default.
3. Pick a tag and proceed to create the limited account.
    **Expected outcome:** New limited account should be listed with selected tag in Participants list.
4. Choose "Upgrade Account" from this account's action menu.
    **Expected outcome:** Participant's Tag should be rendered in the header of the Upgrade Account popup.
5. Close the Upgrade Account popup.
6. Navigate to Activities tab and choose Take Now for an activity.
7. Expand the "Who is this activity about?" dropdown.
[...]

PR #1714, opened 2024-05-10
feature(mi): Add ability to edit subject tags (M2-6331)

JustinNusca:
### 📝 Description

🔗 [M2-6331](https://mindlogger.atlassian.net/browse/M2-6331): [Participant Overview] Edit Participant - Add Tag Field

> [!NOTE]  
> This PR includes a commit that updates related types authored by @farmerpaul in [M2-6481](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1719). I recommend reviewing by viewing only the last two commits, the change set for which can be seen [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1714/files/50f7fbff2109f0da40e5e7098dec10407bd95744..8476e8bde18ba6aee6aa10c44396b78d92f69afa).

This PR updates the "EditRespondantPopup" dialog, in order to support the updating of a given subject's tag value. It adds a similar dropdown component to the one added in the "AddParticipantPopup" component, in https://github.com/ChildMindInstitute/mindlogger-admin/pull/1706. The prop interface has also been tweaked to allow for pre-populating the field with a subject's pre-existing tag value.

If the given subject has a tag of "Team", then the field is disabled, with the "Team" value represented within it.

### 📸 Screenshots

| Non-team member | Team member |
|-|-|
| localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_participants (2) | localhost_3000_dashboard_fdb0c9fd-0e62-40b6-94a8-7fc6a55f4da8_participants (3) |

### 🪤 Peer Testing

1. Navigate to the **Applet > Participants** screen.
2. For a *non-team member* participant...
    1. Proceed to the "Edit Participant" dialog via the "More actions" menu.
    2. Observe the presence of a new Tag selection field. Select a new Tag value and submit.
    4. Observe that the tag has been updated for the given subject.
3. For *team member* participant...
    1. Proceed to the "Edit Participant" dialog via the "More actions" menu.
    2. Observe the presence of a new Tag selection field. It should be disabled, and the tag value cannot be changed.

[M2-6331]: https://mindlogger.atlassian.net/browse/M2-6331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[M2-6481]: https://mindlogger.atlassian.net/browse/M2-6481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
+++ b/src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
@@ -67,7 +67,7 @@ export const EditRespondentPopup = ({
       values: {
         secretUserId: secretUserId.trim(),
         nickname: nickname?.trim(),
-        ...(tag ? { tag } : {}),
+        ...(tag && { tag }),
       },
       subjectId,
     });
@@ -84,15 +84,16 @@ export const EditRespondentPopup = ({
   return (
     <Modal
       open={popupVisible}
-      onClose={onCloseHandler}
+      onClose={() => onCloseHandler(false)}
       onSubmit={handleSubmit(submitForm)}
       disabledSubmit={isLoading}
       title={t('editParticipant')}
       buttonText={t('save')}
-      hasSecondBtn
-      onSecondBtnSubmit={onCloseHandler}
-      secondBtnText={t('cancel')}
+      hasLeftBtn
+      onLeftBtnSubmit={() => onCloseHandler(false)}
+      leftBtnText={t('cancel')}
       data-testid={dataTestid}
+      width="56"
     >
       <>
         {isLoading && <Spinner uiType={SpinnerUiType.Secondary} noBackground />}
--- a/src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
+++ b/src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
@@ -1,11 +1,13 @@
-import { useEffect, useState, ChangeEvent } from 'react';
+import { useState, ChangeEvent } from 'react';
+import { Box } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { useForm } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
 
-import { Modal, Spinner, SpinnerUiType } from 'shared/components';
+import { Modal, Spinner, SpinnerUiType, Svg } from 'shared/components';
+import { PARTICIPANT_TAG_ICONS, USER_SELECTABLE_PARTICIPANT_TAGS } from 'shared/consts';
 import { StyledErrorText, StyledModalWrapper } from 'shared/styles';
-import { InputController } from 'shared/components/FormComponents';
+import { InputController, SelectController } from 'shared/components/FormComponents';
 import { useAsync } from 'shared/hooks/useAsync';
 import { editSubjectApi } from 'api';
 import { falseReturnFunc, getErrorMessage } from 'shared/utils';
@@ -14,7 +16,6 @@ import { banners } from 'redux/modules';
 
 import { EditRespondentForm, EditRespondentPopupProps } from './EditRespondentPopup.types';
 import { editRespondentFormSchema } from './EditRespondentPopup.schema';
-import { StyledController } from './EditRespondentsPopup.styles';
 
 export const EditRespondentPopup = ({
   popupVisible,
@@ -23,6 +24,7 @@ export const EditRespondentPopup = ({
 }: EditRespondentPopupProps) => {
   const { t } = useTranslation('app');
   const dispatch = useAppDispatch();
+  const isTeamMember = chosenAppletData?.subjectTag === 'Team';
 
   const [isServerErrorVisible, setIsServerErrorVisible] = useState(true);
 
@@ -30,7 +32,11 @@ export const EditRespondentPopup = ({
 
   const { handleSubmit, control, setValue, getValues, trigger } = useForm<EditRespondentForm>({
     resolver: yupResolver(editRespondentFormSchema()),
-    defaultValues: { secretUserId: '', nickname: '' },
+    defaultValues: {
+      secretUserId: chosenAppletData?.respondentSecretId ?? '',
+      nickname: chosenAppletData?.respondentNickname ?? '',
+      tag: chosenAppletData?.subjectTag ?? ('' as const),
+    },
   });
 
   const dataTestid = 'dashboard-respondents-edit-popup';
@@ -54,13 +60,14 @@ export const EditRespondentPopup = ({
   const submitForm = () => {
     if (!chosenAppletData) return;
 
-    const { secretUserId, nickname } = getValues();
+    const { secretUserId, nickname, tag } = getValues();
     const { subjectId } = chosenAppletData;
 
     editRespondent({
       values: {
         secretUserId: secretUserId.trim(),
         nickname: nickname?.trim(),
+        ...(tag ? { tag } : {}),
       },
       subjectId,
     });
@@ -72,12 +79,6 @@ export const EditRespondentPopup = ({
     trigger('secretUserId');
   };
 
-  useEffect(() => {
-    const { respondentNickname, respondentSecretId = '' } = chosenAppletData || {};
-    setValue('secretUserId', respondentSecretId);
-    setValue('nickname', respondentNickname || '');
-  }, [chosenAppletData]);
-
   const hasServerError = error && isServerErrorVisible;
 
   return (
@@ -86,7 +87,7 @@ export const EditRespondentPopup = ({
       onClose={onCloseHandler}
       onSubmit={handleSubmit(submitForm)}
       disabledSubmit={isLoading}
-      title={t('editRespondent')}
+      title={t('editParticipant')}
       buttonText={t('save')}
       hasSecondBtn
       onSecondBtnSubmit={onCloseHandler}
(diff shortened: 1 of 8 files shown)


src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -86,7 +87,7 @@ export const EditRespondentPopup = ({
       onClose={onCloseHandler}
       onSubmit={handleSubmit(submitForm)}
       disabledSubmit={isLoading}
-      title={t('editRespondent')}
+      title={t('editParticipant')}
```
farmerpaul, line 90: I noticed that the JIRA ticket says there isn't a design in Figma for the Edit Participant screen, but I didn't have trouble [finding it](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1714/files) myself:

Edit Participant

It seems there's a couple discrepancies here:
- it should be bit narrower than the default Modal (`width="56"` same as in `UpgradeAccountPopup`)
- the Cancel button should be specified using the `leftBtn*` props instead of `secondBtn*`

Also might not hurt to just edit the JIRA ticket to include the correct screenshot and remove that last bullet in the "Notes for testing" section.

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -96,29 +97,59 @@ export const EditRespondentPopup = ({
...
-              />
-            </StyledController>
-          </form>
-          {hasServerError && <StyledErrorText>{getErrorMessage(error)}</StyledErrorText>}
+          <Box
+            component="form"
+            sx={{ display: 'flex', flexDirection: 'column', gap: 2.4 }}
+            onSubmit={handleSubmit(submitForm)}
+            noValidate
+          >
```
farmerpaul, line 106: Nit 😇 
```suggestion
          <StyledFlexColumn
            component="form"
            sx={{ gap: 2.4 }}
            onSubmit={handleSubmit(submitForm)}
            noValidate
          >
```

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.types.ts
```diff
@@ -3,6 +3,7 @@ import { ChosenAppletData } from '../../Respondents.types';
 export type EditRespondentForm = {
   secretUserId: string;
   nickname?: string;
+  tag?: string;
```
farmerpaul:
```suggestion
  tag?: ParticipantTag;
```
(or `UserSelectableParticipantTag`, with my changes?)

farmerpaul (commented):
I know it's not quite ready to review, just had a couple early comments! Can do a proper review once you've had a chance to do that cherry-picking.

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.types.ts
JustinNusca: Yep — I was waiting for your changes so I could incorporate them into this PR. The branch is now based of your commit that updates the types, and I've changed them accordingly, here.

Note that I did stick with `ParticipantTag`, since this form may be rendered with the `Team` value in a disabled state.

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
JustinNusca: I'd prefer to leave this as-is. The wrapping of a StyledComponent around a MUI component breaks the polymorphic type validation, which is being used to validate the types of HTML attributes passed to the rendered `<form>` element here (ie, the `noValidate` attr). 

While I haven't dug into it too much, my hunch is that the error is related to conflicting functionality in MUI and Emotion, which MUI is using for styling. Emotion supports an `as` prop (which still works with MUI components, despite type errors), whereas MUI supports the same feature via a `component` prop.

Applying the changes you suggest would also require manually re-asserting `StyledFlexColumn` as a different component type, which would require an intermediary variable, like...

```ts
const StyledFlexColumnForm = StyledFlexColumn as StyledComponent<BoxProps<'form'>>;
```

farmerpaul: Wow, good investigation there Justin. This choice to use `component` vs `as` sure makes things needlessly awkward indeed, so yeah, totally understandable not to use the utility.

Honestly, I'm not even a fan of the pattern we have in place here for bundling style utilities. I'd be more in favour of something simpler that facilitates composition, which also isn't possible in the current structure. Something like this, for example:

```ts
// something like typography.const.ts

export const bodyLarge: SxProps = {
  fontSize: variables.font.size.lg,
  lineHeight: variables.font.lineHeight.lg,
  fontWeight: variables.font.weight.regular,
};

export const ellipsis: SxProps = {
  textOverflow: 'ellipsis',
  overflow: 'hidden',
  whiteSpace: 'nowrap',
};
```

```tsx
// component file

<Box sx={{ ...bodyLarge, ...ellipsis }}>
  // …
</Box>
```

This also gives us the benefit of TypeScript intellisense, which we don't get using CSS string literals.

… but that would be big refactor, so not up for discussion obviously. 😅

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -23,14 +24,19 @@ export const EditRespondentPopup = ({
...
 
   const onCloseHandler = (shouldRefetch = false) => onClose(shouldRefetch);
 
   const { handleSubmit, control, setValue, getValues, trigger } = useForm<EditRespondentForm>({
     resolver: yupResolver(editRespondentFormSchema()),
-    defaultValues: { secretUserId: '', nickname: '' },
+    defaultValues: {
+      secretUserId: chosenAppletData?.respondentSecretId ?? '',
+      nickname: chosenAppletData?.respondentNickname ?? '',
+      tag: chosenAppletData?.subjectTag ?? ('' as const),
```
farmerpaul, line 38: Oh neat… of course you can do `as const` in general in these types of cases, instead of `as ParticipantTag`. Nice tip. :+1:

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -54,13 +60,14 @@ export const EditRespondentPopup = ({
...
 
-    const { secretUserId, nickname } = getValues();
+    const { secretUserId, nickname, tag } = getValues();
     const { subjectId } = chosenAppletData;
 
     editRespondent({
       values: {
         secretUserId: secretUserId.trim(),
         nickname: nickname?.trim(),
+        ...(tag ? { tag } : {}),
```
farmerpaul: Does this work, or is the ternary format a better practice?
```suggestion
        ...(tag && { tag }),
```

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
```diff
@@ -86,39 +87,70 @@ export const EditRespondentPopup = ({
       onClose={onCloseHandler}
```
farmerpaul: The table shouldn't refresh after cancelling the edit action by hitting close (I noticed this omission in a couple other popups I worked on):
```suggestion
      onClose={() => onCloseHandler(false)}
```

farmerpaul (approved):
Just a couple comments but otherwise looks good!

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
JustinNusca: The statements are approximately equivalent in effect. False-y values act like an empty object if you apply the object spread operator to them, so the ternary isn't really necessary, unless you favour the explicitness in how its read (by which I just mean that there's a literal readable instance of an empty object). 

I don't have any particular opinion — I've updated it to your suggestion for conciseness.

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
JustinNusca: Yeah — I didn't think it was significant enough to call out when reviewing your other PR, because either approach is totally valid, but not requiring import statements this way is one reason I like unions like these over enums where possible 👍.

src/modules/Dashboard/features/Respondents/Popups/EditRespondentPopup/EditRespondentPopup.tsx
JustinNusca: I totally agree with you. 

I haven't checked this very thoroughly, but I'm a bit worried about creating deeply-nested JSX trees as we create components with `styled` to wrap other components created with `styled` and so-on. Objects passed to `sx` would have to be parsed on mounting, but at least by comparison it's just the single instance of that occurring.

And yeah, so long as we're doing CSS in JS, I _much_ prefer the advantages object notation provides regarding type-checking, intellisense, and out-of-the-box syntax highlighting that string literals don't.

PR #1726, opened 2024-05-16
fix(mi): Activities tab button alignment (M2-6620)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-6620](https://mindlogger.atlassian.net/browse/M2-6620)

Makes sure Activities header right button alignment remains right-aligned when `enableActivityFilterSort` feature flag is turned off.

### 📸 Screenshots

| Screenshot 2024-05-16 at 2 22 22 PM | Screenshot 2024-05-16 at 2 21 52 PM |

### 🪤 Peer Testing

- Add the line `features.enableActivityFilterSort = false;` to just before the `return` statement of `useFeatureFlags.ts`
- Navigate to Applet > Activities
    **Expected outcome:** Header buttons should be correctly right-aligned.

### DIFF
--- a/src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
+++ b/src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
@@ -21,7 +21,7 @@ export const ActivitiesToolbar = ({
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
 
   return (
-    <StyledFlexWrap sx={{ gap: 1.2, placeContent: 'space-between', ...sx }} {...otherProps}>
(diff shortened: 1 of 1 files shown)


JustinNusca (approved):
👍 Does what it says.

It's not captured in the ticket, and I'm not blocking merging over this or anything, but one thing I'm wondering about:

This issue is also present on the Participant Details > Activities screen. In that case, however, _none_ of the buttons on that toolbar are functional, currently. 

I know I added it myself to match the Figma designs for this screen, but I'm wondering if we shouldn't just omit rendering it entirely for the time being.

**EDIT**: Actually, it looks like what I'm describing is already captured in a [separate ticket](https://mindlogger.atlassian.net/browse/M2-6487), so don't even worry about it here!

**EDIT 2**: I nabbed the ticket I mentioned, it's open in a PR [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1729).

PR #1727, opened 2024-05-16
feat: Re-add Public links (M2-6498)

JustinNusca:

### 📝 Description

🔗 [M2-6498](https://mindlogger.atlassian.net/browse/M2-6498): [FE][Add Participants] Public Link - Add Public Link to App

This PR adds the ability to generate public links to allow users to particpate in an applet, either via a shared guest account or by accepting the invite link with the account of their choosing. More specific details about the expected functionality can be gleaned from the testing steps.

To implement this, I've added two components: 
- `PublicLinkToggle`, for showing the current status of an applet's public link, and providing options for creation/deletion accordingly.
- `PublicLinkPopup`, for handling the confirmation dialogs and actually confirming/deleting public links.

The modal footers for this design were particularly onerous to map to the existing props for footer buttons, either in layout, button variant, button color, or need for conditional rendering. Instead of creating multiple new props for each discrete bit of configuration, which is the existing pattern, I opted to just create a `footer` component that accepts arbitrary rendered content.

### 📸 Screenshots

> [!NOTE] 
> The gifs below only demonstrate the changes within the admin app itself, but the generated links should be functional and should allow appropriate access to the applet within the web-app accordingly.

| "Limited Account" public link | "Full Account" public link |
|-|-|
|  |  |

### 🪤 Peer Testing

> [!WARNING]  
> There is a known issue with public link generation that will result in error responses from the backend that prevent the creation of "Limited Account" Public links in specific situations.
>
> This will be handled separately in [M2-6623](https://mindlogger.atlassian.net/browse/M2-6623).
> 
> To avoid this issue, after deleting a "Limited Account" public link, make sure you remove the "Guest Account Submission" subject from an applet's participants _prior_ to attempting to create a _new_ "Limited Account" public link.

#### For "Full Accounts"

From the **Applet Dashboard > Participants** screen:
1. Press the "Add Participant" button. Observe the public link indicator, and the option to "Create Public Link".
2. Press "Create Public Link", and in the resulting dialog, press "Yes, an account is required".
3. Observe the state of the public link indicator has changed to show a public link, and the option to "Delete Public Link".
4. Copy the public link, either via the button or selecting and copying the text directly.
5. Switch to the **Web app**. Authenticate with an account that is not already a participant in the given applet, and navigate to the link you copied.
6. You should be able to accept an invite to the given applet, and subsequently able to interact with the applet's activities.
7. Switch back to the **Admin app**. On refreshing the page, you should see the user is now present in the Participants table.
8. Press the "Add Participant" button, and then "Delete Public Link". Confirm your choice in the following dialog, and observe that the public link has been removed.
9. Attempting to visit the now-deleted link in the web app will result in an "invalid or expired invitation" message.

#### For "Limited Accounts"

From the **Applet Dashboard > Participants** screen:
1. Press the "Add Participant" button. Observe the public link indicator, and the option to "Create Public Link".
2. Press the "Create Public Link" button, and in the resulting dialog, press "No, an account is not required".
3. Observe the state of the public link indicator has changed to show a public link, and the option to "Delete Public Link".
4. Close the dialog, and observe the Participants table is refreshed to include a "Guest Account Submission" participant.
5. Re-open the dialog and copy the public link, either via the button or selecting and copying the text directly.
6. Switch to the **Web app**. While unauthenticated, navigate to the link you copied. You should be able to view the applet and interact with the applet's activities.
7. Switch back to the **Admin app**. Press the "Add Participant" button, and then "Delete Public Link". Confirm your choice in the following dialog, and observe that the public link has been removed.
8. Attempting to visit the now-deleted link in the web app will result in an "invalid or expired invitation" message.

[M2-6498]: https://mindlogger.atlassian.net/browse/M2-6498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-6623]: https://mindlogger.atlassian.net/browse/M2-6623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/PublicLinkPopup/PublicLinkPopup.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </>
+          )}
+        </StyledFlexAllCenter>
+      }
+      hasActions={false}
+      onClose={() => onClose?.()}
+      title={hasPublicLink ? t('deletePublicLink') : t('publicLinkCreate')}
+      {...otherProps}
+    >
+      <StyledFlexColumn sx={{ gap: 1.6, px: 3.2 }}>
```
farmerpaul: We have a shared `<StyledModalWrapper>` component that we use to apply standard standard horizontal & vertical padding to the main content of modals. (Ideally, I think such standard paddings should be applied in the `<Modal>` definition itself, however I think there are some exceptions that would require some investigation before applying universally.) For now, you could just wrap your `<StyledFlexColumn>` with `<StyledModalWrapper>` (or apply needed flex styling to it as is done in `AddParticipantPopup`) and remove the `px: 3.2`. Doing so will also make the vertical spacing match Figma.

src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/PublicLinkPopup/PublicLinkPopup.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <Modal
+      data-testid={dataTestId}
+      footer={
+        <StyledFlexAllCenter sx={{ gap: 1.6, width: '100%' }}>
+          {hasPublicLink ? (
+            <Button
+              color="error"
+              data-testId={dataTestId && `${dataTestId}-delete-bttn`}
```
farmerpaul: Nit: I've only ever seen buttons abbreviated as `btn` in code (despite their two "t"s). Seems `btn` is also the de-facto abbreviation used in the codebase. NBD, just a stickler for consistency. 😛

src/shared/components/Modal/Modal.tsx
```diff
@@ -108,6 +109,11 @@ export const Modal = ({
           )}
         </StyledDialogTitle>
         {children}
+        {footer && (
+          <StyledFlexSpaceBetween sx={{ gap: 1.6, px: 3.2, pt: 3.6, pb: 2.4 }}>
```
farmerpaul: Does it make sense to also apply `footerStyles` here, given its name? Or perhaps rename that prop to `actionsStyles`? Apparently it's no longer even used in the codebase, probably a relic after some restyling was done. Maybe we could delete it?

farmerpaul (approved):
Looks great, works as expected. I mentioned a few nitpicky things, but good to go.

src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/PublicLinkPopup/PublicLinkPopup.tsx
JustinNusca: Hah — every instance of `bttn` was added by me. Aside from the changes in this PR, there's some variables in a test file that I named with the two-t style a few weeks ago. I went ahead and [renamed all of them](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1727/commits/6b7148f41bd4370dbce091f3bb29fc28171c4ae2).

src/shared/components/Modal/Modal.tsx
JustinNusca: It looks like there's actually one instance of `footerStyles` being used, here in `RecordAudio.tsx`:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/fcc80a7eaa39605c4452f60a51d65a632d3f8ed0/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/AudioPlayer/RecordAudio/RecordAudio.tsx#L76-L78

I kept going back-and-forth on this — My original thinking was that this wrapper element around the `footer` content should just be a sensible default that allowed for the what seemed like the common button configurations design was moving towards (per their demo a few weeks back) — single buttons are centered, multiple buttons are arranged with the `space-between`. 

For any more nuanced cases and discrete groups (which I'm sure there's a lot of — even the popup I added here was a bit atypical, with multiple centered buttons), devs can just provide their own wrappers in `footer` directly to create whichever layouts they please.

That said, `footerStyles` can sure be handy, if all you need is a different `justifyContent` property.

I'm loathe to add even more props to this component, though. If we're opening up that wrapper to configuration, I'd be more inclined to just have `footerProps` or similar, and spread that to this element, instead of inevitably adding more props for configuring each individual attribute as the need arises.

But that also seems like overkill to me, especially since I wouldn't even really need this in my PR here.

In the end I opted to [just pass `footerStyles`](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1727/commits/6b7148f41bd4370dbce091f3bb29fc28171c4ae2#diff-4a37298ff0bc9fc644283461e6892de09b67a6ccc9c1bd286afe927582ac6d2cR115). 😅

farmerpaul: Works for me! Thanks for sharing your thought process.

FWIW I think all options for button flexibility could be handled by a single array of ButtonProps objects, which is what we did on my last project for a very similar situation, and it worked great. It was very flexible, allowing them to dictate variants, disabled state, click events, etc. Layout could be dictated by just providing left & right `auto` margins as needed, which allows you to centre them as well as lay them out in a space-between fashion. Some future refactor, perhaps. 🙂

PR #1728, opened 2024-05-16
feat(mi): Updated take now modal form (M2-6480)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-6480](https://mindlogger.atlassian.net/browse/M2-6480)

This PR updates the form on the TakeNowModal to match the new designs in Figma. The form now has a new checkbox, a new dropdown field, and the order of the fields have been rearranged. The tooltips have also been removed.

Submission of the form will now send these query parameters to the respondent web app:
- `startActivityOrFlow` - The ID of the activity
- `sourceSubjectId` - The subject ID of the user in the "**Who will be providing the responses?**" dropdown
- `targetSubjectId` - The subject ID of the user in the "**Who are the responses about?**" dropdown
- `subjectId` - The same as `targetSubjectId` for backwards compatibility with the respondent web app
- `respondentId` - The user ID of the user in the "**Who will be providing the responses?**" dropdown when the checkbox is checked, otherwise the user ID of the user in the "**Who will be inputting the responses?**" dropdown

#### Before

Screenshot 2024-05-16 at 3 22 21 PM

#### After

Screenshot 2024-05-16 at 3 25 27 PM

Screenshot 2024-05-16 at 3 26 18 PM

#### Pre-requisites

A workspace containing:
- An applet with:
  - An activity with at least one item
  - At least one full account participant
  - At least one limited account participant

#### Steps

- Launch the dashboard and open the applet
- Navigate to the activities tab
- Click the context menu on the activity and select the "Take Now" menu item
- Observe that the modal now matches the design in Figma
- Play around with the fields according to the scenarios below (as well as anything else you can come up with!)

| Scenario | Who will be providing the responses | This person will be inputting their own responses | Who will be inputting the responses | Who are the responses about | startActivityOrFlow | sourceSubjectId                     | targetSubjectId                     | subjectId (same as targetSubjectId) | respondentId          |
|----------|-------------------------------------|---------------------------------------------------|-------------------------------------|-----------------------------|---------------------|-------------------------------------|-------------------------------------|-------------------------------------|-----------------------|
| 1        | Owner                               | ✅                                                 | 🚫                                  | Limited account participant | Activity ID         | Owner subject ID                    | Limited account subject ID          | Limited account subject ID          | Owner user ID         |
| 2        | Owner                               | ✅                                                 | 🚫                                  | Full account participant    | Activity ID         | Owner subject ID                    | Full account participant subject ID | Full account participant subject ID | Owner user ID         |
| 3        | Owner                               | ✅                                                 | 🚫                                  | Owner                       | Activity ID         | Owner subject ID                    | Owner subject ID                    | Owner subject ID                    | Owner user ID         |
| 4        | Owner                               | ✅                                                 | 🚫                                  | Manager                     | Activity ID         | Owner subject ID                    | Manager subject ID                  | Manager subject ID                  | Owner user ID         |
| 5        | Full account participant            | ❌                                                 | Owner/Manager                       | Limited account participant | Activity ID         | Full account participant subject ID | Limited account subject ID          | Limited account subject ID          | Owner/Manager user ID |
| 6        | Full account participant            | ❌                                                 | Owner/Manager                       | Full account participant    | Activity ID         | Full account participant subject ID | Full account participant subject ID | Full account participant subject ID | Owner/Manager user ID |
| 7        | Full account participant            | ❌                                                 | Owner/Manager                       | Owner/Manager               | Activity ID         | Full account participant subject ID | Owner/Manager subject ID            | Owner/Manager subject ID            | Owner/Manager user ID |
| 8        | Limited account participant         | ❌                                                 | Owner/Manager                       | Limited account participant | Activity ID         | Limited account subject ID          | Limited account subject ID          | Limited account subject ID          | Owner/Manager user ID |
| 9        | Limited account participant         | ❌                                                 | Owner/Manager                       | Full account participant    | Activity ID         | Limited account subject ID          | Full account participant subject ID | Full account participant subject ID | Owner/Manager user ID |
| 10       | Limited account participant         | ❌                                                 | Owner/Manager                       | Owner/Manager               | Activity ID         | Limited account subject ID          | Owner/Manager subject ID            | Owner/Manager subject ID            | Owner/Manager user ID |

### ✏️ Notes

It may be useful to test this in conjunction with https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/459. If you opt to go this route, run this query on your local database after going through the take now flow:

```sql
SELECT RS.secret_user_id || coalesce(' (' || string_agg(DISTINCT UAA2.role, ', ') || ')', '')                respondent_secret_id,
       SS.secret_user_id ||
       coalesce(' (' || string_agg(DISTINCT UAA1.role, ', ') || ')', '')                                     source_subject_secret_id,
       TS.secret_user_id ||
       coalesce(' (' || string_agg(DISTINCT UAA3.role, ', ') || ')', '')                                     target_subject_secret_id,
FROM answers A
         join subjects TS on A.target_subject_id = TS.id
         join subjects SS on A.source_subject_id = SS.id
         join subjects RS on A.respondent_id = RS.user_id
         FULL OUTER JOIN user_applet_accesses UAA1 ON SS.user_id = UAA1.user_id
         FULL OUTER JOIN user_applet_accesses UAA2 ON RS.user_id = UAA2.user_id
         FULL OUTER JOIN user_applet_accesses UAA3 ON TS.user_id = UAA3.user_id
WHERE A.applet_id = (SELECT id
                     FROM applets
                     WHERE creator_id =
                           (SELECT id FROM users WHERE email = encode(sha224('someone5@example.com'::bytea), 'hex'))
                     LIMIT 1)
  AND A.created_at > '2024-05-21 00:00:00.000000'
GROUP BY A.respondent_id, A.source_subject_id, RS.secret_user_id, SS.secret_user_id, A.target_subject_id,
         TS.secret_user_id, A.created_at
ORDER BY A.created_at;
```

It should yield a table like this, which you can cross-reference with the table above:

| respondent\_secret\_id                                       | source\_subject\_secret\_id                                  | target\_subject\_secret\_id                                  |
|:-------------------------------------------------------------|:-------------------------------------------------------------|:-------------------------------------------------------------|
| a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | limitedAccount1                                              |
| a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | participant1 \(respondent\)                                  |
| a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   |
| a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | a608cfc9-708d-4537-aaff-91d460b080c2 \(owner, respondent\)   | 65c128f7-4a71-44e8-8ddd-efe820a6fa36 \(manager, respondent\) |

Otherwise, add a `console.log` of the URL around here:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/7d103ca99fad9eaf9f4249ccf1561720eed0e075/src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx#L212

Such as

```typescript
// window.open(url.toString(), '_blank');
for (const [name, value] of url.searchParams.entries()) {
  console.log(`${name}: ${value}`);
}
```

This will allow you to observe the results more readily:

```txt
startActivityOrFlow: ad0d41f2-f2fc-45b5-b0a0-1699df04be95
sourceSubjectId: 22c87606-933a-4803-8ad6-cffdde28bb7f
targetSubjectId: efe325cf-2cb7-4843-9fea-e24d82857295
respondentId: ccde1cfc-6c27-47bc-98df-67cbbeb8c5e4
subjectId: efe325cf-2cb7-4843-9fea-e24d82857295
```

#### Other notes

I made a number of updates to the modal to bring it more in line with Figma, but I still opted to maintain a text field as the displayed value of the dropdown. I have a WIP implementation of the `ParticipantSnippet` but it still needs some work to add the search field as part of the dropdown, so I opted to omit it from this PR

I slightly reworded the warning message below the _"Who will be providing the responses?"_ dropdown, so that it can apply to both limited and full accounts. This is because we don't have a design for this, and full account participants are currently also unable to input their own responses


src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -64,42 +71,101 @@ export const LabeledUserDropdown = ({
...
               ...props,
             }}
           />
         )}
-        getOptionLabel={(value) =>
-          `${value.secretId}${value.nickname ? ` (${value.nickname})` : ''}`
-        }
+        isOptionEqualToValue={(option, value) => option.id === value.id}
+        groupBy={(option: ParticipantDropdownOption) => {
+          if (!showGroups) return undefined;
```
ChaconC: You should move the `showGroups` validation to satisfy type for `groupBy?` by returning `undefined` instead of `() => undefined`

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -26,169 +21,350 @@ import { StyledImageContainer, StyledImg } from '../ActivitySummaryCard/Activity
...
+                  handleSearch={(query) => handleSearch(query, ['team', 'participant'])}
+                  canShowWarningMessage={true}
+                  showGroups={true}
+                />
+                <FormControlLabel
+                  control={<Checkbox sx={{ margin: 0, width: 48, height: 48 }} />}
+                  sx={{ gap: 0.4 }}
+                  checked={isSelfReporting}
+                  onChange={(_e, checked) => setIsSelfReporting(checked)}
+                  disabled={true}
```
ChaconC: I'm curious why this is `disabled` for all cases. Reading the AC it would seem like selecting a Team member should check the box but still allow to change the Respondent

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -109,18 +175,32 @@ export const LabeledUserDropdown = ({
...
         loading={isSearching}
         loadingText={t('loadingEllipsis')}
-        noOptionsText={t('takeNowNotFound')}
+        noOptionsText={t('takeNow.modal.dropdown.notFound')}
         onInputChange={(_e, search) => debouncedSearchHandler(search)}
         {...rest}
       />
+      <RenderIf condition={shouldShowWarningMessage}>
+        <StyledFlexTopCenter
+          sx={{
```
ChaconC: Missing `borderBottomLeftRadius` and `borderBottomRightRadius` when `shouldShowWarningMessage` is true

ChaconC (approved):
This looks really good!
I tested the whole set of cases and was able to validate the different `subjectId` as expected and was able to complete the TakeNow flow using the linked web app PR.

Some minor comments plus a question about the respondent selection but looks ok to approve

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Good call. Thanks!

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: That's true, according to the AC you should be able to do something like that. But I set it to always be disabled because I think it's not a valid use case. The only values available in the middle dropdown are team members.

I initially had it as `disabled={sourceSubject?.tag !== 'Team'}` which would allow this, but it felt like a weird use case. Do you think I should change it back?

image

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: I'm not sure I catch what you mean here. It looks like that's what's already happening?

ChaconC: There's a type error for `groupBy?`
```
Type '(option: ParticipantDropdownOption) => string | undefined' is not assignable to type '(option: ParticipantDropdownOption) => string'.
```

I'm suggesting changing this to 
```javascript
groupBy={
  !showGroups
    ? undefined
    : (option: ParticipantDropdownOption) => {
        if (option.tag !== 'Team') {
          return t('takeNow.modal.dropdown.participantGrouping');
        } else {
          return t('takeNow.modal.dropdown.teamMemberGrouping');
        }
      }
}
```

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
ChaconC: After talking to @mbanting I do believe this would be a valid use case, especially for R2.

Example scenario:

* A group of users use a **single Admin account** for the Admin app.
* A member of this workspace sets up Take Now to be completed by **Manager A** using the Admin account.
* On the web app, the user logs in as **Manager A** and completes the assessment.

If we force the checkbox to always be enabled they would have to use the Admin account to complete the assessment

src/resources/app-en.json
```diff
@@ -1215,14 +1215,25 @@
   "tag": "Tag",
   "takeMeToTheResponseData": "Take me to the response data",
   "takePhoto": "Take Photo",
-  "takeNow": "Take Now",
```
ChaconC, line 1218: This is causing other buttons to miss text 
CleanShot 2024-05-21 at 18 51 29

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Interestingly I don't get this type error locally (in-editor or tsc), but I see what you mean now 👍

src/resources/app-en.json
sultanofcardio: Assuming this is the button in `RespondentDataHeader`. Fixed it. Thanks

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
farmerpaul: I remember seeing that TS error as well; maybe in your case you weren't getting some tsc errors due to that `sx` issue breaking the TS server in VS Code?

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -64,42 +71,105 @@ export const LabeledUserDropdown = ({
...
-              height={24}
-              data-testid={`${rest['data-testid']}-tooltip-icon`}
-            />
-          </Box>
-        </Tooltip>
+        <StyledTitleMedium sx={{ fontWeight: 700, color: variables.palette.on_surface }}>
+          {label}
+        </StyledTitleMedium>
+        <RenderIf condition={!!tooltip}>
+          <Tooltip tooltipTitle={tooltip}>
```
farmerpaul: I think I get the intention of creating this `RenderIf` component – to avoid awkward `&&` and ternary conditional expressions in JSX? First that I've seen it, but I think I get the rationale.

I might take advantage of it sometimes for ease of reading/avoid ternaries, but just thought I'd mention it doesn't play as nicely with type-checking in situations with required props, for example:

```tsx
// some component definition with a required prop

export const MyComponent = (props: { title: string }) => { … }

// ❌ TypeScript complains with RenderIf: Type 'undefined' is not assignable to type 'string'.
const title: undefined | string;

<RenderIf condition={!!title}>
  <MyComponent title={title} />
</RenderIf>

// ✅ regular conditional works 
const title: undefined | string;

{!!title && (
  <MyComponent title={title} />
)}
```

I guess in the `RenderIf` version, `<MyComponent title={title} />` always gets evaluated (i.e., rendered), even if its render output doesn't get used in the final parent component rendering (when `title` is falsey). While in the second version, `<MyComponent title={title} />` it only gets rendered when `title` is truthy.  (At least I think so… it may make a difference in some cases.)

Curious if you find there are other benefits of using `RenderIf`?

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -26,169 +21,350 @@ import { StyledImageContainer, StyledImg } from '../ActivitySummaryCard/Activity
...
-    },
+  const [activity, setActivity] = useState<BaseActivity | null>(null);
+  const [allParticipants, setAllParticipants] = useState<ParticipantDropdownOption[]>([]);
+  const [allTeamMembers, setAllTeamMembers] = useState<Manager[]>([]);
+  const [defaultTargetSubject, setDefaultTargetSubject] =
+    useState<ParticipantDropdownOption | null>(null);
+  const [defaultSourceSubject, setDefaultSourceSubject] =
+    useState<ParticipantDropdownOption | null>(null);
+
+  const participantToOption = useMemo(
```
farmerpaul: Would it not make more use to do `useCallback` here, or there a benefit of `useMemo` in this context?

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -26,169 +21,350 @@ import { StyledImageContainer, StyledImg } from '../ActivitySummaryCard/Activity
...
+  );
+
+  useEffect(() => {
+    if (appletId) {
+      if (allParticipants.length === 0 && !isFetchingParticipants) {
+        fetchParticipants({
+          params: {
+            ownerId,
+            appletId,
+            limit: 100,
```
farmerpaul, line 133: Do we need to be concerned about applets with over 100 participants?

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -26,169 +21,350 @@ import { StyledImageContainer, StyledImg } from '../ActivitySummaryCard/Activity
...
+
+        // This conditional shouldn't be necessary, but TS is unable to propagate the type information from
+        // (isSelfReporting || loggedInUser), so we have to check them again (or use the forbidden non-null assertion)
+        if (isSelfReporting && sourceSubject.userId) {
+          url.searchParams.set('respondentId', sourceSubject.userId);
+        } else if (!isSelfReporting && loggedInUser && loggedInUser.userId) {
+          url.searchParams.set('respondentId', loggedInUser.userId);
+        }
+
+        // TODO: Remove once the web app is updated to process `targetSubjectId` instead of `subjectId`
```
farmerpaul, line 204: Is there a JIRA ticket # for this?

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
```diff
@@ -17,14 +19,16 @@ import {
...
   mockSchema,
   mockSuccessfulHttpResponse,
-  mockGetRequestResponses,
 } from 'shared/utils/axios-mocks';
 import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 import { ParticipantsData } from 'modules/Dashboard/features/Participants';
 
 import { Activities } from './Activities';
+import { RespondentStatus } from '../../../types';
+import { ManagersData } from '../../Managers';
```
farmerpaul: I think absolute import paths might be a bit clearer?

src/shared/styles/styledComponents/Modal.ts
```diff
@@ -5,7 +5,7 @@ import theme from 'shared/styles/theme';
 import { StyledFlexColumn } from './Flex';
 
 export const StyledModalWrapper = styled(Box)`
-  padding: ${theme.spacing(2.4, 3.2)};
+  padding: ${theme.spacing(1.6, 3.2, 2.4)};
```
farmerpaul: It seems that there's a lot of inconsistency in Figma about the top padding measurement. It's 16px in some places, and 24px in many other places. Maybe you could get away with picking the average – 20px… 😏

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -64,42 +71,105 @@ export const LabeledUserDropdown = ({
...
+            nickname={tag === 'Team' ? null : nickname}
             {...psProps}
             boxProps={{
-              sx: { p: theme.spacing(0.6, 1.6), cursor: 'pointer' },
+              sx: {
+                p: theme.spacing(0.6, 1.6),
+                cursor: 'pointer',
+                ':last-child': {
+                  borderBottom: `${variables.borderWidth.md} solid ${variables.palette.surface_variant}`,
+                },
```
farmerpaul: This is a total nit, but I noticed that the last group has a bottom border, and I don't think it should:

Screenshot 2024-05-22 at 3 03 42 PM


So instead of applying it here in the `renderOption` callback, you could apply it in `renderGroup` below:

```tsx
        renderGroup={(params) => {
          const { key, group, children } = params;

          return (
            <Box
              component="li"
              key={key}
              sx={{
                '&:not(:last-child)': {
                  borderBottom: `${variables.borderWidth.md} solid ${variables.palette.surface_variant}`,
                },
              }}
            >
              <StyledLabelBoldLarge
                sx={{
                …
```

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -109,18 +179,34 @@ export const LabeledUserDropdown = ({
...
       />
+      <RenderIf condition={shouldShowWarningMessage}>
+        <StyledFlexTopCenter
+          sx={{
+            gap: 1.6,
+            backgroundColor: variables.palette.yellow_light,
+            p: theme.spacing(0.8, 1.6),
+            mt: -1.6,
+            borderBottomLeftRadius: variables.borderRadius.sm,
+            borderBottomRightRadius: variables.borderRadius.sm,
```
farmerpaul: There's a lot of custom styling happening in the component definition now. I'm personally not too fussy about that, but I've seen SS folks request that we keep `sx` attributes to 2 or 3 max, and otherwise create a styled component for more advanced styling.

farmerpaul (approved):
Tested and reviewed the code, looks awesome! Great work. I had some only minor comments, nothing blocking. ✨

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Yeah the only use case I created it for was avoiding the conditionals inside the JSX. I find logic + markup hard to read, especially when there's a lot of it. This is a good point I hadn't considered - `MyComponent` is still a function that gets called, even if its output doesn't get used.

I could change the definition of `RenderIf` to something like this

```typescript
/**
 * Render a component tree if a condition is met, otherwise render a fallback component. By default, the fallback is null.
 */
export const RenderIf = <T extends React.ComponentType<any>>({
  condition,
  Component,
  fallback,
  ...props
}: RenderIfProps<T> & React.ComponentProps<T>) => {
  if (condition) {
    return <Component {...props} />;
  }

  return fallback?.() ?? null;
};
```

So that its usage would now look like this

```typescript
<RenderIf condition={!!title} Component={MyComponent} title={title!} />
```

This would avoid the unnecessary evaluation of `MyComponent`, but we still have to assert that the title is not undefined. I don't love it, but I think it's an improvement. What do you think?

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: It would. This is a mistake caused by a lot of refactoring 😅

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: Not necessarily. I forgot to mention this in the notes section, but I'm not clear on how we should handle the population of dropdown fields that could reasonably contain a lot of entries. So my approach here is to just do an initial load of 100 entries (an arbitrary number), and then any additional ones can be found via search (which queries the API)

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: No, but I was thinking we shouldn't need one. This is just a precaution because this will get merged before https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/459. Technically it doesn't have to be removed, so it's just something we can remove as part of some other PR

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
sultanofcardio: Indeed. The struggle with imports continues 😞

src/shared/styles/styledComponents/Modal.ts
sultanofcardio: I'll touch base with Matt about this and circle back

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Very sharp on this one. Thanks man

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: I honestly don't mind it either, but I'll do some clean up

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
farmerpaul: I don't love it either… 😅 I'm not sure it necessarily makes the code easier to read than logic/conditionals in JSX, and I'm not really a fan of the non-null assertion.

I guess another alternative would just be:

```tsx
  const tooltip = !!tooltip && <Tooltip …/>;

  …

  return (
    <StyledFlexColumn sx={{ gap: 1.6, ...sx }}>
      <Box sx={{ display: 'flex', gap: 0.4 }}>
        <StyledTitleMedium sx={{ fontWeight: 700, color: variables.palette.on_surface }}>
          {label}
        </StyledTitleMedium>
        {tooltip}
        …
```

🤷🏻‍♂️

sultanofcardio: I'll keep working on it. For now I'll fall back to a conditional

src/shared/styles/styledComponents/Modal.ts
sultanofcardio: I [confirmed](https://metalab.slack.com/archives/C06E0CN3RT5/p1716905495728559) with Sam that this should be 24px 👌

PR #1754, opened 2024-05-24
feat: [Admin Panel] - add Date with conditions in Rule row (M2-6691)

alex-i-e:

### 📝 Description

🔗 [Jira Ticket M2-6691](https://mindlogger.atlassian.net/browse/M2-6691)


src/modules/Builder/components/ConditionRow/Condition/Condition.tsx
```diff
@@ -17,6 +25,138 @@ import {
   getStateOptions,
 } from './Condition.utils';
 
+export const SwitchCondition = ({
```
IhorNesterenko99: Maybe it's better to create its own component for SwitchCondition? This will make Condition.tsx not so big.

IhorNesterenko99 (approved):

PR #1757, opened 2024-05-27
feat: Add Applet's Overview page (M2-6555)

JustinNusca:
### 📝 Description

🔗 [M2-6555](https://mindlogger.atlassian.net/browse/M2-6555): [FE][Overview] Applet Overview Page

> [!NOTE]
> This PR depends upon the backend changed introduced in https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1322.
>
> You should check out that branch before running this PR locally for testing purposes.

This PR adds content for the "Overview" tab to the Applet Dashboard. It fetches data from a new API endpoint, `/answers/applet/${appletId}/submissions`, and renders the recent submissions in a paginated table. It also renders a new "QuickStats" component, which renders relevant data points for the Applet.

As part of this change, I modified the DashboardTable component to allow for passing arbitrary `TableCell` props to the row table cells. These table cell props may continue to be passed via the existing `rows` prop, which I've updated to match the TableCell prop types.

### 📸 Screenshots

| Populated | Empty State |
|-|-|
| localhost_3000_dashboard_c81bee80-1da8-479f-9237-212a850e0c58_overview | localhost_3000_dashboard_7b743a48-3840-453b-9481-1a8d00908c45_overview |

### 🪤 Peer Testing

From the Applet dashboard page…

1. Navigate to the Overview tab, if you are not there already.
2. Observe the presence of page content, including a "Quick stats" box and a table containing a list of submissions.
3. In the Quick stats box, press the "Add Participant" button next to the participant count.
4. Proceed through the dialog and add a new participant. After submitting, observe that data is refreshed and the participants count is updated accordingly.
5. Complete some activities as different particpants — both directly within the web app, or via the "Take Now" flow to complete an activity on behalf of another subject.
6. After completion, return to the Overview tab, and observe that the list of recent submissions has been updated accordingly.

[M2-6555]: https://mindlogger.atlassian.net/browse/M2-6555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/features/Applet/Overview/QuickStats/QuickStats.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const QuickStats = ({ stats = [], ...otherProps }: QuickStatsProps) => (
+  <StyledRoot component="ul" {...otherProps}>
+    {stats.map(({ label, icon, value, sx, ...otherStatProps }, i) => (
+      <StyledFlexColumn
+        component="li"
+        key={i}
+        sx={{ gap: 1.2, flexGrow: 1, position: 'relative', ...sx }}
+        {...otherStatProps}
+      >
+        <Svg height="24" fill={variables.palette.outline_variant} id={icon} width="24" />
```
farmerpaul: This should be `on_surface_variant` according to [Figma](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=1761-40649&t=pntZXC2cWLgB76WL-4):
```suggestion
        <Svg height="24" fill={variables.palette.on_surface_variant} id={icon} width="24" />
```

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
```diff
@@ -0,0 +1,151 @@
...
+      {
+        label: i18n.t('appletOverview.statParticipants'),
+        icon: 'profile' as const,
+        value: (
+          <StyledFlexTopCenter component="span" sx={{ gap: 1.6 }}>
+            {(participantsCount ?? 0).toLocaleString(i18n.language)}
+
+            <IconButton onClick={commonProps.onPressAddParticipant}>
+              <Svg aria-label={i18n.t('addParticipant')} id="add" width={20} height={20} />
+            </IconButton>
```
farmerpaul: Just noting a few differences between your implementation and what's in Figma (perhaps this was a recent change by design?):
- My impression based on it being hidden in [this frame](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=1761-40649&t=pntZXC2cWLgB76WL-4), and the title of the [frame in which it does appear](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=7517-215226&t=pntZXC2cWLgB76WL-4) – "Participants Hover" – is that the add participant button is only visible when hovering over the Participants cell
- It seems to have a slightly different colour variation in Figma; not sure if that's something we still need to add support for in the theme (I guess IconButtons use a distinct configuration from regular Buttons)
- In Figma, the button is aligned to the upper-right corner of the cell as opposed to beside the value; was there a design decision to move it here to account for the wider cell width?

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
```diff
@@ -0,0 +1,151 @@
...
+                      ) : (
+                        <ParticipantSnippet
+                          tag={targetSubjectTag}
+                          secretId={targetSubjectId}
+                          nickname={targetNickname}
+                        />
+                      )}
+                    </StyledMaybeEmpty>
+                  ),
+                  sx: { maxWidth: theme.spacing(31.2) },
```
farmerpaul, line 114: Seems the row heights are collapsing a bit in height without the presence of the icon buttons; to maintain the sense of space here and align with Participant & Manager table styling, a minimum row height might be in order (I guess `<td>`s treat `height` as minimum height, and don't respond to `minHeight` 🤷🏻‍♂️).

I suppose this could also be part of the `DashboardTable` base styling? I noticed it feels a little too tight on `/dashboard/respondents` too.
```suggestion
                  sx: { maxWidth: theme.spacing(31.2), height: theme.spacing(6.4) },
```

src/modules/Dashboard/features/Applet/Overview/Overview.tsx
```diff
@@ -1,18 +1,86 @@
...
+                // DashboardTable passes it's `handleChangePage` prop a 0-based
+                // index value, but accepts its `page` prop as 1-based.
+                setPage(nextPage + 1);
+              }}
+              page={page}
+              rowsPerPage={limit}
+            />
+          </StyledFlexColumn>
+
+          {appletId && (
```
farmerpaul: There's a minor UX quirk that happens here, but doesn't happen on the Participants screen:
1. Click the add participant button
2. Select Limited or Full account and press Next
3. Close the popup
4. Click the add participant button
**Current behaviour:** The popup opens but is already on the second step.
**Expected behaviour** (see last bullet of [this task's AC](https://mindlogger.atlassian.net/browse/M2-5436)): If at any time you click the X in the popup, it closes the popup and next time it's opened, it resets to the first step

The reason it works in `Participants.tsx` is because we render the popup conditionally based on the popup visibility state variable, which unmounts the popup, causing it to be reset to the first step when remounted. However, what I don't _love_ about that is that it disables the subtle hiding CSS transition (fade out) of the popup and its overlay provided by Material UI. Also, it makes the `popupVisible` prop pointless. But at the time, I opted to implement it that way to avoid diverging from existing patterns used by other popups.

You could opt to improve upon this by, in `AddParticipantPopup`, adding an effect that resets its state whenever `popupVisible` becomes `false`.

Alternatively, you could just add `addPopupOpen` here and have it behave the same as it does on the Participants screen:

```suggestion
          {appletId && addPopupOpen && (
```

src/modules/Dashboard/features/Applet/Overview/Overview.tsx
```diff
@@ -1,18 +1,86 @@
...
+          />
+
+          <StyledFlexColumn sx={{ gap: 1.6 }}>
+            <StyledTitleLarge>{t('appletOverview.titleRecentSubmissions')}</StyledTitleLarge>
+
+            <DashboardTable
+              {...dashboardTableProps}
+              handleChangePage={(_, nextPage) => {
+                // DashboardTable passes it's `handleChangePage` prop a 0-based
+                // index value, but accepts its `page` prop as 1-based.
```
farmerpaul, line 67: 🫠

src/modules/Dashboard/features/Applet/Overview/QuickStats/QuickStats.styles.ts
```diff
@@ -0,0 +1,24 @@
+import styled from '@emotion/styled';
+import { Box } from '@mui/material';
+
+import { theme, variables } from 'shared/styles';
+
+export const StyledRoot = styled(Box)({
+  borderRadius: variables.borderRadius.lg2,
+  boxShadow: '0px 0px 16px 0px #00000014',
```
farmerpaul: In Figma this is using the token "Soft shadow", and while we do have a token called `variables.boxShadow.soft`, its blur radius is only 8px. Maybe we need to create a new one, `variables.boxShadow.soft2`, for the 16px-version?

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  content: () => (
+                    <StyledMaybeEmpty>
+                      {formatDistanceStrict(
+                        new Date(
+                          createdAtDate.getTime() - createdAtDate.getTimezoneOffset() * 60000,
+                        ),
+                        new Date(),
+                        {
+                          addSuffix: true,
+                          locale: locales[i18n.language as keyof typeof locales],
```
farmerpaul: You could also use the `Languages` const in `api.const.ts`:
```suggestion
                          locale: locales[i18n.language as Languages],
```

farmerpaul (commented):
Awesome work @JustinNusca, looks good and works well! I had a few comments mostly relating to some minor differences I saw in Figma and a minor UX issue.

~~Also for some reason that is currently confounding me, I can't perform Take Now. I have the latest `dev` branch of `mindlogger-web-refactor`, and I'm using the BE branch you indicated in the PR. I try to perform Take Now from the popup, and it take me to the web app but just shows me the list of activities and doesn't automatically start the assessment. I don't think this has anything to do with your PR nor do I feel I need to be able to test that, it's just odd! It's been over a week since I performed Take Now so maybe something's changed somewhere that requires an extra step on my part…~~

I was able to do Take Now fine when reviewing https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/460, so maybe it's related to recent parameter/API changes.

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
JustinNusca: Yeah, the designs definitely got some tweaks within the last day or so — I just saw this for the first time this afternoon 😅. It looks like there's some very recent associated comments and feedback so this might still be being worked on.

The existing placement of this button is 100% my own improvisation to meet the AC for the ticket, in lieu of not having any design in place for it, at the time.

I agree that the appearance the designers are iterating on now is better than what I've done — but I also agree with you about the uncertainty around how it might appear when there's just two data points, which they haven't accounted for yet.

I'm not happy with how what I came up with looks, but at the same time I'm not sure how finalized what's present now is. 🤔 I'll see if I can't bring it closer to the Figma for the time being.

src/modules/Dashboard/features/Applet/Overview/QuickStats/QuickStats.styles.ts
JustinNusca: Good call out — I did a search and could find the equivalent styles being used elsewhere, so I added it as a one-off as shown. Since design has a specific token for this though, I went ahead and added one as well.

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
JustinNusca: I went ahead and added it to the `DashboardTable` directly — I wound up having to apply it to the table cell styling specifically, rather than at the row level, but 🤷‍♀️.

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
JustinNusca: I've pushed some updates that should hopefully improve the look of things here. The stats UI has been tweaked to match the latest revisions, and while the [Overview screen in Figma](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=6934-109894&t=UR48vWAwTRF282b0-4) still depicts four items, the [Stats component itself](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=6934-109894&t=UR48vWAwTRF282b0-4) includes design variations for 1 or more.

I don't see any designs that include the "Add Participant" button at the moment, but I've included a variation similar to what was previously present, [shown here](https://private-user-images.githubusercontent.com/1670836/335371352-6b2379e9-3c08-4b0e-b67c-fe287ac54d9b.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTcxMDU4MTcsIm5iZiI6MTcxNzEwNTUxNywicGF0aCI6Ii8xNjcwODM2LzMzNTM3MTM1Mi02YjIzNzllOS0zYzA4LTRiMGUtYjY3Yy1mZTI4N2FjNTRkOWIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUzMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MzBUMjE0NTE3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZmQ0MGU0ODI3NWNmNWMzNzZkZTI1NmQzNTQ0ZjgxMzkwODU3NDQ3ZjA2M2I1MWU3NGZjMzU2ZmUyNGQ0OWRmYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.8tD0qIikOCmUqZx9h58BljHnyI1AoZdXKjw615fgdtM) in the updated screenshots I added. Mostly from my own reservations about accessibility, I'm a bit reluctant to make this hidden until mouse-over, unless the designers explicitly request it.

src/modules/Dashboard/features/Applet/Overview/Overview.tsx
JustinNusca: Yeah — I wound up just making the suggested change here (`appletId && addPopupOpen && ...`). I would normally love to ensure that we have properly functioning Modal transitions, but there's just enough complexity here around resetting state, and how the Popup component is actually multiple variable Modal components, and how I'd prefer to tackle this as a more holistic solution that allows us to avoid manually doing the same thing on all other Popups to solve the same problem ... that I think this is best done as a dedicated task on its own.

mbanting (approved):
LGTM. Nice work @JustinNusca

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
farmerpaul: Changes look great! But looking at the screens now, it appears the screen that displays an "Add Participant" button has been removed altogether. Maybe it was decided that adding participants from this screen no longer makes sense? I personally like it, but I think it might also be good to confirm with design, and if keeping it, make sure to mention this in JIRA to set QA expectations when they compare against Figma.

farmerpaul: @JustinNusca – if there's issues getting clarity here I think it's fine to with just resolve the convo, rebase and merge, in the interests of getting this PR across the line. Can always push a hotfix later.

farmerpaul (approved):

src/modules/Dashboard/features/Applet/Overview/Overview.utils.tsx
JustinNusca: Yeah, I'm finally circling back to this now that the relevant backend PR has been merged. 

There's a couple details that I don't think ever got sorted neatly (the designs have more-or-less settled on omitting the "Add Participant" button entirely, but the acceptance criteria still specifies it.

Everything is functional and polished enough that I agree with you: I'm going to proceed with merging, and if there _is_ outstanding details, handle it in a follow-up PR.

PR #1760, opened 2024-05-28
fix: Conditionally pass user ID to schedule-related components (M2-6775)

JustinNusca:

### 📝 Description

🔗 [M2-6775](https://mindlogger.atlassian.net/browse/M2-6775): [Admin Panel] It is possible to update Default schedule from the particular participant's individual schedule page

This PR addresses an issue with the Schedule page, where attempts to update the _default_ schedule from the Participant Details page for a subject with a default schedule would attempt to modify events as if they were on an _individual_ schedule.

This was because we were unconditionally passing the subject's User ID if available (i.e. if the subject was a _Full_ participant), instead of only if they were actively on an individual schedule.

#### Create/Edit Event

| Before | After |
|-|-|
| create-edit-before | create-edit-after |

#### Clear Scheduled Events

| Before | After |
|-|-|
| clear-before | clear-after |

### 🪤 Peer Testing

From the **Participant Details > Schedule** page for a *full* participant on the *default* schedule…
1. Press the "Create Event" button. Proceed to create an event.
2. Notice that the event has been created and applied to the default schedule.
3. Refresh the page. Notice that the subject is still on the default schedule, and changes to the schedule have persisted.
4. Press the "Clear Scheduled Events" button. Confirm your choice in the resulting dialog.
5. Notice that scheduled events have been removed from the default schedule.
6. Refresh the page. Notice that the subject is still on the default schedule, and changes to the schedule have persisted.

From the **Participant Details > Schedule** page for a *full* participant on the *individual* schedule…
1. Press the "Create Event" button. Proceed to create an event.
2. Notice that the event has been created and applied to the individual schedule.
3. Refresh the page. Notice that the subject is still on the individual schedule, and changes to the schedule have persisted.
4. Press the "Clear Scheduled Events" button. Confirm your choice in the resulting dialog.
5. Notice that scheduled events have been removed from the individual schedule.
6. Refresh the page. Notice that the subject is still on the individual schedule, and changes to the schedule have persisted.

[M2-6775]: https://mindlogger.atlassian.net/browse/M2-6775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Schedule/CreateEventPopup/CreateEventPopup.test.tsx
+++ b/src/modules/Dashboard/features/Applet/Schedule/CreateEventPopup/CreateEventPopup.test.tsx
@@ -28,6 +28,8 @@ export const preloadedState = {
     },
   },
   calendarEvents: {
+    alwaysAvailableVisible: { data: true },
+    scheduledVisible: { data: true },
     createEventsData: {
       data: [
         {
--- a/src/modules/Dashboard/features/Applet/Schedule/Legend/Legend.test.tsx
+++ b/src/modules/Dashboard/features/Applet/Schedule/Legend/Legend.test.tsx
@@ -1,8 +1,11 @@
-import { screen } from '@testing-library/react';
+import { PreloadedState } from '@reduxjs/toolkit';
+import { fireEvent, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import axios from 'axios';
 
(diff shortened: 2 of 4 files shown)


farmerpaul (approved):
Tested, and changes look good, good to go! 🚀

PR #1761, opened 2024-05-28
fix(mi): Update applet permissions (M2-6721, M2-6727, M2-6728)

ChaconC:

### 📝 Description

Adds additional permission checks for removing certain elements for `Coordinator`, `Reviewer` and `Editor` roles

🔗 [Jira Ticket M2-6721](https://mindlogger.atlassian.net/browse/M2-6721)
🔗 [Jira Ticket M2-6727](https://mindlogger.atlassian.net/browse/M2-6727)
🔗 [Jira Ticket M2-6728](https://mindlogger.atlassian.net/browse/M2-6728)

Changes include:

- Add `checkIfCanAccessData` util
- Remove "Add Activity" buttons for Reviewers, Coordinator roles
- Remove "Add Participant" buttons for Reviewer roles
- Remove navigation to DataViz for Coordinator, Editor roles
- Remove data export buttons for Coordinator, Editor roles

### Applet > Activities (Coordinator)
* No Settings button
* No Add Activity button
* No Export buttons

| CleanShot 2024-05-28 at 16 29 30 | CleanShot 2024-05-28 at 16 30 12 |

### Applet > Activities (Reviewer)
* No Settings button
* No Add Activity button

| CleanShot 2024-05-29 at 08 21 55 | CleanShot 2024-05-29 at 08 21 41 |

### Applet > Participants (Reviewer)
* No Add Participant button
* No Settings button

| CleanShot 2024-05-29 at 08 25 26 | CleanShot 2024-05-29 at 08 25 14 |

### Hiding Dots Menu 
| CleanShot 2024-05-28 at 16 33 11 | CleanShot 2024-05-28 at 16 33 03 |

#### Setup

* Create an Applet with Reviewer, Coordinator and Editor accounts
* The applet should have at least 1 response so that DataViz navigation works correctly

#### Testing
**Log in as Reviewer**
* Select shared workspace
* Navigate to Applet > Activities
  * Validate no "Add Activity" button 
* Navigate to Applet > Participants
  * Validate no "Add Participant" button
  * Validate no Settings button
* Select a Participant
* Navigate to DataViz

**Log in as Coordinator**
* Select shared workspace
* Navigate to Applet > Activities
  * Validate no "Add Activity" button 
* Navigate to Applet > Participants
  * Validate no Export button
  * Validate no Settings button
* Select a Participant
* Confirm no navigation to DataViz

### ✏️ Notes

Expected permissions can be reviewed in the [Confluence Page](https://mindlogger.atlassian.net/wiki/x/FABSCg) checklists for each role.

| Role | View Participants | Add/Edit Participants | Add/Edit Activity | Data Export/Visualization | Take Now | Assign* |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Owner/Manager | ✅ | ✅ | ✅ | ✅ | ✅ |  ✅ | 
| Coordinator | ✅ | ✅ | ❌ | ❌ | ❌ |  ✅ | 
| Reviewer | ✅ | ❌ | ❌ | ✅ | ✅ | ❌ | 
| Editor* | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | 

* Assign button is controlled by the `enableActivityAssign` feature flag

**Note about Editor Role:**
~Because Editors are not allowed to view participants, all routes inside an applet will fail to render due to being unable to retrieve `/respondents`. Will link a related ticket for this issue.~

The previously mentioned Editor Role limitation has been addressed and Editors should be able to view content in the Activities tab only.


src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -31,12 +32,16 @@ export const getParticipantActions = ({
...
   dataTestid,
   showAssignActivity = false,
+  roles,
 }: GetParticipantActionsProps) => {
   const context = { respondentId, respondentOrSubjectId, email, secretId, nickname, tag };
   const isUpgradeable = status === RespondentStatus.NotInvited;
   const isPending = status === RespondentStatus.Pending;
   const isEditable = !!filteredApplets?.editable.length;
   const isViewable = !!filteredApplets?.viewable.length;
+  const canEdit = checkIfCanEdit(roles) && isViewable && !isPending;
```
farmerpaul: I'm a little confused by the difference between `isEditable` and `canEdit`… they are defined slightly differently. Could we get away with using just one, or if not, name them more distinctly?

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -418,7 +422,7 @@ export const Participants = () => {
             sx={{ width: '32rem' }}
             data-testid={`${dataTestid}-search`}
           />
-          {appletId && (
+          {canAddParticipant && (
```
sultanofcardio, line 449: Remember to add this to the empty state add participant button below as well

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
```diff
@@ -31,17 +31,15 @@ export function useFlowGridMenu({
...
   const canDoTakeNow =
     featureFlags.enableMultiInformantTakeNow &&
     hasParticipants &&
-    (isManagerOrOwner(roles?.[0]) ||
-      roles?.includes(Roles.Coordinator) ||
-      roles?.includes(Roles.SuperAdmin));
+    (isManagerOrOwner(roles?.[0]) || roles?.includes(Roles.SuperAdmin));
+
+  const canAccessData = checkIfCanAccessData(roles);
+  const showDivider = (canEdit || canAccessData) && canDoTakeNow;
```
farmerpaul:
```suggestion
  const showDivider = (canEdit || canAccessData) && (featureFlags.enableActivityAssign || canDoTakeNow);
```

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
```diff
@@ -31,17 +31,15 @@ export function useFlowGridMenu({
...
 
   const canDoTakeNow =
     featureFlags.enableMultiInformantTakeNow &&
     hasParticipants &&
-    (isManagerOrOwner(roles?.[0]) ||
-      roles?.includes(Roles.Coordinator) ||
-      roles?.includes(Roles.SuperAdmin));
+    (isManagerOrOwner(roles?.[0]) || roles?.includes(Roles.SuperAdmin));
+
+  const canAccessData = checkIfCanAccessData(roles);
```
JustinNusca, line 43: 👍 When you messaged me the other day about this, I was already thinking "why aren't we just using a shared util fn here" — I wasn't even aware of `checkIfCanEdit` previously.

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
```diff
@@ -31,17 +31,15 @@ export function useFlowGridMenu({
...
-    roles?.includes(Roles.SuperAdmin);
+  const canEdit = checkIfCanEdit(roles);
 
   const canDoTakeNow =
     featureFlags.enableMultiInformantTakeNow &&
     hasParticipants &&
-    (isManagerOrOwner(roles?.[0]) ||
-      roles?.includes(Roles.Coordinator) ||
-      roles?.includes(Roles.SuperAdmin));
+    (isManagerOrOwner(roles?.[0]) || roles?.includes(Roles.SuperAdmin));
```
JustinNusca: Any particular reason you opted not to abstract this to a similar  "checkRoles" utility fn? I notice this is the same check made in `ActivityGrid.utils.tsx`.

src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
```diff
@@ -19,6 +21,10 @@ export const ActivitiesToolbar = ({
   const { t } = useTranslation('app');
   const { featureFlags } = useFeatureFlags();
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+  const workspaceRoles = workspaces.useRolesData();
+  const roles = appletId ? workspaceRoles?.data?.[appletId] : undefined;
+
```
JustinNusca: Extremely nit-picky, and I'm not opposed to linebreaks for readability or whatever, but is the extra line break here intentional? 🧐

src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
```diff
@@ -74,15 +80,17 @@ export const ActivitiesToolbar = ({
...
+            {canEditApplet && (
+              <Button
+                component={Link}
+                to={generatePath(page.builderAppletActivities, { appletId })}
+                variant="contained"
+                data-testid={`${dataTestId}-add-activity`}
+                sx={{ minWidth: theme.spacing(13.2) }}
+              >
+                {t('addActivity')}
+              </Button>
```
sultanofcardio: In addition to this, according to scenario 3 of [M2-6727](https://mindlogger.atlassian.net/browse/M2-6727), it sounds like the reviewer should not be able to see the activities tab either. I'm not sure if that's what was meant, since the language was vague but I thought I'd point that out. Are you clear on this?

In case it turns out they shouldn't, I'm thinking of what the solution would be. In our multi-informant environment we use a pop-up like this to block access to the team tab

image

However, for the schedule tab I see something like this, which feels a little cleaner

image

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.utils.tsx
```diff
@@ -44,7 +42,7 @@ export const getActivityActions = ({
       action: exportData,
       title: t('exportData'),
       context: { appletId, activityId },
-      isDisplayed: true,
+      isDisplayed: canAccessData,
```
farmerpaul, line 45: In `FlowGrid.hooks.tsx`, the Export Data menu item has `disabled: true` at all times, not yet sure why. Presumably because it's unimplemented? However, here, it's not disabled, but when it's available (i.e. when `canAccessData == true`), it just shows a TODO alert.

Do you know if this action is expected to be implemented by launch? In any case, it seems that `ActivityGrid` and `FlowGrid` should treat this menu item the same…

src/shared/utils/checkRole.ts
```diff
@@ -11,3 +11,10 @@ export const checkIfCanEdit = (roles?: Roles[]) =>
       roles?.includes(Roles.Editor) ||
       roles?.includes(Roles.SuperAdmin),
   );
+
+export const checkIfCanAccessData = (roles?: Roles[]) =>
+  Boolean(
+    isManagerOrOwner(roles?.[0]) ||
+      roles?.includes(Roles.SuperAdmin) ||
+      roles?.includes(Roles.Reviewer),
```
sultanofcardio: There seems to be conflicting information in these tickets. Scenario 4 of [M2-6727](https://mindlogger.atlassian.net/browse/M2-6727) says that “A reviewer can only view or export data.” However, the 3rd case of [M2-6728](https://mindlogger.atlassian.net/browse/M2-6728) says that “Reviewer should not be able to export data in shared account”.

Based on [Cari's comment](https://mindlogger.atlassian.net/browse/M2-6727?focusedCommentId=21467), it seems the reviewer should be able to export data so maybe we should get the description of [M2-6728](https://mindlogger.atlassian.net/browse/M2-6728) updated

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
```diff
@@ -128,6 +134,12 @@ export const Activities = () => {
     );
   };
 
+  const getClickHandler = () => {
+    if (!subjectId || !appletId || !canAccessData) return undefined;
+
+    return onClickItemHandler;
+  };
```
JustinNusca, line 142: If looks like the actual handler itself has an early return statement that makes the 2/3 of the same checks.

I wonder if it would be more conventional to just update [those lines](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1761/files#diff-55e1e9f202a8f57998166fcac43a4bb07de7a8d708ad50847b1675c7c48033f8R114-R115) to be something like…

```jsx
const handleClickItem = (activityId: string) => {
  if (!subjectId || !appletId || !canAccessData) return;
  // …rest of fn…
```

… compared to adding a new wrapper function that conditionally returns a handler on the basis of mostly duplicate checks.

sultanofcardio:
The tickets look like they need some story points as well

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
```diff
@@ -166,12 +166,9 @@ describe('Dashboard > Applet > Activities screen', () => {
...
 
-      await waitFor(() =>
-        expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-      );
-      userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
-
-      if (canEdit) {
+      const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
+      if (actionDots && canEdit) {
+        userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting: Can you use the existing reference to `actionDots` since its existence is being checked?

```suggestion
        userEvent.click(actionDots);
```

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
```diff
@@ -211,12 +208,10 @@ describe('Dashboard > Applet > Activities screen', () => {
...
 
-        await waitFor(() =>
-          expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-        );
-        userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
+        const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
 
-        if (canDoTakeNow) {
+        if (actionDots && canDoTakeNow) {
+          userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting:
```suggestion
          userEvent.click(actionDots);
```

src/modules/Dashboard/features/Applet/Activities/Activities.test.tsx
```diff
@@ -259,12 +254,13 @@ describe('Dashboard > Applet > Activities screen', () => {
...
 
-        await waitFor(() =>
-          expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-        );
-        await userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
-
-        await waitFor(() => expect(screen.queryByTestId(`${testId}-activity-take-now`)).toBe(null));
+        const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
+        if (actionDots) {
+          await userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting:
```suggestion
          await userEvent.click(actionDots);
```

src/modules/Dashboard/features/Participant/Activities/Activities.test.tsx
```diff
@@ -163,12 +163,9 @@ describe('Dashboard > Applet > Participant > Activities screen', () => {
...
 
-      await waitFor(() =>
-        expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-      );
-      userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
-
-      if (canEdit) {
+      const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
+      if (actionDots && canEdit) {
+        userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting:
```suggestion
        userEvent.click(actionDots);
```

src/modules/Dashboard/features/Participant/Activities/Activities.test.tsx
```diff
@@ -207,12 +204,9 @@ describe('Dashboard > Applet > Participant > Activities screen', () => {
...
 
-        await waitFor(() =>
-          expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-        );
-        userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
-
-        if (canDoTakeNow) {
+        const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
+        if (actionDots && canDoTakeNow) {
+          userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting:
```suggestion
          userEvent.click(actionDots);
```

src/modules/Dashboard/features/Participant/Activities/Activities.test.tsx
```diff
@@ -254,12 +248,17 @@ describe('Dashboard > Applet > Participant > Activities screen', () => {
...
 
-        await waitFor(() =>
-          expect(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]).toBeVisible(),
-        );
-        await userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
-
-        await waitFor(() => expect(screen.queryByTestId(`${testId}-activity-take-now`)).toBe(null));
+        const actionDots = screen.queryAllByTestId(`${testId}-activity-actions-dots`)[0];
+        if (actionDots) {
+          await userEvent.click(screen.getAllByTestId(`${testId}-activity-actions-dots`)[0]);
```
mbanting:
```suggestion
          await userEvent.click(actionDots);
```

farmerpaul (commented):
Hey @ChaconC, tested and it seems to be working as described, though I haven't delved into the permissions as deeply as it looks like @sultanofcardio has (kudos to him!). I identified some other minor things in the code.

The other thing I identified is, when logged in as a Coordinator, navigating to **Applet > Participant Details > Schedule** tab shows an error alert, and kicks you out of the applet and workspace altogether. I think it should probably show the same thing that **Applet > Schedule** tab shows (a less jarring no-permissions state). Not sure if this is worth scoping into your current PR but it does seem related and hopefully a small thing?

src/shared/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -23,8 +23,18 @@ export const ActionsMenu = <T = unknown,>({
     setAnchorEl(null);
   };
   const openMenu = Boolean(anchorEl);
+  const hasMenuItems = useMemo(() => {
+    // Hides the ActionMenu if all the actions include .isDisplayed === false
+    const displayedMenuItems = menuItems.filter((item) => 'isDisplayed' in item).length;
+    if (displayedMenuItems) {
+      return menuItems.filter((item) => item.isDisplayed === true).length > 0;
+    }
```
mbanting: You can try using `some` instead of checking existence via the length

```suggestion
const hasMenuItems = useMemo(() => {
  // Check if there are any items with the 'isDisplayed' property
  const hasDisplayedProperty = menuItems.some((item) => 'isDisplayed' in item);

  // If there are items with 'isDisplayed' property, check if any of them are displayed
  if (hasDisplayedProperty) {
    return menuItems.some((item) => item.isDisplayed);
  }```

mbanting (dismissed):
Looks great @ChaconC . I tested the suggested peer testing and it's all working as specified.

Left a couple of minor suggestions.

src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
JustinNusca: I wanted to piggy-back off of this comment, because I ran into a couple similar issues. 

I'm not sure if it's within the scope of these tickets, which seem very specific about particular buttons/menu items, rather than access to tabs per-se, but I think it's worth discussing.

While the Applet-level Schedule tab has a nice "Forbidden" empty state for Reviewers, the Schedule tab at the Participant level raises the same "No access to applet" alert and redirects to the applets list:
 
Screenshot 2024-05-29 at 4 57 00 PM

It seems like maybe we should be consistent in our behaviour here? I think the "Forbidden" empty states are preferable and less disruptive, but if that's the case then we need updated empty states for the Participants and Teams tab as well — the default empty states for both include buttons prompting the user to "Add a participant/manager", which ought to be conditional based on role. (eg, per @sultanofcardio's [other comment](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1761#discussion_r1619563399).

src/shared/utils/checkRole.ts
mbanting: To potentially further narrow the permissions for a reviewer, another [comment](https://mindlogger.atlassian.net/browse/M2-6727?focusedCommentId=21467) by Cari and an additional reference in [Confluence](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/172589139/Invite+Reviewer+Checklist) states Reviewer can 
> view or export ONLY assigned user's (Respondents) data

ChaconC: @mbanting correct, that's the comment I went with, and Reviewers only see the participants they have assigned + team members so this seems to cover the criteria.

Edit: Added a note for QA in M2-6728 for clarification, but my take is Reviewers are able to export data for their assigned participants.

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
ChaconC: No reason really, was just limiting my changes. I'll add it to `checkRoles` and add test cases 👍

src/modules/Dashboard/features/Participant/Activities/Activities.tsx
ChaconC: I'm going to remove the check on the handler instead, I explicitly wanted to return `undefined` for those cases so that the cursor would change back to a pointer in the `ActivityCard` component (e.g. by having `onClick` be `undefined` as opposed to `() => void`)

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.utils.tsx
ChaconC: Not sure why `disabled: true` was added but also thought it was odd, maybe @JustinNusca can chime in.

Another thing, yes this functionality is meant for R1 and removing the TODO alerts will be covered by [M2-6736](https://mindlogger.atlassian.net/browse/M2-6736)

src/modules/Dashboard/features/Participants/Participants.utils.tsx
ChaconC: Good catch, looking at the file history this was a remnant from my original approach, then I added `checkIfCanAccessData` and replaced `canEdit` with `canAccessData` but didn't remove this boolean.

I'll move the role validation to `isEditable` and we can get rid of `canEdit`

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.utils.tsx
ChaconC: Another thing, yes this functionality is meant for R1 and removing the TODO alerts will be covered by [M2-6736](https://mindlogger.atlassian.net/browse/M2-6736)

src/shared/utils/checkRole.ts
sultanofcardio: It's the same comment, just wanted to highlight this so we could clarify for QA testing. I'll resolve this thread now 👍

src/modules/Dashboard/features/Applet/Activities/ActivitiesToolbar/ActivitiesToolbar.tsx
ChaconC: @sultanofcardio went ahead and added permission checks to return the `NoPermissions` component instead of the modal. This should fix the problem you highlighted for Reviewer and Coordinator roles, and also applied the same changes for all Editor tabs

@JustinNusca This wasn't a huge lift so I've added the same behavior in the Participant > Schedule for Reviewers and Editors

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.utils.tsx
JustinNusca: I meant to reply earlier to speak to this — yeah, I set `disabled: true` initially.

At the time, it seemed like we were in the practice of adding a lot of placeholder UI for functionality that wasn't yet present. I opted to just set `disabled`: it felt less confusing/disruptive from the user perspective than having an enabled-but-nonfunctional button and/or alert prompts, with the expectation that the presence of the menu item would be clear enough for whoever eventually needed to wire it up.

In the end, it seems like all these "Export Data" options had the requisite functionality for supporting these options all along, and yeah, there's a couple tasks currently scheduled for Sprint 10 that should handle them.

farmerpaul (commented):
Changes look good, thanks for making those! Before I approve, though, just wanted to point out one thing I'm experiencing:

I just tried testing access to an applet as a Reviewer (maybe that's what I did last time, too, even though I wrote Coordinator in my comment, can't remember), and I'm getting more of those applet access error popups when I click into Participant details. When it happens, I again get kicked out of both the applet and the workspace, which I presume shouldn't happen. I've recorded a video:

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/4d5f5474-3caf-47bb-ad29-f16d8ec93c77

mbanting (approved):
Thanks for addressing my suggestions/feedback!

ChaconC:
> Changes look good, thanks for making those! Before I approve, though, just wanted to point out one thing I'm experiencing:
I just tried testing access to an applet as a Reviewer (maybe that's what I did last time, too, even though I wrote Coordinator in my comment, can't remember), and I'm getting more of those applet access error popups when I click into Participant details. When it happens, I again get kicked out of both the applet and the workspace, which I presume shouldn't happen. I've recorded a video:

@farmerpaul I've removed `Reviewer` from `TakeNow` permissions as this was an error on my part, and I updated the check  in `TakeNowModal` to prevent the error in the clip.

JustinNusca (approved):
After your most recent changes, I'm no longer seeing the issue @farmerpaul described with the `reviewer` role, and the others appear to have the correct permissions reflected in the UI, too. 👍

Thanks for going ahead and updating the "forbidden"/empty states for the various screens, too!

PR #1762, opened 2024-05-28
fix(mi): Add `/add-user` redirect (M2-6737)

JustinNusca:
### 📝 Description

🔗 [M2-6737](https://mindlogger.atlassian.net/browse/M2-6737): [FE][Router] - Forward existing `/add-user` to Add Participant dialog

This PR updates route handling for the `dashboard/${appletId}/add-user` route. Because (when the `enableMultiInformant` feature flag is set) this route is no longer represented by a distinct tab on the Applet's Dashboard, and because the relevant functionality has been migrated to the "Add Participant" popup, we now redirect to `dashboard/${appletID}/participants`, with the "Add Participant" popup visible as part of the initial state.

To support this change, I swapped the state value for controlling the visibility of the "Add Participant" popup to instead use query parameters via React Router's `useSearchParams` hook.

### 🪤 Peer Testing

With the `enableMultiInformant` feature flag **set**…

1. Attempt to navigate to the `/add-user` route. For example, from `/dashboard/${appletId}/participants`, edit the URL bar in your browser to replace `/participants` with `/add-user`.
2. Observe that you have been redirected to `/dashboard/${appletId}/participants`, and the "Add Participant" popup is open.

With the `enableMultiInformant` feature flag **unset**…

1. Attempt to navigate to the `/add-user` route. For example, select the "Add Users" tab from the Applet Dashboard's list of tabs.
2. Observe that you can navigate to this page normally.

[M2-6737]: https://mindlogger.atlassian.net/browse/M2-6737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx
@@ -1,10 +1,11 @@
 import { useState } from 'react';
+import { Button } from '@mui/material';
 import { useTranslation } from 'react-i18next';
 import { Controller, useForm, useWatch } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
 
 import { Modal, Spinner, ToggleButtonGroup, ToggleButtonVariants } from 'shared/components';
-import { StyledErrorText, StyledModalWrapper } from 'shared/styles';
+import { StyledErrorText, StyledFlexEnd, StyledModalWrapper } from 'shared/styles';
 import { useFormError } from 'modules/Dashboard/hooks';
 import { NON_UNIQUE_VALUE_MESSAGE, Roles } from 'shared/consts';
 import { Mixpanel, getErrorMessage } from 'shared/utils';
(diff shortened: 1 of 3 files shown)


farmerpaul (approved):
Works great, nice choice.

I just discovered that, maybe since removing the "Add CSV" button in `AddParticipantPopup`, the "Next" button is now centre-aligned when it shouldn't be. This could be fixed by just adding `footerStyles={{ justifyContent: 'flex-end' }}` to line [178 of `AddParticipantPopup.tsx`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/cd2e62f787c5c3328b842d3543a9d06739edad22/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx#L178) (glad we kept that prop 😏), if you don't mind squeezing that in?

JustinNusca:
> I just discovered that, maybe since removing the "Add CSV" button in AddParticipantPopup, the "Next" button is now centre-aligned when it shouldn't be. This could be fixed by just adding footerStyles={{ justifyContent: 'flex-end' }} to line [178 of AddParticipantPopup.tsx](https://github.com/ChildMindInstitute/mindlogger-admin/blob/cd2e62f787c5c3328b842d3543a9d06739edad22/src/modules/Dashboard/features/Applet/Popups/AddParticipantPopup/AddParticipantPopup.tsx#L178) (glad we kept that prop 😏), if you don't mind squeezing that in?

Ah, I noticed that too, but was on the fence about addressing it as an issue. It seems a lot of the modals the design team has been putting out have favoured a centered button whenever there's a single option. I can tweak that, though.

**Updated**:

Screenshot 2024-05-30 at 1 54 50 PM

... Turns out that `justifyContent` is also set in the StyledComponent for the Modal's `StyledDialogActions`, and _that_ has priority over the same property passed via the `sx` prop, so I couldn't override it. I wound up just swapping to the `footer` prop I added in my other PR to work around that entirely.

PR #1767, opened 2024-05-29
fix(mi): improve Participants/Managers sorting (M2-6627)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-6627](https://mindlogger.atlassian.net/browse/M2-6627)

Adds support for conditional column sortability on the Participants, Managers, and dashboard Respondents screens via the API's new `orderingFields` attributed provided by https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1363. This allows sorting by encrypted fields.

- Fixes some column IDs to match the field names understood by the API
- Also fixes sorting when some participants are pinned
- Adds support for default selected ordering to the `useTable` hook
- If using older version of API without support for `orderingFields`, falls back to `sortable: true` for those columns as before
- Also includes these fixes on `/dashboard/respondents`:
  - the "Pin" column is no longer cropped to an incorrect width
  - the Account Status column renders the same pending/full/limited chip style as seen on the Participants screen

### 🪤 Peer Testing

Prerequisites:

- admin app connected to a local instance of the API that is running [this branch](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1363)
- an admin account
- an applet in that account
- several managers associated with the applet, with different **first & last names** & **emails**
- several participants associated with the applet, with different **nicknames**, including **limited**, **pending**, and **full** accounts

1. Navigate to **Applet > Managers**, and sort by **first name**, **last name**, and **email** in both ascending and descending order
    **Expected outcome:** Managers are sorted correctly.
2. Navigate to **Applet > Participants**, and sort by **nickname** in both ascending and descending order
    **Expected outcome:** Participants are sorted correctly.
3. Sort by **Account Type** in ascending order
    **Expected outcome:** Participant accounts are sorted in this order: `pending`, `full`, `limited`
4. Navigate to **Dashboard > Respondents**, and sort by **nickname** in both ascending and descending order
    **Expected outcome:** Participants are sorted correctly.
6. **Pin** some accounts to the top and sort by **nickname** in ascending order.
    **Expected outcome:** Pinned participants are always listed at the top, and sorted by nickname within that section, and the remaining participants are sorted by nickname within their section.


src/modules/Dashboard/features/Managers/Managers.utils.tsx
```diff
@@ -18,30 +18,30 @@ export const getHeadCells = (id?: string): HeadCell[] => {
     {
       id: 'firstName',
       label: t('firstName'),
-      enableSort: true,
+      enableSort: sortableColumns?.includes('firstName') ?? true,
```
JustinNusca: I'm not 100% sure about the fallback value (`?? true`), both here and throughout the similar `getHeadCells` fns).

I get that it allows for this PR to be more "backwards compatible" in preserving the existing sort options in the event that this might be merged without the BE pull request (or otherwise address cases where `sortableColumns` would be undefined) — but this existing functionality is broken anyway, no? Would it make more sense to just default to `false`, and prevent the UI from exposing sort options that lack proper support?

To put it another way — I feel like correcting existing broken behaviour, even if just by omission, is better than preserving it.

src/modules/Dashboard/features/Respondents/Respondents.tsx
```diff
@@ -309,14 +331,13 @@ export const Respondents = () => {
         },
       }),
       status: {
-        content: () =>
-          status !== RespondentStatus.Invited && (
-            <StatusFlag
-              status={status}
-              onInviteClick={() => handleInviteClick({ respondentOrSubjectId, email })}
-              isInviteDisabled={!filteredRespondents?.[respondentOrSubjectId]?.editable.length}
-            />
```
JustinNusca, line 318: Is it perhaps worth removing the `StatusFlag` component entirely at this point? It looks like the only place it's rendered now is in its own test file. Seems like it might be simple enough to cleanup if we consider this an "unsupported" component now.

JustinNusca (approved):
First, I want to say, from the bottom of my heart:

> `fix: reduce React runtime warnings in console`

Thank you.

I haven't looked at the corresponding BE PR, so if there's additional insight there that might inform comments here, I don't have them 😅. 

I _did_ test this branch out, both against the BE PR and the current head of the `feature/multiinformant-metapod` BE branch, and wasn't able to find any issues — either with the new sorting behaviour, or this branch's backwards compatibility with the old behaviour (which is to say, you've managed to preserve its brokenness 😅).

I have a couple small questions, but nothing that I would consider critical or anything that would hold up the PR from my perspective. 👍

src/modules/Dashboard/features/Managers/Managers.utils.tsx
farmerpaul: Yes, you're probably right! I was on the fence about it but you've convinced me, will update.

src/modules/Dashboard/features/Respondents/Respondents.tsx
farmerpaul: Good catch, removing.

PR #1769, opened 2024-05-31
feat(mi): Add context menu details for pending accounts (M2-6459)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-6459](https://mindlogger.atlassian.net/browse/M2-6459)

This PR updates the context menu actions in the participants table to include invitation summary details, and some context actions related to invitations.

> [!NOTE]
> This PR depends on https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1368. You must have those changes locally before attempting to test (maybe review that PR too while you're at it 😉)

#### Full Accounts

| image | image |

#### Pending Accounts

| image | image |

#### Limited Accounts

| image | image |

### 🪤 Peer Testing

>[!NOTE]
> Check if https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1368 has been merged yet. If it has, then be sure you have the latest changes on the `feature/multiinformant-metapod` API branch. If it has not, check out the `feature/M2 6747-respondents-invitation-data` (please insert the missing hyphen after M2) API branch and run your tests using that.

Create an applet workspace with at least one participant of each of the following types:
- Limited account
- Pending full account
- Full account
- Full account invitation rejected

1. Launch the admin web app, and navigate to the participants tab of the applet.
2. Open the context menu for each of the accounts above and verify that they match the screenshots provided (and Figma!).
3. Verify the functionality of the copy email address and copy invitation link actions

>[!TIP]
> An invited full account that rejected the invitation is the same as a limited account

### ✏️ Notes

N/A


src/modules/Dashboard/features/Applet/AddUser/InvitationsTable/InvitationsTable.utils.test.ts
```diff
@@ -1,6 +1,8 @@
 import { mockI18Next } from 'shared/tests';
 
 import { getHeadCells, getInvitationsTableRows } from './InvitationsTable.utils';
+import { Invitations } from '../AddUser.types';
+import { mockedAppletId } from '../../../../../../shared/mock';
```
ChaconC: Small suggestion, though linter will complain about import order

```suggestion
import { mockedAppletId } from 'shared/mock';
```

src/modules/Dashboard/features/Participants/Participants.types.ts
```diff
@@ -3,13 +3,16 @@ import { MenuActionProps } from 'shared/components';
...
+import { Invitation } from '../Applet/AddUser/AddUser.types';
+
 export type ParticipantActionProps = {
   respondentId: string | null;
   respondentOrSubjectId: string;
   email: string | null;
   secretId: string | null;
   nickname?: string | null;
   tag?: ParticipantTag | null;
+  invitation: Invitation | null;
```
ChaconC, line 15: Can you mark the multiple `invitation` properties as optional to avoid having to declare `invitation = null` in related files e.g. `Activities.test`, `Participants.utils.test`, etc

src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -30,15 +80,76 @@ export const getParticipantActions = ({
...
+    )}`;
+
+    title += title.length > 0 ? '\n' : '';
+    title += `${invitationDate}`;
+  }
+
+  const hasTitle = title.length > 0;
+
+  return cleanUpDividers([
+    { title, isDisplayed: hasTitle },
```
ChaconC: This item has hover state and `cursor: pointer`, is this expected?

CleanShot 2024-05-31 at 10 04 45

ChaconC (commented):
All changes work great, just a couple minor suggestions.

I did run into an issue where I had a Full (Pending) account where I rejected the invitation, then tried to invite them again and got this message: `Your invitation has already been removed`. Looks like a separate issue, but I can't remove them anymore
CleanShot 2024-05-31 at 09 56 23

Is this something that will be handled by "Cancel invitation" at a later date?


Another thing, noticed Full/Team accounts have a lot of items now making the menu look crowded. The ticket mentions this header is supposed to be for Pending accounts only, but I see the different states in Figma. Flagged for comment 
CleanShot 2024-05-31 at 10 10 45

src/modules/Dashboard/features/Applet/AddUser/InvitationsTable/InvitationsTable.utils.test.ts
sultanofcardio: Thanks. Fixed

src/modules/Dashboard/features/Participants/Participants.types.ts
sultanofcardio: I contemplated doing this (I also added several other properties), but marking them as optional would be incorrect based on the data being returned from the BE. So I decided to go this route, and just set the properties to values that would allow the other tests to pass, since they weren't previously using them anyway

src/modules/Dashboard/features/Participants/Participants.utils.tsx
sultanofcardio: I don't think it's meant to have these, but there's no special header item in the action menu so it just inherits the same behaviour as actual actions. I'll see to removing it

sultanofcardio:
Thanks for the quick review @ChaconC.

> I did run into an issue where I had a Full (Pending) account where I rejected the invitation, then tried to invite them again and got this message: Your invitation has already been removed. Looks like a separate issue, but I can't remove them anymore... Is this something that will be handled by "Cancel invitation" at a later date?

Thanks for catching this. I will restore the remove participant action for pending accounts for now until cancel invitation can be implemented. I was able to successfully invite a user twice after rejecting though

> Another thing, noticed Full/Team accounts have a lot of items now making the menu look crowded. The ticket mentions this header is supposed to be for Pending accounts only, but I see the different states in Figma. Flagged for comment

The copy email address action actually isn't supposed to be there for full accounts, so I'll remove it

ChaconC (approved):
All good now!

src/shared/components/Menu/Menu.tsx
```diff
@@ -68,36 +68,64 @@ export const Menu = <T = unknown,>({
...
+                  disableRipple
+                  sx={{
+                    '&&': {
+                      cursor: 'default',
+                      userSelect: 'text',
+                      ':hover': {
+                        backgroundColor: 'inherit',
+                      },
+                    },
+                  }}
```
JustinNusca: Unfortunately, this isn't totally sufficient for making this a non-selectable/"read-only" element.

`MenuItem` will render a list item with `tabIndex` and `role` attributes that makes this focusable and suggests interaction, and the hover styles we're overriding here will apply when the element is focused (eg, by clicking or via keyboard navigation).

Here's an example of how the macOS screen reader will report this element, for example:
 
Screenshot 2024-05-31 at 11 25 46 AM

I tried to figure out a solution locally by passing a `component` prop that just strips out the relevant props that MenuItem would attempt to render, but doing that seems to mess up the keyboard support and focus management this component has going on.

Ultimately, I think we should just not use the `MenuItem` component at all here, and instead handle this whole case by doing something like:

```jsx
case MenuItemType.Info:
  return (
    <Box data-testid={dataTestId} key={index} sx={{ px: 2, py: 1.2 }}>
      {menuItemContent}
    </Box>
  );
```

This is actually what the MUI-Core docs suggest for handling non-interactive content in a `Menu`, [in this example](https://mui.com/base-ui/react-menu/#wrapping-menu-items)

sultanofcardio: That's a good suggestion, thanks. I started by conditionally passing props to the `MenuItem` before I ultimately created a separate `MenuItemType`, so it makes sense to use a different component here.

src/shared/components/Menu/Menu.tsx
```diff
@@ -68,36 +68,49 @@ export const Menu = <T = unknown,>({
...
+                  data-testid={dataTestId}
+                  disabled={disabled}
+                  onClick={disabled ? undefined : handleMenuItemClick}
+                >
+                  {menuItemContent}
+                </MenuItem>
+              );
+            case MenuItemType.Info:
+              return (
+                <Box data-testid={dataTestId} key={index} sx={{ px: 2, py: 1.2 }}>
```
farmerpaul: 🦅 
Screenshot 2024-06-03 at 4 17 04 PM

```suggestion
                <Box data-testid={dataTestId} key={index} sx={{ px: 1.6, py: 0.8 }}>
```

src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -46,7 +108,71 @@ export const getParticipantActions = ({
...
+    )}`;
+
+    title += title.length > 0 ? '\n' : '';
+    title += `${invitationDate}`;
+  }
+
+  const hasTitle = title.length > 0;
+
+  return cleanUpDividers([
+    { type: MenuItemType.Info, title, isDisplayed: hasTitle },
```
farmerpaul: The logic here could be slightly simplified by pushing each item conditionally onto an array, then doing a `.join('\n')` prior to output:
```tsx
  const title: string[] = [];

  if (emailAddress) {
    title.push(`${i18n.t('email')}: ${emailAddress}`);
  }

  if (hasInvitation && invitation.firstName && invitation.lastName) {
    const fullName = `${i18n.t('fullName')}: ${invitation.firstName} ${invitation.lastName}`;
    title.push(fullName);
  }

  // etc.

  return cleanUpDividers([
    { type: MenuItemType.Info, title: title.join('\n'), isDisplayed: hasTitle },
    …
```

src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -1,26 +1,76 @@
...
-import { ChosenAppletData, GetParticipantActionsProps } from './Participants.types';
+import {
+  ChosenAppletData,
+  GetParticipantActionsProps,
+  ParticipantActionProps,
+} from './Participants.types';
 import { ParticipantsColumnsWidth } from './Participants.const';
 
+/**
+ * Remove dividers that are not between normal displayed menu items
```
farmerpaul, line 29: Am I correct in this suggested description? It seems a bit clearer:
```suggestion
 * Reduce consecutive visible dividers to one
```

src/modules/Dashboard/features/Participants/Participants.utils.tsx
```diff
@@ -1,26 +1,76 @@
...
+
+    return item;
+  });
+
+  if (result.at(lastDisplayedIndex)?.type === MenuItemType.Divider) {
+    result[lastDisplayedIndex].isDisplayed = false;
+  }
+
+  return result;
+}
```
farmerpaul, line 62: This feels like functionality that only pertains to the Participants screen out of circumstance, but actually something we would want applied more generally. Would it make sense to move it into the Menu component itself?

src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -268,12 +273,14 @@ export const Participants = () => {
       status,
       email,
     } = user;
+
+    const detail = details.find((detail) => detail.appletId === appletId) as RespondentDetail;
```
farmerpaul: Eventually I think we're going to replace `/workspace/respondents` with this component, where `appletId` will be undefined. We can always cross that bridge when we get there, but just wanted to note that this type assertion might cause errors in that case. If it's not too awkward, I'd opt to anticipate that possibility now knowing that it's planned.

farmerpaul (approved):
Looks good, only had some minor comments that I don't think are blockers, so happy to approve.

~~I did notice that the menu for pending full accounts has the Remove Participant item:~~

~~[image]~~

~~which is missing in your screenshot. Was that just based on permissions of the logged in user when you took that screenshot?~~ Just saw your comment, I see that this screenshot was from an earlier version of the PR. All good. :+1:

src/modules/Dashboard/features/Participants/Participants.utils.tsx
sultanofcardio: It also hides dividers at the start and end of the list, so I think the existing description better summarises that

src/modules/Dashboard/features/Participants/Participants.utils.tsx
sultanofcardio: I don't think we should generalise this. A more preferable solution would be not including elements in the first place that shouldn't be displayed, but I didn't think that would fit in this PR

src/modules/Dashboard/features/Participants/Participants.tsx
sultanofcardio: Do you mean `/dashboard/respondents`? That sounds like a change best left for when we get there because that would lead to too much refactoring here. For example, when you click on a participant, the navigation to the detail screen uses the subject ID, which is applet-specific 🤷

farmerpaul: Yes, `/dashboard/respondents` (oops). I'm not saying deal with other aspects of this screen right now, like `defaultOnClick`. Just that any changes we make needn't veer us further from the goal of having this screen generalizable. But like you said, easily done when we get there.

sultanofcardio: I think a good middle ground for now would be to fall back to the current behaviour

PR #1774, opened 2024-06-03
feature: Add `activity` & `subject`-specific data exports (M2-6736)

JustinNusca:
### 📝 Description

🔗 [M2-6736](https://mindlogger.atlassian.net/browse/M2-6736): [Export Data] Missing Exports for Activity and Participant data from CTA or menu

This PR implements the ability to filter entries that do not match a given `activity_id`, `activity_flow_id`, `target_subject_id`, or any combination thereof from the data to be exported when clicking an "Export Data" option.

Specifically, I've updated the "Export Data" menu items in the dropdown menus for both Flows and Activities on both the `Applet > Activities` and `Applet > Participant Detail > Activities` screens to add these filters as needed.

Now, when clicking the "Export Data" option on an activity, the resulting data set should be filtered to _only_ include data for the given activity. Additionally, if performing this action from **Participant Details**, the resulting dataset will additionally be filtered to _only_ include data for the given subject.

Finally, this PR also fixes an issue where the `DataExportPopup` would call its submit function multiple times, resulting in multiple identical files being downloaded simultaneously.

> [!Note]
> There's a number of additional changes I'd like to make here, but I prioritized attempting to  implement the necessary functionality in time for deployment tomorrow. These include:
> 1. I think it makes sense to update the `/answers/applet/{applet_id}/data` endpoint to support parameters for the necessary IDs we need to filter by, instead of applying these filters after-the-fact on the frontend. This would amount of data we need to fetch, transform, and then end up discarding. It looks like this endpoint already supports a `targetSubjectIds` parameter, which expects an array of strings, but the response did not change when I included that parameter.
> 2. There's a bunch of repetitive logic to wire up the `DataExportPopup` for activities. I think it would make sense to handle it in the `ActivityGrid` component and then re-use that, much like the `FlowGrid` component does, but this would involve refactoring a lot of unrelated functionality that I'd prefer to leave alone, given the time sensitivity of this task. It also would have introduced a lot of conflicts with https://github.com/ChildMindInstitute/mindlogger-admin/pull/1761, which was not yet merged at the time I was working on this.
> 3. My fix for multiple files being downloaded involves using a ref, which doesn't require waiting for a re-render/state update when checking if the "isExporting" flag is set. I also kept the existing state value, because the rendered UI depends on it. This should work fine, but I think it's worth investigating a more appropriate fix when I have more time, as I think having both `state` and `ref` values that are ostensibly meant to represent the same value is potentially confusing.
> 4. More thorough tests. 🫠

#### Activity Exports

| All Participants | Specific Participant |
|-|-|
| activity | subject-activity |

#### Activity Flow Exports

| All Participants | Specific Participant |
|-|-|
| flow | flow-export |

### 🪤 Peer Testing

Given an applet with a number of activities, and participants that have completed different subsets of those activities…

1. On the **Applet > Activities** screen…
    1. Click a Flow Card's dropdown menu, and choose "Export Data".
    2. After exporting, open the resulting `.csv`, and observe that it only includes entries where the `activity_flow_id` column matches the ID of the Flow you selected.
    3. Click an Activity Card's dropdown menu, and choose "Export Data".
    4. After exporting, open the resulting `.csv`, and observe that it only includes entries where the `activity_id` column matches the ID of the Activity you selected.
1. On the **Applet > Participant Details > Activities** screen…
    1. Click an Flow Card's dropdown menu, and choose "Export Data"
    2. After exporting, open the resulting `.csv`, and observe that it only includes entries where `activity_flow_id` matches the ID of the Flow you selected, and `target_subject_id` matches the ID of the Subject you are viewing.
    3. Click an Activity Card's dropdown menu, and choose "Export Data".
    4. After exporting, open the resulting `.csv`, and observe that it only includes entries where `activity_id` matches the ID of the Activity you selected, and `target_subject_id` matches the ID of the Subject you are viewing.

[M2-6736]: https://mindlogger.atlassian.net/browse/M2-6736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
+++ b/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
@@ -20,7 +20,11 @@ import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 
 import { useTakeNowModal } from '../TakeNowModal/TakeNowModal';
 
-export const useActivityGrid = (dataTestId: string, activitiesData: ActivitiesData | null) => {
+export const useActivityGrid = (
+  dataTestId: string,
+  activitiesData: ActivitiesData | null,
+  onClickExportData?: (activityId: string) => void,
+) => {
   const navigate = useNavigate();
   const { appletId } = useParams();
   const workspaceRoles = workspaces.useRolesData();
@@ -53,10 +57,9 @@ export const useActivityGrid = (dataTestId: string, activitiesData: ActivitiesDa
         );
       },
       exportData: ({ context }: MenuActionProps<ActivityActionProps>) => {
-        const { activityId } = context || {};
-        // TODO: Implement export data
-        // https://mindlogger.atlassian.net/browse/M2-6039
-        alert(`TODO: Export data (${activityId})`);
+        if (context?.activityId) {
+          onClickExportData?.(context?.activityId);
+        }
       },
       assignActivity: ({ context }: MenuActionProps<ActivityActionProps>) => {
         const { activityId } = context || {};
--- a/src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
+++ b/src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
@@ -24,10 +24,12 @@ export function useFlowGridMenu({
   hasParticipants = false,
   testId = '',
   subject,
+  onClickExportData,
 }: {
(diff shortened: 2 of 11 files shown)


farmerpaul (approved):
Tested and reviewed, all looks good! ✔️

sultanofcardio (approved):

PR #1777, opened 2024-06-04
fix: Pin button fixes (M2-6843)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-6843](https://mindlogger.atlassian.net/browse/M2-6843)

This PR improves the Pin button's residual keyboard focus state when pinning/unpinning rows in the Participants and dashboard Respondents screens.

When pinning or unpinning a row, it's desirable for the pressed pin button to continue to have keyboard focus – and thus have a grey background. However, previously, the wrong row had its pin button remain focused due to array index being used to identify row components (which is bad practice in React). This PR fixes that so that the same pressed Pin button continues to have keyboard focus after refresh. This and other improvements include:

- Adding support for `keyExtractor`s to `Table` and `DashboardTable` components, so that a unique identifier (besides array index) can be used to identify React components by `key`
- Improving styling of `Pin` component to better clarify its state when focused
- Fixing a newly discovered sorting bug whereby the refresh that happens after pinning temporarily forgets which is the selected column being sorted by; solved it in the `useTable` hook using a ref (see [explanation in last commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1777/commits/cb68a1b9fbc836214c00a634e1dd60e01a290a20)).

> [!TIP]
> It will be easier to review the code with [whitespace ignored](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1777/files?diff=unified&w=1).

### 📸 Screenshots

**Pinning/unpinning using the mouse as well as keyboard navigation:**

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/15707370-386d-45f4-bb85-bf8fb12b6f6d

### 🪤 Peer Testing

- On an applet's **Participants** screen or the dashboard **Respondents** screen, **pin or unpin** a row.
    **Expected outcome:** The pin button for the same row that was just pinned/unpinned preserves keyboard focus.
- Sort by **Nickname** column, then **pin or unpin** a row.
    **Expected outcome:** The rows continue to be sorted by Nickname within the respective pinned/unpinned sections.


src/shared/hooks/useTable.ts
```diff
@@ -28,15 +28,24 @@ export const useTable = (
...
+      ...(ordering && { ordering }),
+    }),
+    [ownerId, limit, searchValue, page, ordering],
+  );
+
+  // Ref required to eliminate closure reference errors and reduce redundant rerenders
+  const paramsRef = useRef(params);
+  useEffect(() => {
+    paramsRef.current = params;
+  }, [params]);
```
sultanofcardio, line 46: I don't have a preference either way, but I'm curious if you considered wrapping the various `handleX` functions in `useCallback` hooks which could probably achieve the same thing? The outcome would be the same, I'm just curious about your thought process in choosing a ref instead (or maybe it's not that deep 🤷)

sultanofcardio (approved):

src/shared/hooks/useTable.ts
sultanofcardio: On second thought, this probably wouldn't solve the closure reference part of the issue so nvm

farmerpaul: Yeah, I had the same line of thinking, and it's what I tried at first. I was dismayed that it didn't make any difference. I still don't understand how in this case some variable other than the current value of the state variable is getting referenced from closures. I pestered Copilot for a solution and this was the one it suggested that finally worked!

PR #1778, opened 2024-06-05
fix: Reviewer's Assessment Activity must not be displayed in the Activity Flow Builder (M2-6852)

IhorNesterenko99:

### 📝 Description

🔗 [Jira Ticket M2-6852](https://mindlogger.atlassian.net/browse/M2-6852)

Changes include:

- Clearing the Activities list of isReviewable Activities in the Activity Flow Builder.
- Clearing the Activities list of isReviewable Activities while creating the Activity Flow.
- Removing isReviewable Activities from Activity Flows when the isReviewable checkbox is checked.
- Removing isReviewable Activities from previously saved Activity Flows if they were saved with isReviewable Activities.
- Added unit tests.

### DIFF
--- a/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
+++ b/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
@@ -1,3 +1,4 @@
+import { ChangeEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Box } from '@mui/material';
 
@@ -18,14 +19,15 @@ import {
   MAX_NAME_LENGTH,
   TEXTAREA_ROWS_COUNT_SM,
 } from 'shared/consts';
-import { byteFormatter } from 'shared/utils';
+import { byteFormatter, getEntityKey } from 'shared/utils';
 import { BuilderContainer } from 'shared/features';
-import { ItemFormValues } from 'modules/Builder/types';
+import { ActivityFlowFormValues, ItemFormValues } from 'modules/Builder/types';
 import {
   useRedirectIfNoMatchedActivity,
   useCurrentActivity,
   useCustomFormContext,
 } from 'modules/Builder/hooks';
+import { getUpdatedActivityFlows } from 'modules/Builder/utils';
 
 import { Uploads } from '../../components';
 import { StyledContainer } from './ActivityAbout.styles';
@@ -41,11 +43,12 @@ export const ActivityAbout = () => {
   useRedirectIfNoMatchedActivity();
 
   const { control, setValue, watch } = useCustomFormContext();
-  const { fieldName } = useCurrentActivity();
+  const { fieldName, activity } = useCurrentActivity();
   const hasVariableAmongItems = useCheckIfItemsHaveVariables();
   const hasRequiredItems = useCheckIfItemsHaveRequiredItems();
 
   const activityItems = watch(`${fieldName}.items`);
+  const activityFlows: ActivityFlowFormValues[] = watch('activityFlows');
   const hasUnsupportedReviewableItemTypes = activityItems?.some(
     (item: ItemFormValues) =>
       ![...itemsForReviewableActivity, ''].includes(item.responseType as ItemResponseType),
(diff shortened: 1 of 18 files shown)


alex-i-e:
Could you please clarify what activities were in Flow_1 (combine reports) before change Assessment Activity checkbox?

After change that checkbox in that flow we got Activity_1 and Activity_1 which looks strange:
image

IhorNesterenko99:
@alex-i-e 
We had Activity__2 in the Flow, which was removed after checking the checkbox. And yes, it can be strange because there's no limit on how many times the same Activity can be added to the same Flow.

src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
```diff
@@ -95,6 +98,15 @@ export const ActivityAbout = () => {
     },
   ];
 
+  const handleIsReviewableChange = (event: ChangeEvent<HTMLInputElement>) => {
+    if (!event.target.checked) return;
+
+    const activityKey = getEntityKey(activity || {});
```
alex-i-e: maybe it worth to add early return before it:
```
if (!activity) return; // or `setValue('activityFlows', activityFlows);` - if needed
```
```suggestion
    const activityKey = getEntityKey(activity);
```

src/modules/Builder/utils/getUpdatedActivityFlows.ts
```diff
@@ -0,0 +1,18 @@
+import { ActivityFlowFormValues } from 'modules/Builder/types';
+
+export const getUpdatedActivityFlows = (
+  activityFlows: ActivityFlowFormValues[],
+  keyToRemove: string,
+) =>
+  activityFlows.reduce((acc: ActivityFlowFormValues[], flow) => {
```
alex-i-e, line 7: could you a bit simplified the code to keep the reference as it was:
`keyToRemove ? (...) : activityFlows`

src/modules/Builder/features/ActivityFlowBuilder/ActivityFlowBuilder.tsx
```diff
@@ -52,7 +52,12 @@ export const ActivityFlowBuilder = () => {
     name: `${activityFlowName}.items`,
     keyName: REACT_HOOK_FORM_KEY_NAME,
   });
-  const activities: AppletFormValues['activities'] = watch('activities');
+  const formActivities: AppletFormValues['activities'] = watch('activities');
```
alex-i-e: here and in other places, please consider to use [useWatch](https://react-hook-form.com/docs/usewatch) :
```suggestion
  const formActivities: AppletFormValues['activities'] = useWatch({ name: 'activities'});
```
watch - This API will trigger re-render at the root of your app or form, consider using a callback or the useWatch api if you are experiencing performance issues.

src/modules/Builder/features/ActivityFlowBuilder/ActivityFlowBuilder.test.tsx
```diff
@@ -202,4 +216,33 @@ describe('Activity Flow Builder', () => {
...
   });
+
+  test('Ensures no reviewable activities remain when adding/replacing an activity)', async () => {
+    renderActivityFlowBuilder(mockedAppletFormDataWithReviewableActivity);
+
+    const addActivity = screen.getByTestId(`${mockedActivityFlowBuilderTestid}-add`);
+    await userEvent.click(addActivity);
+
+    const activitiesInAddActivity = screen.getAllByTestId(
+      new RegExp(`^${mockedActivityFlowBuilderTestid}-add-activity-\\d+$`),
```
alex-i-e: Let's move `new RegExp(`^${mockedActivityFlowBuilderTestid}-add-activity-\\d+$`)` into a constant here and in row 243 to reduce new object creation in the loops.

src/modules/Builder/utils/getUpdatedActivityFlows.ts
IhorNesterenko99: Could you please provide more details?

alex-i-e (approved):

PR #1784, opened 2024-06-06
fix: Export participant-specific data from Participant Details "Export Data" button (M2-6743)

JustinNusca:
### 📝 Description

🔗 [M2-6743](https://mindlogger.atlassian.net/browse/M2-6743): [Export Data] Data is exported for all participants from single participant view

This PR fixes an issue where the "Export Data" button in the header of the Participant Details screens exported data for the entire applet, rather than the specific subject represented by the Participant Details screen.

It uses the same [updated prop interface (ie, `filters`) I added in my previous PR](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1774), and allows passing these filters from the `HeaderOptions` component, where they will be passed to the ExportDataPopup. When modifying the prop interface for HeaderOptions, I went ahead and additionally allowed for a full spread of `BoxProps` as well, in case we ever need to modify styles for that component on specific screens.

I also made one small tweak to the existing behaviour for filters as part of this change — I'm passing the `targetSubjectId` filter value as the `targetSubjectIds` parameter for the relevant API endpoint. This will be used whenever the subject ID cannot be derived from the given applet data, or when the `isAppletSetting` prop is set. This should help reduce the overall size of requested data for export, which should in turn reduce the number of checks for the other filter settings.

### 📸 Screenshots

data-export-details

### 🪤 Peer Testing

Given an applet with a number of activities, and participants that have completed different subsets of those activities…

1. On the Applet > Participant Details > Activities screen…
    1. Click "Export" in the header area on any tab.
    2. After exporting, open the resulting .csv, and observe that it only includes entries where the `target_subject_id` matches the ID of the Subject you are viewing.

It may also be worthwhile to try some of the other data export options, and ensure that there haven't been any regressions.

[M2-6743]: https://mindlogger.atlassian.net/browse/M2-6743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


farmerpaul:
@JustinNusca I only just noticed this one at the bottom of the list of open PRs. The ticket associated with this PR was verified as fixed by QA over two weeks ago, apparently covered by M2-6736; can this one be closed?

farmerpaul:
I think I see the disconnect now – this ticket is specifically about the Export Data button in the _header_, while the associated ticket doesn't mention the header, and isn't what QA tested when they confirmed it was fixed. I'll give this a quick run-through!

farmerpaul (approved):
Tested, looks right to me!

Since the associated ticket doesn't describe this specific user flow, does this need a new ticket for QA to confirm?

farmerpaul:
Also, it looks like this needs to get rebased/targeted to `develop`. Let me know if you need anything you need to help get this unblocked.

JustinNusca:
Thanks for checking in on this! This definitely fell out of notice under the amount of other PRs that have been going on. The linked task was a clone of a related task (data exports in the context menus), and I suspect that the description was never updated appropriately — I remember the actual change this PR makes being gone over specifically in the estimation session alongside the related, and I suppose I intuited the difference between the two tasks without pausing to update the written description in Jira.

I'll file a new task to better track this and rebase this branch shortly, but I wanted to give a holler beforehand that this hadn't escaped my notice just yet.

PR #1787, opened 2024-06-07
feat: DataViz: Activity Flows on the 'Summary' tab - improve performance when a lot of items with charts (M2-6674)

IhorNesterenko99:

### 📝 Description

🔗 [Jira Ticket M2-6674](https://mindlogger.atlassian.net/browse/M2-6674)

Changes include:

- added missing skeletons for all charts
- added flowResponseOptionsCount form value to determine if activate static for Flow
- replaced watch method with useWatch hook
- minor TS types adjustments
- unit tests fixes

### 📸 Screenshots

 | After                                 |
 | ------------------------------------- |
 | https://app.screencast.com/AAYOIclW8Ugo7 |

### DIFF
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/hooks/useRespondentAnswers/useRespondentAnswers.ts
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/hooks/useRespondentAnswers/useRespondentAnswers.ts
@@ -21,11 +21,21 @@ import { getObjectFromList } from 'shared/utils/getObjectFromList';
 import { getFormattedResponses } from '../../utils/getFormattedResponses';
 import { FetchAnswers } from '../../RespondentDataSummary.types';
 import { getDateISO, getIdentifiers, processIdentifiersChange } from './useRespondentAnswers.utils';
+import { useDataSummaryContext } from '../../DataSummaryContext';
 
 export const useRespondentAnswers = () => {
   const { appletId, respondentId } = useParams();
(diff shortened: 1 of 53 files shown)


src/modules/Dashboard/features/RespondentData/RespondentDataSummary/hooks/useRespondentAnswers/useRespondentAnswers.ts
```diff
@@ -204,6 +207,7 @@ export const useRespondentAnswers = () => {
 
         setValue('flowSubmissions', flowSubmissions);
         setValue('flowResponses', flowResponses);
+        setValue('flowResponseOptionsCount', responseOptionsCount);
```
alex-i-e: As I see this is an incorrect use of the `react-hook-form`, not for its intended purpose: for each new field the library will create the bunch of attributes and handlers. That means each time user change the real input it will do the redundant checks with such fields and will side effect the real state of the form.

As I see we can use own context here.

alex-i-e: UPD: please check the same behaviour for `flowSubmissions`, `flowResponses` at least

alex-i-e: Perfect!

alex-i-e (approved):
Thank you for important updates! LGTM!

PR #1790, opened 2024-06-11
feature: Add warning before editing default schedule from within Participant Details (M2-6775)

JustinNusca:
### 📝 Description

🔗 [M2-6775](https://mindlogger.atlassian.net/browse/M2-6775): [Admin Panel] It is possible to update Default schedule from the particular participant's schedule page

This PR updates the Schedule pages to modify the behaviour when editing a schedule.

Previously, when viewing the Schedule tab on the Participant Details screen, for a participant on the default schedule, making changes to the default schedule would result in changes to the default schedule.

This is/was the expected functionality, but it may be confusing to users who expect that, because they are on a screen related to a specific participant, that changes to the schedule would apply _only_ to that participant.

To help remove this misunderstanding, we now show a "warning" dialog whenever the user presses an interactive element that may result in a schedule update. More specifically, whenever the user presses "Clear Schedule", "Create Event", or "Edit Event" while viewing the Participant Details screen's Schedule tab for a user with the default schedule, a popup explaining that changes will be made to the default schedule is now shown. The user is given the option to continue (which will open the final resulting popup), or to create an individual schedule first, if possible.

To help accomplish this, I've moved a fair amount of logic to a new context-provider component, `ScheduleProvider`, and an accompanying hook, `useSchedule`. The providers component manages the state of the various popups, receives props for configuring behaviour, which may be passed along to nested children via `useSchedule`, and provides event handler functions for passing to buttons for opening the various popups. This change allows us to move some duplicated logic for these popups split across the `Calendar` and `Legend` components, and manage them from a single location.

### 📸 Screenshots

| Editing Default Schedule | Creating Individual Schedule |
|-|-|
| edit-default-schedule | edit-individual-schedule |

### 🪤 Peer Testing

The testing flow is similar to [previous ones](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1760) I've opened for Schedule-related tasks and fixes, and involve verifying that the creation/editing/clearing of events functions as expected in specific scenarios:

From the **Participant Details > Schedule** page for a *full* participant on the *default* schedule…
1. Press the "Create Event" button, and notice that a new popup has opened, notifying the user that changes will be made to the default schedule. Press "Update the default schedule".
2. Proceed with creating the event. Notice that the event has been created and applied to the default schedule.
4. Press the Event in the calendar, and notice that a new popup has opened, notifying the user that changes will be made to the default schedule. Press "Update the default schedule".
5. Proceed with making a change to the event (eg, to the date). Notice that the event has been updated and applied to the default schedule.
6. Press the "Clear Scheduled Events" button, and notice that a new popup has opened, notifying the user that changes will be made to the default schedule. Press "Update the default schedule".
7. Proceed with clearing scheduled events, and notice that scheduled events have been removed from the default schedule.

From the **Participant Details > Schedule** page for a *full* participant on an *individual* schedule…
1. Repeat the process above, but notice that in steps 1, 4, and 6, **no** new popup is created, and that you may proceed with changes that are applied to the individual schedule.

From the **Applet > Schedule** page:
1. Repeat the process above, but notice that in steps 1, 4, and 6, **no** new popup is created, and that you may proceed with changes that are applied to the default schedule.

From the **Participant Details > Schedule** page for a *full* participant on the *default* schedule…
1. Press the "Create Event" button, and notice that a new popup has opened, notifying the user that changes will be made to the default schedule. Press "Create an Individual Schedule".
2. Notice that the Schedule tab has been refreshed and the user is now on an individual schedule.
3. Proceed with Creating an event, editing an event, and clearing scheduled events on this individual schedule.
4. Notice that **no** new popup is shown prior to making updaes, and that updates to the schedule apply to the individual schedule.

[M2-6775]: https://mindlogger.atlassian.net/browse/M2-6775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/features/Applet/Schedule/ScheduleProvider/ScheduleProvider.tsx
sultanofcardio, line 1: I'm not sure what exactly causes this, but it seems we're no longer able to remove scheduled events. It has something to do with the introduction of this `ScheduleProvider`, I'm just not sure what.


https://github.com/ChildMindInstitute/mindlogger-admin/assets/14842108/3f5a35d1-04a9-4536-80e0-f0645ec243a6

Inside `EditEventPopup`, moving `handleEditEventClose();` from `onRemoveEventClick` to `handleRemoveEvent` seems to fix it

https://github.com/ChildMindInstitute/mindlogger-admin/blob/ce94338ee38b8bbef027662b1d3ff35fa0407123/src/modules/Dashboard/features/Applet/Schedule/EditEventPopup/EditEventPopup.tsx#L66-L84

sultanofcardio (commented):
Everything else seems to work as prescribed. Just one tiny regression

src/modules/Dashboard/features/Applet/Schedule/ScheduleProvider/ScheduleProvider.tsx
JustinNusca: Good catch!

The suggested change here works, but I think it results in a different UX than we previously had: it results in multiple dialogs placed on top of each other, rather than opening/closing each in succession.

I spent some more time trying to understand the issue, and it has to do with the `open` prop and the fact that `EditEventPopup` is in fact a collection of multiple dialogs. The `open` prop actually only controls the open state of the "default" dialog, and the visibility of other dialogs is controlled by internal state values.

I naively assumed that the `open` prop was essentially "none of the dialogs contained herein should be visible", and treated it as such. This included resetting the state value of the `selectedEvent` in question in my `ScheduleProvider` on close, which resulted in the component unmounting entirely.

The way it was _actually_ expected to work here is: keep the component mounted, but pass `open={false}` and `editedEvent={selectedEvent}` so that the _other_ dialogs in the component may be seen and function.

That's really non-obvious to me, so I was tempted to do more refactoring to make the behaviour more explicit, but in the interest of expediency, I opted to just match the previously existing setup. You can see the commit that addresses this, [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1790/commits/e3fda9975dd08fd7d406e56117a66e676cacdcd8).

One detail that I _did_ change though, is that the "Remove Event" button uses a standard button variant and no longer overlaps the "Close" button:

remove-event

sultanofcardio: Thanks for spending some time digging into the mechanics of the various dialogs. I can't access my laptop to test this until the power comes back but based on your explanation, I expect this should work now so I'm gonna approve 👌

sultanofcardio: Also, thanks for fixing the button overlap. I kinda mentally tuned that out for some reason, but great to have it resolved

sultanofcardio (approved):

PR #1802, opened 2024-06-13
fix: Remove pagination from Overview table (M2-6555)

JustinNusca:
### 📝 Description

🔗 [M2-6555](https://mindlogger.atlassian.net/browse/M2-6555): [FE][Overview] Applet Overview Page

This PR removes the pagination controls from the Recent Submissions table on the **Applet → Overview** screen.

Our existing table components do not support _un-paginated_ tables, and the table controls could not be previously opted out of, so I updated the `DashboardTable` to add an `enablePagination` prop which defaults to true, and updated the `DashboardTable`'s types accordingly, so that previously-required pagination-related props should be omitted when `enablePagination` is set to `false`.

### 📸 Screenshots

localhost_3000_dashboard_c81bee80-1da8-479f-9237-212a850e0c58_overview

### 🪤 Peer Testing

Have an applet, ideally with more than 25 submissions against various activities.

1. Navigate to the **Applet → Overview** screen for the applet.
2. Observe that the recent submissions table does not have any pagination controls, and older submissions cannot be viewed.

[M2-6555]: https://mindlogger.atlassian.net/browse/M2-6555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
+++ b/src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
@@ -8,39 +8,40 @@ import { StyledEllipsisText } from 'shared/styles';
 import { DashboardTableProps } from './DashboardTable.types';
 import { StyledTableCell, StyledTableHead } from './DashboardTable.styles';
 
-export const DashboardTable = ({
-  columns,
-  rows,
-  keyExtractor = (_row: Row, index: number) => `row-${index}`,
-  maxHeight = '100%',
-  uiType = UiType.Primary,
-  className = '',
-  order,
-  orderBy,
-  emptyComponent,
-  page,
-  handleRequestSort,
-  handleChangePage,
-  count,
-  hasColFixedWidth,
-  rowsPerPage = DEFAULT_ROWS_PER_PAGE,
-  onScroll,
-  'data-testid': dataTestid,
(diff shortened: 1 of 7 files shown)


src/modules/Dashboard/components/DashboardTable/DashboardTable.tsx
```diff
@@ -8,39 +8,40 @@ import { StyledEllipsisText } from 'shared/styles';
...
-        rowsPerPage={rowsPerPage}
-        page={page - 1}
-        onPageChange={handleChangePage}
-        labelRowsPerPage=""
-        rowsPerPageOptions={[]}
-        data-testid={`${dataTestid}-table-pagination`}
-      />
-    </StyledTableCellContent>
-  );
+export const DashboardTable = (props: DashboardTableProps) => {
```
JustinNusca, line 11: Since our `props` are a discriminated union, I couldn't de-structure it directly in the parameters here, because I needed to access `props.enablePagination` as written a few lines later. Not doing so would cause our type-checker to lose track of which of the union's component types we're trying to narrow down to, and throw errors.

sultanofcardio (approved):

PR #1803, opened 2024-06-13
feat(mi): Include full account participants in take now (M2-6533)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-6533](https://mindlogger.atlassian.net/browse/M2-6533)

This PR updates the behaviour of the take now modal in the following ways:
- Full account participants, when selected in the "Who will be providing the responses" dropdown, no longer trigger a warning message below the field. The warning is still triggered for those who are pending invitation.
- Full account participants may now self-report. This means that the “This person will be inputting their own responses” checkbox is not disabled, and the user is no longer forced to select a team member in the "Who will be inputting the responses" dropdown.
- Full account participants may now be selected as an option in the "Who will be inputting the responses" dropdown. This field was previously limited to team members only

These behaviours are controlled by the `enableParticipantMultiInformant` feature flag, and revert to their previous behaviour if the flag is turned off.

#### Before

A full account participant will trigger a warning message and force the selection of a team member to input the responses

image

#### After

A full account participant may optionally self-report

image

They may also report on behalf of someone else

image

### 🪤 Peer Testing

You may manually toggle the value of the `enableParticipantMultiInformant` feature flag while testing by modifying this function

https://github.com/ChildMindInstitute/mindlogger-admin/blob/60bb60127f708a6c8ea01ecade05dae91a1ae459/src/shared/hooks/useFeatureFlags.ts#L11-L17

```typescript
const featureFlags = () => {
  const keys = Object.keys(FeatureFlagsKeys) as (keyof typeof FeatureFlagsKeys)[];
  const features: FeatureFlags = {};
  keys.forEach((key) => (features[key] = flags[FeatureFlagsKeys[key]]));

  // TODO: remove after testing
  features.enableParticipantMultiInformant = true;

  return features;
};
```

Activate the take now modal and test the conditions mentioned in the description above. Don't proceed to the web app, as it has not yet been updated to facilitate take now by a participant.

### ✏️ Notes

Tests have not yet been added to contrast behaviours depending on the new feature flag

### DIFF
--- a/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
+++ b/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
@@ -13,6 +13,7 @@ import {
   variables,
 } from 'shared/styles';
 import { ParticipantSnippet } from 'modules/Dashboard/components/ParticipantSnippet';
+import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 
 import { LabeledUserDropdownProps, ParticipantDropdownOption } from './LabeledUserDropdown.types';
 import { StyledGroupLabel, StyledWarningMessageContainer } from './LabeledUserDropdown.styles';
@@ -37,6 +38,9 @@ export const LabeledUserDropdown = ({
   const debounceRef = useRef<NodeJS.Timeout | null>(null);
   const [combinedOptions, setCombinedOptions] = useState<ParticipantDropdownOption[]>(options);
   const [isSearching, setIsSearching] = useState(false);
+  const {
+    featureFlags: { enableParticipantMultiInformant },
+  } = useFeatureFlags();
 
   const debouncedSearchHandler = useCallback(
     (search: string) => {
@@ -69,7 +73,9 @@ export const LabeledUserDropdown = ({
     [debounce, handleSearch],
   );
 
-  const shouldShowWarningMessage = !!canShowWarningMessage && !!value && !value.userId;
+  const warningMessageValueCondition =
+    !!value && (enableParticipantMultiInformant ? !value.userId : value.tag !== 'Team');
+  const shouldShowWarningMessage = !!canShowWarningMessage && warningMessageValueCondition;
 
   let groupBy: LabeledUserDropdownProps['groupBy'];
 
--- a/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
+++ b/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
@@ -69,7 +69,7 @@ export const LabeledUserDropdown = ({
     [debounce, handleSearch],
   );
 
-  const shouldShowWarningMessage = !!canShowWarningMessage && !!value && value.tag !== 'Team';
+  const shouldShowWarningMessage = !!canShowWarningMessage && !!value && !value.userId;
 
   let groupBy: LabeledUserDropdownProps['groupBy'];
 
--- a/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
+++ b/src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
@@ -34,7 +34,7 @@ export const LabeledUserDropdown = ({
   showGroups,
   ...rest
 }: LabeledUserDropdownProps) => {
-  const { t } = useTranslation('app');
+  const { t, i18n } = useTranslation('app');
   const debounceRef = useRef<NodeJS.Timeout | null>(null);
   const [combinedOptions, setCombinedOptions] = useState<ParticipantDropdownOption[]>(options);
   const [isSearching, setIsSearching] = useState(false);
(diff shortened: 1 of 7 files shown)


src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -211,13 +227,25 @@ export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
...
       async (
         query: string,
-        types: [SearchType, ...SearchType[]],
+        types:
+          | [AnyTeamSearchType, ...AnyTeamSearchType[]]
+          | [FullTeamSearchType, ...FullTeamSearchType[]],
       ): Promise<ParticipantDropdownOption[]> => {
         const requests = [];
 
+        // @ts-expect-error Yes, the array can in fact contain this value
```
JustinNusca, line 243: I'm not sure what the `includes` definition is doing to mangle the union type like this, but we _could_ make our own assertions, like so:

```ts

const isAnyParticipant = (types as (typeof types)[number][]).includes('any-participant');

const isFullParticipant = (types as (typeof types)[number][]).includes('full-participant');
```

... which is admittedly clunky-looking, but might just be preferable to squelching errors? I'll leave it to you to decide if you want to act on this one.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -69,7 +73,9 @@ export const LabeledUserDropdown = ({
     [debounce, handleSearch],
   );
 
-  const shouldShowWarningMessage = !!canShowWarningMessage && !!value && value.tag !== 'Team';
+  const warningMessageValueCondition =
+    !!value && (enableParticipantMultiInformant ? !value.userId : value.tag !== 'Team');
```
farmerpaul: Might be good to separate out some of these expressions, as the logic isn't super obvious to me:
```tsx
  const hasAccount = !!value?.userId;
  const isTeamMember = value?.tag === 'Team';
  const warningMessageValueCondition = enableParticipantMultiInformant
    ? !hasAccount
    : !isTeamMember;
```

JustinNusca (commented):
This looks pretty good, but I'm noticing an issue with the options provided to the "Who will be inputting the responses?" dropdown, specifically.

It looks like the logged-in user is duplicated as a possible value, across two different "Team Members" sub-sections, like so:



Haven't quite worked out where the duplicate entry is coming from yet, though.

farmerpaul (commented):
Looks good, mostly! I just encountered the same thing that @JustinNusca mentioned, and I also haven't investigated as to why – hopefully it's some obvious thing!

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/dbf5f804-a503-427f-be4d-48e2c885bec6

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -126,6 +137,11 @@ export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
     [participantsOnly, teamMembersOnly],
   );
 
+  const fullAccountParticipantsAndTeamMembers = useMemo(
+    () => [...teamMembersOnly, ...fullAccountParticipantsOnly],
+    [fullAccountParticipantsOnly, teamMembersOnly],
+  );
```
JustinNusca, line 143: Related to the earlier comments about the "Who will be inputting the responses?" dropdown that Paul and I left — I think we should be filtering out duplicates that may appear in both `teamMembersOnly` and `fullAccountParticipantsOnly` arrays, here.

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: I'm not so sure either. For me `typeof types` initially resolves to `any`, so `(typeof types)[number][]` is just `any[]`. I'm good with leaving this as-is for now

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
sultanofcardio: Apparently I forgot to filter out team members from the `fullAccountParticipantsOnly` array, so I made that update.

cc @farmerpaul

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Thanks for the suggestion 👍

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
```diff
@@ -187,7 +202,7 @@ export const LabeledUserDropdown = ({
         {...rest}
       />
       {shouldShowWarningMessage && (
-        <StyledWarningMessageContainer>
+        <StyledWarningMessageContainer data-testid={`${rest['data-testid']}-warning-message`}>
```
farmerpaul, line 210: Just caught this issue while testing. It's not related to your changes, but I thought it worth fixing. It's actually related to lines 180-181 above (I can't attach this comment to them directly, so picking the next closest line):

```tsx
          return `${value.secretId}${value.nickname ? ` (${value.nickname})` : ''}${
            value.tag ? ` (${value.tag})` : ''
```

Tags should be getting passed through i18n – see how it's done in [`ParticipantTagChip.tsx`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/M2-6533-take-now-full-accounts/src/modules/Dashboard/components/ParticipantTagChip/ParticipantTagChip.tsx#L21).

farmerpaul (approved):
Thanks for the fix, works in testing! Just identified a minor bug that I thought you could squeeze in a fix for here.

src/modules/Dashboard/components/TakeNowModal/LabeledDropdown/LabeledUserDropdown.tsx
sultanofcardio: Done 👌

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
JustinNusca: Hm, that's very odd, since we explicitly define the type for the `types` parameter a couple lines up, so I have no idea why `typeof types` would be `any` 🤔.

FWIW, this worked locally and provided intellisense autocompletion for the possible `.includes` values. Not a big deal, though.

PR #1814, opened 2024-06-18
feature: Add question text tooltips to Item Flow condition dropdowns (M2-6908)

JustinNusca:
### 📝 Description

🔗 [M2-6908](https://mindlogger.atlassian.net/browse/M2-6908): [Admin] - Add Tooltip to Item Flow section

This PR updates the Item Flow section of the Activity builder so that dropdowns in the Item flow conditions now show tooltips, which display the question/"Displayed Content" of the selected item in the dropdown.

To do this, I've added a new `ItemFlowSelectController` component, which is a styled variant of the `SelectController`, and wrapped in a `Tooltip` that expects a markdown-formatted string as content, for use in the Condition & SummaryRow components.

As part of this, I've added a styled variant of the `MDPreview` component from our [`md-editor-rt`](https://github.com/imzbf/md-editor-rt) dependency, which removes much of the "themed" styles it otherwise provides.

Finally, I've also updated the `Tooltip` component so that it is styled with a calculated `maxHeight` property that prevents tooltips from extending off of the vertical edges of the viewport. As part of this change, I added a new dependency: `popper-max-size-modifier`, to calculate the max height of the Tooltip. 

This also meant adding an explicit `@popperjs/core` dependency to the project. This was already a transitive dependency via `@mui/core`, which uses Popperjs to implement it's Tooltips already, but I wanted to include specific type annotations that referenced definitions in that package, and wanted to avoid referencing a dependency that was only implicitly included.

Finally, there is a small design update that changes the dropdown inputs in the Condition row, to match the dropdown inputs in the Summary row. This is how this UI is depicted in the designs, and there were existing styles that were intended to implement this, but incorrect selectors previously prevented them from being applied.

### 📸 Screenshots

| Before | After |
|-|-|
| before | after |

### 🪤 Peer Testing

For an applet with an activity that contains various items…

1. Navigate to the "Edit Activity" screen, and then to the "Item Flow" tab.
2. Create a new condition, if necessary. Interact with the item dropdowns and observe that they show tooltips containing the rendered markdown content entered in the Item configuration's "Displayed Content" input.

### ✏️ Notes

- One detail I noticed is that `sup`, `sub`, and `blockquote` elements don't seem to be parsed correctly by MDPreview, and the markdown tags for them just render as plain-text characters. This seems like a pre-existing issue with the third-party component, so I opted not to address this within the scope of the task.

[M2-6908]: https://mindlogger.atlassian.net/browse/M2-6908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Builder/features/ActivityItemsFlow/ItemFlow/SummaryRow/SummaryRow.tsx
+++ b/src/modules/Builder/features/ActivityItemsFlow/ItemFlow/SummaryRow/SummaryRow.tsx
@@ -5,11 +5,9 @@ import { useWatch } from 'react-hook-form';
 import { StyledTitleMedium } from 'shared/styles';
 import { getEntityKey } from 'shared/utils';
 import { SelectEvent } from 'shared/types';
+import { ItemFlowSelectController } from 'modules/Builder/components/ItemFlowSelectController/ItemFlowSelectController';
 import { ItemFormValues } from 'modules/Builder/types';
-import {
-  StyledSummaryRow,
-  StyledSummarySelectController,
-} from 'shared/styles/styledComponents/ConditionalSummary';
+import { StyledSummaryRow } from 'shared/styles/styledComponents/ConditionalSummary';
 import { useCustomFormContext } from 'modules/Builder/hooks';
 
 import { SummaryRowProps } from './SummaryRow.types';
@@ -18,7 +16,7 @@ import { useItemsInUsage } from './SummaryRow.hooks';
 
 export const SummaryRow = ({ name, activityName, 'data-testid': dataTestid }: SummaryRowProps) => {
   const { t } = useTranslation('app');
-  const { control, setValue } = useCustomFormContext();
+  const { control, setValue, getValues } = useCustomFormContext();
   const items = useWatch({ name: `${activityName}.items` });
   const itemsInUsage = useItemsInUsage(name);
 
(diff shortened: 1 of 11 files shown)


src/modules/Builder/features/ActivityItemsFlow/ItemFlow/SummaryRow/SummaryRow.tsx
```diff
@@ -34,20 +33,26 @@ export const SummaryRow = ({ name, activityName, 'data-testid': dataTestid }: Su
     [items, activityName, setValue],
   );
 
+  const { question } =
+    ((items ?? []) as Item[]).find(({ id }) => id === getValues(`${name}.itemKey`)) ?? {};
```
ChaconC: If you use `ItemFormValues` instead of `Item` we can avoid the `unknown as string` cast below

ChaconC (approved):
Tested a few items and the tooltip is displayed correctly, great work
Minor suggestion but non-blocking

mbanting (changes requested):
@JustinNusca I didn't review the PR, but leaving a comment to please merge to `develop` instead of the `feature/multiinformant-metapod` branch. Also noted on the ticket. Thanks!

farmerpaul (approved):
Looks good! Approving pending @mbanting's request.

mbanting (approved):
Nice work @JustinNusca !

PR #1815, opened 2024-06-18
feat(mi): Pending applet team member details (M2-6720)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-6720](https://mindlogger.atlassian.net/browse/M2-6720)

This PR updates the Applet team tab with more contextual information about pending team members. The following changes were made:

#### Table Row

- The role chip colour for pending team members in the table is now `warning` instead of `secondary`
- The word "Pending" (localised) is appended in parentheses to the role for pending team members

image

#### Context Menu

The row context menu for pending team members has been updated with the following additional items
- Invitation date
- Copy email address action
- Copy invitation link action

I've also removed the edit and remove actions for pending team members because they are non-functional

image

> [!NOTE]
> The format of this date differs slightly from the one I previously implemented for participants. It uses `<date> at <time>`, instead of `<date>, <time>`. The [participant designs](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=7037-268934&t=9s8IirFWr9gkWKsN-4) seem to have also been updated so I changed the date format there as well

#### Table Row

| image | image |

#### Context Menu

| image | image |

### 🪤 Peer Testing

1. Invite one or more team members to one of your applets, without accepting the invitation.
2. Navigate to the applet team tab and confirm the design changes to the table
3. Open the row context menu and confirm the invitation date is present in the correct format
4. Confirm the functionality of the copy actions
5. Change the language to French and repeat steps 1 - 4
6. Go to the dashboard team tab and confirm that pending team members do not appear

### ✏️ Notes

The rest of the context menu design was updated in a small way ("Access" was changed to "Team Member"), so I added those updates to this PR


src/modules/Dashboard/features/Managers/Managers.utils.tsx
```diff
@@ -52,26 +56,57 @@ export const getHeadCells = (sortableColumns?: string[], appletId?: string): Hea
...
-      action: removeAccessAction,
-      title: t('removeAccess'),
+      action: actions.removeTeamMemberAction,
+      title: t('removeTeamMember'),
       context: manager,
       customItemColor: variables.palette.dark_error_container,
       'data-testid': 'dashboard-managers-remove-access',
     },
-  ];
+  );
```
farmerpaul: These menu items should also be conditional on whether the member is pending invite or not. In Figma, they're [not present for pending invitations](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=7037-274213&t=sABkGTzwRH9OPgUR-4). Also, trying to perform either action (with Edit, only after pressing Save) causes a "No access to applet" error.

I'm guessing the [additional invitation actions](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=6709-188615&t=sABkGTzwRH9OPgUR-4) were purposefully omitted from this task because they're coming in a later sprint.

farmerpaul (approved):
Just made a note about the additional menu items that should be hidden for invitations, but I'm sure it will be a simple change, so preapproving!

src/modules/Dashboard/features/Managers/Managers.utils.tsx
sultanofcardio: I intentionally didn't remove those because of the lack of the cancel invitation action in this ticket (and I don't think removing these was the intent anyway). I got some [feedback](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1769#pullrequestreview-2091138461) to this end when I implemented this for the participant menu, and I figured it would make sense to leave removing that until when the cancel action is being implemented.

However, it seems that removing a team member functions differently than removing a participant because all participants have a subject, whereas invited team members have no access until they accept the invitation, and so the request fails.

I'll remove the menu items because they have no functionality anyway

PR #1822, opened 2024-06-21
feat: add rich tooltip to conditional dropdown (M2-6933)

ChaconC:
### 📝 Description

Adds rich tooltips to conditional dropdowns

🔗 [Jira Ticket M2-6933](https://mindlogger.atlassian.net/browse/M2-6933)

### 📸 Screenshots

CleanShot 2024-06-20 at 18 05 15

### 🪤 Peer Testing

- Navigate to Applet Editor
- Make sure you have at least 2 items
  - Preferably make sure the Question Title uses text formatting
- Navigate to Item Flow
- Create a conditional rule 
- Validate dropdown tooltips


src/shared/components/FormComponents/SelectController/SelectController.types.ts
```diff
@@ -10,18 +10,20 @@ export type Option = {
...
 
 export type GetMenuItem = {
   labelKey: string;
   value: string | boolean;
   itemDisabled: boolean;
   icon?: JSX.Element;
-  withoutKey?: boolean;
   hidden?: boolean;
+  tooltip?: string | JSX.Element;
+  tooltipPlacement?: 'bottom' | 'right';
```
JustinNusca: I wonder if it wouldn't be better to type these in the same way as what the Tooltip component actually expects, eg:

```jsx
  tooltip?: React.ReactNode;
  tooltipPlacement?: TooltipProps['placement'];
```

It doesn't make a difference to our actual use case in this PR, but being accurate like this could save us the possibility of having to debug a type error at some point down the line, if we ever iterate on this and we pass a prop value that we would otherwise expect to work.

This goes for both this and the `Option` type, above.

src/shared/components/Tooltip/Tooltip.tsx
```diff
@@ -17,7 +17,7 @@ const applyMaxSize = {
       ...state.styles.popper,
       display: 'flex',
       placeItems: state.placement.includes('bottom') ? 'flex-start' : 'flex-end',
-      height: `${height - 24}px`,
+      height: state.placement.includes('bottom') ? `${height - 24}px` : undefined,
```
JustinNusca: I understand this change, but I think it also introduces a couple drawbacks. 

This height value obviously doesn't work great for tooltips arranged to the left or right of the reference element, but opting out entirely means we allow tooltip content to potentially extend off screen. The `height` value itself it also still necessary for tooltips with "top" placements. 

Applying some of these styles is admittedly a bit cumbersome, because we can't directly apply the `maxHeight` to the actual styled tooltip (these styles are applied via its' own modifier function to the wrapper element around the tooltip, and are used to position the tooltip within it).

I gave tweaking the styles a shot myself to see if we could improve things somewhat, and this is what I came up with:

```jsx
    const isVerticallyArranged =
      state.placement.includes('bottom') || state.placement.includes('top');

    state.styles.popper = {
      ...state.styles.popper,
      display: 'flex',
      ...(isVerticallyArranged && {
        placeItems: state.placement.includes('bottom') ? 'flex-start' : 'flex-end',
      }),
      height: isVerticallyArranged ? `${height - 24}px` : undefined,
      maxHeight: `${height - 24}px`,
    };
  },
```

Here's the before/after of the change:

| Before | After |
|-|-|
| before | Screenshot 2024-06-25 at 12 16 32 PM |

... and tooltip content that _doesn't_ fill the max height is still positioned correctly.

src/shared/components/FormComponents/SelectController/SelectController.tsx
```diff
@@ -71,55 +71,68 @@ export const SelectController = <T extends FieldValues>({
...
+    tooltipPlacement,
   }: GetMenuItem) => (
     <StyledMenuItem
-      {...(!withoutKey && { key: labelKey })}
+      key={labelKey}
       uiType={uiType}
       value={value as string}
       disabled={itemDisabled}
       className={hidden ? 'hidden-menu-item' : ''}
+      component={forwardRef((props, ref) => {
```
JustinNusca: I know that this is a suggestion I made to you, and I'm critiquing my own code here, but using the `forwardRef` and an inline function here, inside the `SelectController` body, means that we're recreating the `StyledMenuItem` (or rather, the component the `StyledMenuItem` will render as) on every render.

I wonder if we can't move this to a separately defined component, maybe like so:

```tsx
const SelectControllerMenuItem = forwardRef<
  HTMLLIElement,
  {
    disabled?: boolean;
    tooltip?: React.ReactNode;
    tooltipPlacement?: TooltipProps['placement'];
  } & React.ComponentPropsWithoutRef<'li'>
>(({ disabled, tooltip, tooltipPlacement, ...props }, ref) => {
  const item = <StyledItem itemDisabled={disabled} ref={ref} selectDisabled={disabled} {...props} />;

  return tooltip ? (
    <Tooltip
      tooltipTitle={tooltip}
      placement={tooltipPlacement}
      children={disabled ? <li {...props} /> : item}
      enterNextDelay={200}
    />
  ) : item;
});
```

... and then update the `component` prop to pass a stable reference, like so:

```suggestion
      component={SelectControllerMenuItem}
      tooltip={tooltip}
      tooltipPlacement={tooltipPlacement}
```

Before you try this change, I should warn you: I tried this locally, but I ran into some issues with prop type validation when using the `component` prop in combination with a `StyledComponent` wrapped around a Material-UI component — This isn't the [first time either](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1714#discussion_r1599152417). 

Just as a heads up:

I think it's a worthwhile change, so I wanted it noted _somewhere_, buutt...

I'm also not going to block approval of this PR based on it, in case you decide not to try and address it (I suspect it's really a symptom of a larger configuration problem between Emotion and Material UI).

JustinNusca (commented):
Looks pretty good and the functionality works as expected.

I just have a couple suggestions for some style & typing tweaks

... and one optional change that I already anticipate being annoying to debug, so I'll leave it to you how much you wanna look into that one 😅.

src/shared/components/Tooltip/Tooltip.tsx
ChaconC: I took a different approach to achieve the purpose behind the new styles that works for all placement options.

| Bottom | Top | Right |
|-|-|-|
| CleanShot 2024-06-26 at 09 31 16 | CleanShot 2024-06-26 at 09 34 02 | CleanShot 2024-06-26 at 09 31 06 |

src/shared/components/FormComponents/SelectController/SelectController.tsx
ChaconC: This approach works and is somewhat more readable, thanks for the suggestion. I was able to get around the prop type validation but now we're relying on `shouldForwardProp` which can hide behavior a bit.

JustinNusca (approved):
👍 Nice job with the `preventOverflow` modifier. Way better than what I was doing to manually work around it with additional styles.

PR #1826, opened 2024-06-21
fix(mi): update edit action for Performance Task (M2-7091)

ChaconC:
### 📝 Description

Fixes navigation to Performance Task type activities from `ActivitySummaryCards`

🔗 [Jira Ticket M2-7091](https://mindlogger.atlassian.net/browse/M2-7091)

#### Before

https://github.com/ChildMindInstitute/mindlogger-admin/assets/2916156/e7764482-da70-4f03-8218-6da5492ffd01

#### After

https://github.com/ChildMindInstitute/mindlogger-admin/assets/2916156/efaf54d1-0094-4e2a-afc7-ad4631329725

### 🪤 Peer Testing

- Depends on [Backend PR #1436](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1436)
- Navigate to Applet
- Select more menu and edit Activity
  - Should navigate to `builder/{appletId}/activities/{activityId}/about`
- Select more menu and edit a Performance Type Activity
  - Should navigate to `builder/{appletId}/activities/performance-task/{performanceTaskType}/{activityId}` where `performanceTaskType` can be  `flanker|gyroscope|touch` 

### ✏️ Notes

You can create a Performance Task Type by selecting the "Add Performance Task" button in the Applet Builder
CleanShot 2024-06-21 at 14 20 02@2x

### DIFF
--- a/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
+++ b/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
@@ -144,7 +144,7 @@ export const useActivityGrid = (
                 menuItems={getActivityActions({
                   actions: actions || defaultActions,
                   appletId,
-                  activityId,
+                  activity,
                   dataTestId,
                   roles,
                   featureFlags,
--- a/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
+++ b/src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
@@ -17,6 +17,8 @@ import {
   StyledSvg,
 } from 'modules/Dashboard/components/ActivityGrid';
 import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
+import { getPerformanceTaskPath } from 'modules/Builder/features/Activities/Activities.utils';
+import { EditablePerformanceTasksType } from 'modules/Builder/features/Activities/Activities.types';
 
 import { useTakeNowModal } from '../TakeNowModal/TakeNowModal';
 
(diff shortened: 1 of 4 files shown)


src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
```diff
@@ -48,9 +50,18 @@ export const useActivityGrid = (
     () => ({
       editActivity: ({ context }: MenuActionProps<ActivityActionProps>) => {
         const { activityId } = context || {};
+        if (!activityId || !appletId) return;
+        const activity = getActivityById(activityId);
+
+        const navigateTo =
+          activity?.isPerformanceTask && activity?.performanceTaskType
+            ? getPerformanceTaskPath(
+                activity?.performanceTaskType as unknown as EditablePerformanceTasksType,
```
farmerpaul, line 61: I wasn't thrilled to see this `as unknown as EditablePerformanceTasksType` coercion, so I decided to investigate why it was necessary. Turns out it's used in [the Builder's Activities.tsk](https://github.com/ChildMindInstitute/mindlogger-admin/blob/fix/M2-7091-task-urls/src/modules/Builder/features/Activities/Activities.tsx#L68) because only certain tasks are editable – not the A/B Trails ones:

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/f670fa58-91c8-409c-9a58-f18b5dcda673

We're not doing that check here and showing the Edit Activity menu item regardless, and when you try to edit one of those performance task types from the Activity Grid, nothing happens. So we should hide that menu item in these cases.

Further, maybe a comment would good here to explain why this is needed, so future devs who are as picky as me about forced coercion aren't spending time investigating. 😛

farmerpaul (approved):
Just one thing I noticed about certain performance task types that I hope will be a quick fix, but preapproving! Thanks for taking care of this, @ChaconC!

sultanofcardio (commented):
Same comments as Paul. Clicking edit on AB Trails activity doesn't work and should be hidden. Personally, I think we should replace the enums with a string union type but that would probably introduce too many changes in this PR

src/modules/Dashboard/components/ActivityGrid/ActivityGrid.hooks.tsx
ChaconC: Added validation to remove edit button from ABTrails tasks
CleanShot 2024-06-24 at 10 42 03

sultanofcardio (approved):

PR #1829, opened 2024-06-24
feat(mi): add Upgrade to Full Account event tracking (M2-7114)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-7114](https://mindlogger.atlassian.net/browse/M2-7114)

This PR tracks new/updated Mixpanel events related to upgrading Limited Accounts to Full Accounts.

### 🪤 Peer Testing

Please refer to [JIRA ticket](https://mindlogger.atlassian.net/browse/M2-7114) for details around when events are triggered and what event properties should be attached to those events.

#### Prerequisites:

- To test, you'll need to have access to the **MindLogger Analytics Dev** project in Mixpanel; please let @farmerpaul know if you are planning to do peer testing so that you can be granted access.
- Ensure you have these env vars set locally:
  ```
  REACT_APP_ENV=dev
  REACT_APP_MIXPANEL_TOKEN=c1e1e662f03d9f1363768731077ea000
  REACT_APP_MIXPANEL_FORCE_ENABLE=true
  REACT_APP_DEVELOP_BUILD_VERSION=head
  ```

#### Testing:
* For each table row in the [JIRA ticket](https://mindlogger.atlassian.net/browse/M2-6735):
  2. Open the **MindLogger Analytics Dev** project in Mixpanel, and navigate to the [Events tab](https://mixpanel.com/project/3333580/view/3841316/app/events). Refresh if needed.

      **Expected outcome:** The event matching the **New Event Name** column should have been added to the top of the list.

      Screenshot 2024-06-18 at 4 32 53 PM

      **Expected outcome:** The custom properties for that event mentioned in the **Details** column should be listed, as well as **Applet ID** as the JIRA ticket says applies to all events.

### DIFF
--- a/src/modules/Dashboard/features/Participants/Participants.tsx
+++ b/src/modules/Dashboard/features/Participants/Participants.tsx
@@ -172,6 +172,11 @@ export const Participants = () => {
     nickname,
     tag,
   }: HandleUpgradeAccount) => {
+    Mixpanel.track('Upgrade to Full Account clicked', {
+      [MixpanelProps.AppletId]: appletId,
+      [MixpanelProps.Via]: 'Applet - Participants',
+    });
+
     setRespondentKey(respondentOrSubjectId);
     setParticipantDetails({ secretId, nickname, tag });
     handleSetDataForAppletPage({
--- a/src/modules/Dashboard/features/Applet/Popups/UpgradeAccountPopup/UpgradeAccountPopup.test.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/UpgradeAccountPopup/UpgradeAccountPopup.test.tsx
@@ -4,14 +4,14 @@ import mockAxios from 'jest-mock-axios';
 
 import { renderWithProviders } from 'shared/utils/renderWithProviders';
 import { mockedAppletId, mockedSubjectId1 } from 'shared/mock';
-import { expectBanner } from 'shared/utils';
-import * as MixpanelFunc from 'shared/utils/mixpanel';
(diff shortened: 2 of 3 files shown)


src/modules/Dashboard/features/Participants/Participants.tsx
```diff
@@ -172,6 +172,11 @@ export const Participants = () => {
     nickname,
     tag,
   }: HandleUpgradeAccount) => {
+    Mixpanel.track('Upgrade to Full Account clicked', {
+      [MixpanelProps.AppletId]: appletId,
+      [MixpanelProps.Via]: 'Applet - Participants',
+    });
```
sultanofcardio, line 178: Technically we can perform this action via the dashboard as well.

https://github.com/ChildMindInstitute/mindlogger-admin/blob/e6ff173fc6aaf31da3e81d92ab4f59caa8eef6da/src/modules/Dashboard/features/Respondents/Respondents.tsx#L162-L170

I wonder if we're interested in tracking those events as well. Is there a ticket for that already? If not, can you enquire about it?

sultanofcardio (approved):
This works great. There's one thing I'm curious about though

src/modules/Dashboard/features/Participants/Participants.tsx
farmerpaul: Wow, another good catch. There's [an existing task](https://mindlogger.atlassian.net/browse/M2-6680) in JIRA to migrate from Dashboard > Respondents to use the Participants screen's code; I'll make a note there.

PR #1830, opened 2024-06-24
feature: Add ability to duplicate Item Flows (M2-6907)

JustinNusca:
### 📝 Description

🔗 [M2-6907](https://mindlogger.atlassian.net/browse/M2-6907): [Admin] - Allow Admin to Duplicate Item Conditions

### 📸 Screenshots

duplicatin

### 🪤 Peer Testing

For an applet, with an activity populated by multiple items…

1. From the **Builder** screens's Activity tab for the given applet, press the Activity you want to edit, and then press the **Item Flow** tab to switch to it.
2. If no Item Flows exist yet, create one and fill in the relevant options.
3. Press the duplicate button on the item flow you want to duplicate.
4. Observe that a new Item Flow has been created. It should be appended to the end of the list of Item Flows, have an identical set of conditions as the Item Flow you chose to copy, and the "Shown Item" dropdown should be empty, with an error prompt that requires the user to set a value.

### ✏️ Notes

> [!NOTE]  
> The [second commit in this PR](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1830/commits/cf96b35e3ab21b022e899b841b9062e5246f573d) duplicates the changes from the previous commit to the `_old` variations of the same components.
>
> Control of which component is rendered is determined by the `enableItemFlowExtendedItems` feature flag.
>
> There should be no functional differences, but you may want to test with the flag both set/unset to verify. Note that when the flag is set, a number of unsupported item types will be rendered and selectable as options in Item Flows, which are not yet supported by the backend, and will return errors, as determined by [this check](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/blob/86cd23bd1151ccc097687a81e2c2504d6194df0b/src/apps/activities/domain/custom_validation.py#L51-L55).
>
> You may toggle this feature flag by adding …
>
> ```jsx
> features.enableItemFlowExtendedItems = false;
> ```
>
> … prior to the `return` statement [here in `useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/6dc81667f7dd06c89f4fab6e80236a011f0ed106/src/shared/hooks/useFeatureFlags.ts#L22-L28).

[M2-6907]: https://mindlogger.atlassian.net/browse/M2-6907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Builder/features/ActivityItemsFlow/ActivityItemsFlow.test.tsx
+++ b/src/modules/Builder/features/ActivityItemsFlow/ActivityItemsFlow.test.tsx
@@ -165,6 +165,23 @@ describe('Activity Items Flow', () => {
     expect(addItemFlow).not.toBeDisabled();
   });
 
+  describe('Duplicate ItemFlow', () => {
+    beforeEach(() => {
+      renderActivityItemsFlow(mockedAppletWithAllItemTypes);
+    });
+
+    test('Duplicates ItemFlow successfully', async () => {
+      fireEvent.click(screen.getByTestId(`${mockedTestid}-add`));
+
+      expect(screen.getAllByTestId(new RegExp(`^${mockedTestid}-\\d+$`))).toHaveLength(1);
+
+      fireEvent.click(screen.getByTestId(`builder-activity-item-flow-0-condition-0-name`));
+      fireEvent.click(screen.getByTestId(`${mockedTestid}-0-duplicate`));
+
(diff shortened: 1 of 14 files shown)


farmerpaul (approved):
Tested & LGTM :+1:

PR #1848, opened 2024-07-03
fix(mi): Close Web App tab after Take Now (M2-7159)

ionyRojas:
### 📝 Description

🔗 [Jira Ticket M2-7159](https://mindlogger.atlassian.net/browse/M2-7159)

Currently the application is window is unable to close itself for some security reason on browsers when using [window.close](https://developer.mozilla.org/en-US/docs/Web/API/Window/close) the solution is to create a message event and share that message event with the parent window whose open this window

Changes include:

- store the new window reference
- add a message listener whos validates the origin of the message

### 📸 Screenshots

https://github.com/ChildMindInstitute/mindlogger-web-refactor/assets/28871425/763761e3-d1e5-4cc5-94c6-39a8b645b102

### DIFF
--- a/src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
+++ b/src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
@@ -294,7 +294,21 @@ export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
         });
 
         setActivityOrFlow(null);
-        window.open(url.toString(), '_blank');
+
+        const newTab = window.open(url.toString(), '_blank');
+        // message received from the TakeNowSuccessModal.tsx file from the web app
+        // it's needed to close the new tab after the user clicks the "Close" button
+        window.addEventListener('message', function messageHandler(event) {
+          if (event.origin === url.origin) {
+            const message = event.data;
+
+            if (message === 'close-me') {
+              newTab?.close();
+            }
+
+            this.removeEventListener('message', messageHandler);
+          }
+        });
       }
     }, [targetSubject, sourceSubject, loggedInUser, isSelfReporting]);
 


src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -294,7 +294,21 @@ export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
         });
 
         setActivityOrFlow(null);
-        window.open(url.toString(), '_blank');
+
+        const newTab = window.open(url.toString(), '_blank');
+        // message received from the TakeNowSuccessModal.tsx file from the web app
+        // it's needed to close the new tab after the user clicks the "Close" button
```
farmerpaul, line 300: Nit: suggested minor improvement in comment clarity 😇 
```suggestion
        // Listen for message received from Web App's TakeNowSuccessModal component when it
        // requests to close its tab when the "Return to Admin App" button is clicked.
```

farmerpaul (approved):
Great solution @ionyRojas! Tested and it works. Just had a minor suggestion for comment clarity, but pre-approving pending that change.

Also, could you follow the recommended [PR title format](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/7503873/Configuration+Management+Plan#Pull-Requests) we established in the Confluence docs? We're trying to steer the team to use Conventional Commits format, and when squash-merging the PR, the PR's title becomes the title of the squash commit. So it should follow Conventional Commits format.

So instead of:
CleanShot 2024-07-03 at 11 33 46@2x

I'd suggest something like:

`fix(mi): Close Web App tab after Take Now (M2-7159)`

(They should be no more than 50-60 characters long, ideally.) Thank you!

farmerpaul:
Thanks for making those changes @ionyRojas. Please update/rebase this branch to use `develop` as the target branch, and then it should be good to merge. Thanks!

ionyRojas:
> Thanks for making those changes @ionyRojas. Please update/rebase this branch to use `develop` as the target branch, and then it should be good to merge. Thanks!

Rebased @farmerpaul

PR #1844, opened 2024-06-28
chore: update develop branch with Multi-Informant R1 changes (M2-7189)

mbanting:
### 📝 Description

🔗 [Jira Ticket M2-7189](https://mindlogger.atlassian.net/browse/M2-7189)

This is the pull request to merge the `features/multiinformant-metapod` branch to `develop`. This is a long-running feature branch that Pod 5 was working on to develop several multi-informant Release 1 features, including
- Take Now by Admins
- Updated information hierarchy and navigation at the dashboard and applet level, including new screens like the Applet overview
- Updated add user flows
- and more enhancements, improvements and bug fixes

Please note that all features that were implemented in `features/multiinformant-metapod` were added via PRs and reviewed, approved and peer tested.

### 🪤 Peer Testing

As mentioned above, every feature added in this branch has already been peer tested, reviewed and approved. However, a quick smoke test can be performed by enabling the following feature flags
- `enable-multi-informant`
- `enable-multi-informant__take-now`

and confirm the presence and core functionality of the following
- Global elements
  - Tabs
  - Breadcrumbs
- Applet Overview
- Activities Overview
- Participants Overview
  - Participant Details
  - Participant Schedule
- Schedule
- Team
- Take Now

### ✏️ Notes

This PR should be merged in sync with the corresponding [backend PR](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1469).

PR #1850, opened 2024-07-03
feat: Add "Edit event" confirmation popup (M2-7188)

JustinNusca:
### 📝 Description

🔗 [M2-7188](https://mindlogger.atlassian.net/browse/M2-7188): [Admin Panel] Multi-informant - Add warning if admin has made changes in the 'Edit event' pop-up and closes it with not saving changes

This PR is a re-implementation of work originally done in an earlier task ([M2-6871](https://mindlogger.atlassian.net/browse/M2-6871)), which was undone as a result of addressing irreconcilable merge conflicts when merging [the multi-informant branch into develop](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1844).

There are similarities in the approach taken by the original work, but they are not 1:1. Some changes I made include:

- Updated the `ScheduleProvider` to include the `SaveChangesPopup` and to manage it's state. This was a context provider I previously added to wrap both the `Legend` and `Calendar` components in order to synchronize state changes across both of these components, and in particular to manage/render various popups that either component may open.
- Removed the `removeWarning` field from the `EventForm` component's form state. I did this so that the `dirty` state of the form could be correctly calculated, in order to determine if the "Save Changes" popup should be shown. `removeWarning` was not used as field, but as derived state for showing visible warning flags to the user, which are not reset when resetting changes. This variable is now passed via props to the relevant components (I also pluralized it, because it's an object that contains multiple possible warnings).
- Save a ref to the `EditEventForm`'s own ref object in the `ScheduleProvider`, and use that to determine if the `SaveChangesPopup` should be shown, as well as call the `EditEventForm`'s submission handler. This is a similar approach to the previous implementation, and `EditEventForm` already exposed a custom ref that used `useImperativeHandle` to pass along a few methods, including the submit handler. I'm not 100% happy with this, and it could be more conventionally done by lifting the form state up, but untangling this would have been a much more significant refactor.

### 📸 Screenshots

You may click each gif for a larger version.

| When Saving Changes | When Discarding Changes | When Cancelling "Save Changes" prompt | 
|-|-|-|
| when-saving-changes | when discarding changes | when cancelling |

### 🪤 Peer Testing

This flow should be replicable for both Scheduled and Unscheduled events, for both Default and Individual schedules.

For an applet with some activities…

1. Navigate to the **Applet → Schedule** and/or the **Participant Details → Schedule** screens.
2. Press an event in the calendar view to open the Edit Event popup.
3. Make a change to the event, and then attempt to close the popup without saving changes.
4. Notice that a "Save Changes" popup opens, prompting you to confirm or discard your changes.
    1. When pressing "Cancel" on the "Save Changes" popup:
      a. Notice that the "Save Changes" popup closes, and returns you to the "Edit Event" popup.
      b. After returning to the "Edit Event" popup, undo any changes you made to the event. 
      c. Close the popup without saving changes. Notice that the "Save Changes" popup does not open, because no changes have been recorded.
    2. When pressing "Don't Save" on the "Save Changes" popup:
        a. Notice that the "Save Changes" and "Edit Event" popups both close, and the event is _not_ updated.
    3. When pressing "Save" on the "Save Changes" popup:
        a. Notice that the "Save Changes" and "Edit Event" popups both close, and the event is updated.

[M2-7188]: https://mindlogger.atlassian.net/browse/M2-7188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-6871]: https://mindlogger.atlassian.net/browse/M2-6871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.types.ts
+++ b/src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.types.ts
@@ -30,7 +30,7 @@ export type EventFormProps = {
   userId?: string;
 };
 
-export type Warning = {
+export type EventFormWarnings = {
   showRemoveAlwaysAvailable?: boolean;
   showRemoveAllScheduled?: boolean;
 };
@@ -53,7 +53,6 @@ export type EventFormValues = {
   idleTime: string;
   notifications: EventNotifications;
   reminder: FormReminder;
-  removeWarning: Warning;
 };
 
 export type NotificationTimeTestContext = {
@@ -87,7 +86,7 @@ export type GetEventFromTabs = {
   hasNotificationsErrors?: boolean;
   hasAlwaysAvailableOption?: boolean;
   'data-testid'?: string;
-  removeWarnings?: Warning;
+  removeWarnings?: EventFormWarnings;
 };
 
 export type GetDaysInPeriod = {
--- a/src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.types.ts
+++ b/src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.types.ts
@@ -1,4 +1,5 @@
 import { Dispatch, SetStateAction } from 'react';
+import { FormState } from 'react-hook-form';
 import * as yup from 'yup';
 
 import {
@@ -13,6 +14,7 @@ import { CalendarEvent } from 'modules/Dashboard/state';
 export type EventFormRef = {
   submitForm: () => void;
   processEvent: () => void;
+  formState: FormState<EventFormValues>;
 };
 
 export type EventFormProps = {
@@ -85,6 +87,7 @@ export type GetEventFromTabs = {
   hasNotificationsErrors?: boolean;
   hasAlwaysAvailableOption?: boolean;
   'data-testid'?: string;
+  removeWarnings?: Warning;
 };
 
 export type GetDaysInPeriod = {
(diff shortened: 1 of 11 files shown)


src/modules/Dashboard/features/Applet/Schedule/EventForm/EventForm.types.ts
```diff
@@ -85,6 +87,7 @@ export type GetEventFromTabs = {
   hasNotificationsErrors?: boolean;
   hasAlwaysAvailableOption?: boolean;
   'data-testid'?: string;
+  removeWarnings?: Warning;
```
farmerpaul: Just a suggestion to rename the `Warning` type as well, following on the heels of your `removeWarning` ⇒ `removeWarnings` change:
```suggestion
  removeWarnings?: EventFormWarnings;
```

farmerpaul (approved):
Looks great, with one small suggestion to rename one of the types.

Also, I noticed that when I change the Availability dropdown, and change it back to its original value, I'm still prompted to save changes. I suppose editing that value causes other form values to become dirty? No big deal IMO, I think it's fine to merge as is, unless it's an easy fix.

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/01b8c211-9ee4-4563-8aec-529c922197fb

mbanting (approved):
Thanks for re-applying this on the updated `develop` branch. Tested successfully locally and reviewed as well.

farmerpaul (approved):
Fix works great!

PR #1852, opened 2024-07-04
fix(mi): Delete scheduled from guest user (M2-7132) 

ionyRojas:
### 📝 Description

🔗 [Jira Ticket M2-7132](https://mindlogger.atlassian.net/browse/M2-7132)

add a validation to avoid the render of the scheduled icon on guess accounts

### 📸 Screenshots
image

### DIFF
--- a/src/modules/Dashboard/features/Participant/Schedule/Schedule.tsx
+++ b/src/modules/Dashboard/features/Participant/Schedule/Schedule.tsx
@@ -25,6 +25,7 @@ export const ParticipantSchedule = () => {
   const { result: appletData } = applet.useAppletData() ?? {};
   const { result: respondentsData } = users.useAllRespondentsData() || {};
   const { result: subjectData } = users.useSubject() ?? {};
+
   const allRespondentsLoadingStatus = users.useAllRespondentsStatus();
   const dispatch = useAppDispatch();
   const preparedEvents = usePreparedEvents(appletData);
@@ -34,6 +35,7 @@ export const ParticipantSchedule = () => {
     allRespondentsLoadingStatus === 'loading' ||
     subjectLoadingStatus === 'idle' ||
     subjectLoadingStatus === 'loading';
+
   const selectedRespondent = useMemo(
     () =>
       respondentsData?.find(({ secretIds }) =>
@@ -41,6 +43,7 @@ export const ParticipantSchedule = () => {
       ) ?? ({} as Respondent),
     [respondentsData, subjectData],
   );
+
   const { details, id: userId } = selectedRespondent ?? {};
   const respondentDetails = details?.[0] ?? {};
   const { respondentSecretId, hasIndividualSchedule, respondentNickname } = respondentDetails;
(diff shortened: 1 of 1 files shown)


farmerpaul:
@ionyRojas now that `feature/multiinformant-metapod` has been [merged](https://metalab.slack.com/archives/C06EKU790TY/p1720025784112139) to `develop`, please do not target PRs against that branch anymore – target `develop` instead. Thanks!

src/modules/Dashboard/features/Participant/Schedule/Schedule.tsx
```diff
@@ -98,7 +101,7 @@ export const ParticipantSchedule = () => {
         >
```
farmerpaul, line 101: I discovered that there's another way to trigger the same "Create Individual Schedule" action:

https://github.com/ChildMindInstitute/mindlogger-admin/assets/1282772/5276b2d8-0c9a-4a58-8ce7-d070c13999f5

Since the intention of this ticket is to prevent "Guest Account Submission" from creating an Individual Schedule, I think it's sensible to include fixing this in the scope of this ticket. I believe this will do the trick, by updating line 97 above to:

```tsx
          showEditDefaultConfirmation={
            !selectedRespondent.isAnonymousRespondent && !hasIndividualSchedule
          }
```

farmerpaul (approved):
This indeed fixes the AC, however I discovered another way to create an individual schedule, but also included a suggestion for how to fix it. Pending that change, this PR is good to go (once you rebase/target it for `develop` 😉).

src/modules/Dashboard/features/Participant/Schedule/Schedule.tsx
JustinNusca: This is a great catch, but rather than modifying `showEditDefaultConfirmation`, I'd suggest updating the `canCreateIndividualSchedule` prop value instead, like so:

```jsx
canCreateIndividualSchedule={!!userId && !selectedRespondent.isAnonymousRespondent}
```

This should prevent the creation of individual schedules entirely, while allowing for the "Edit Default Schedule" confirmation to continue to be shown prior to editing events, which is intended to be shown prior to making changes to the default schedule, like so:

Screenshot 2024-07-04 at 1 19 10 PM

farmerpaul: Great catch @JustinNusca! Much better solution.

mbanting: Thanks @farmerpaul @JustinNusca!
Typically we want to avoid changing the scope but given this is a quick update and it supports the overall objective, let's include this as well @ionyRojas

ionyRojas: changed, thanks for the suggestion 👌

ionyRojas: changed, thanks for the suggestion!

PR #1853, opened 2024-07-04
fix: Disable Take Now on unsupported activities (M2-7130)

JustinNusca:
### 📝 Description

🔗 [M2-7130](https://mindlogger.atlassian.net/browse/M2-7130): [Admin] Admin shouldn't initiate Take Now on Cog Tasks

This PR updates locations where the Take Now option is presented, and conditionally disables the option if the Activity or ActivityFlow contains items that are not supported on the web application.

This is handled by a new check, `getIsWebSupported`, which compares the activity's items to a list of supported types. This check is similar to one already performed in the web app, and the list [is copied from from there](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/f225e78ea30c2106f6a13ce96e3a3a3c69b9a5b2/src/abstract/lib/constants.ts#L5-L19). It'd be cool if we could share this data between apps at some point 😅.

I initially planned to make additional backend modifications to more easily support this (ie,  add child item responseType information directly to ActivityFlows, instead of mapping the list of flow.activityIds to actual activities, and then reducing to the activities' items) or even have this determination made directly on the backend — but @mbanting pointed out that swapping the endpoint used would be a smaller lift to getting this done, and I liked that we got to drop an additional query on the Activities pages as a result. Perhaps this would be worthwhile to revisit post-launch, especially if conditions for allowing actions like "Take Now" may continue to grow.

The original query for Flows was used because I naively assumed the activities query was paginated, based on the supported [query parameters specified in the docs](https://api-dev3.cmiml.net/docs#/Activities/applet_activities_activities_applet__applet_id__get). This endpoint is actually not paginated, and those [parameters are unused](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/blob/297697bd825f059a4d760f2776fc4bc34bb087cc/src/apps/activities/api/activities.py#L60-L99), so it should be safe to use for our purposes here (ie, we don't need to worry about not having a result for an activity used by a flow because we queried for the wrong page or similar issue).

### 📸 Screenshots

Screenshot 2024-07-02 at 10 36 19 AM

### 🪤 Peer Testing

For an applet, with an activity and/or activity flow containing an item with a responseType that is unsupported on the web app…

1. Navigate to the **Applet → Activities** screen.
2. Open the context menu for the unsupported Activity and/or Activity Flow.
3. Notice that the "Take Now" menu item is disabled. A tooltip is shown when hovering over the item.
4. Navigate to the **Participant Details → Activities** screen.
5. Open the context menu for the unsupported Activity and/or Activity Flow.
6. Notice that the "Take Now" menu item is disabled. A tooltip is shown when hovering over the item.

[M2-7130]: https://mindlogger.atlassian.net/browse/M2-7130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/api/api.ts
+++ b/src/modules/Dashboard/api/api.ts
@@ -76,6 +76,8 @@ import {
   DeleteFlowReviewParams,
   AssessmentFlowReviewParams,
   FeedbackNote,
+  GetActivityResponse,
+  GetActivityParams,
 } from './api.types';
 import { DEFAULT_ROWS_PER_PAGE } from './api.const';
 
@@ -471,6 +473,11 @@ export const getReviewFlowsApi = (
 
 export const getAnswerApi = ({ appletId, answerId }: ActivityAnswerParams, signal?: AbortSignal) =>
   authApiClient.get(`/answers/applet/${appletId}/answers/${answerId}`, { signal });
+
+export const getActivityApi = ({ activityId }: GetActivityParams, signal?: AbortSignal) =>
+  authApiClient.get(`activities/${activityId}`, { signal }) as Promise<
+    AxiosResponse<GetActivityResponse>
+  >;
 
 export const getActivityAnswerApi = (
   { appletId, answerId, activityId }: ActivityAnswerParams,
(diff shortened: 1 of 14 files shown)


farmerpaul (approved):
Nice work, and great that we can do away with an extra API request! Tested and everything works as expected.

PR #1868, opened 2024-07-18
feature: Constrain Take Now to availability schedule (M2-7180)

ionyRojas:
### 📝 Description

🔗 [Jira Ticket M2-7180](https://mindlogger.atlassian.net/browse/M2-7180)

- add helpers to validate the activity scheduled is on time and able to take the activity

Changes include:

-  validate the different types of schedule and validate if the current time is between the intervals of the scheduled time by the activity
- 
### 📸 Screenshots

image
image


farmerpaul (commented):
Unfortunately, I won't have time today to review this one before I'm off next week, but just wanted to note that we should always follow the [PR naming convention](https://arc.net/l/quote/mnwkxofj). I'd suggest `fix: Constrain Take Now to availability schedule (M2-7180)` 🙏🏻

ionyRojas:
@mbanting and should I also close this one ??

PR #1881, opened 2024-08-02
Feature/m2 7211 add slider validation

ionyRojas:
### 📝 Description

🔗 [Jira Ticket M2-7211](https://mindlogger.atlassian.net/browse/M2-7211)

Changes include:

-  Add display mode logic when user select the required fields
-  add base object to get the display mode information

### 📸 Screenshots

image
image| Before (Optional)


mbanting:
@ionyRojas please treat these as related but independent tickets that should be merged to `develop` (after all dependent tickets, such as M2-7076). This is a large feature-rich screen and the goal is to deliver this in smaller chunks instead of one big PR to review that will now need to take place in M2-7076.

PR #1882, opened 2024-08-05
feat: Adds Unity (merit) activity as an option on performance tasks (M2-7440)

ramirlm:
(https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/263127186/Admin+Panel+Applet+Builder+Library)

### 📝 Description

🔗 [Jira Ticket M2-7440](https://mindlogger.atlassian.net/browse/M2-7440)

Adds a new performance activity type for Merit POC to include a Unity Activity.

### 📸 Screenshots
CleanShot 2024-08-05 at 14 49 30

### 🪤 Peer Testing

It's necessary to enable the feature flag `enableMeritActivityType` to test the feature.

### DIFF
--- a/src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
+++ b/src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
@@ -1,6 +1,5 @@
 import i18n from 'i18n';
 import { PerfTaskType } from 'shared/consts';
-import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 
 import { ActivityAddProps, PerformanceTasks } from '../Activities.types';
 
@@ -8,14 +7,11 @@ const { t } = i18n;
 
 const getMenuItemTitle = (perfTask: string) => `performanceTasks.${perfTask}`;
 
-export const GetPerformanceTasksMenu = (
+export const getPerformanceTasksMenu = (
   onAddActivity: (props: ActivityAddProps) => void,
   setAnchorEl: (el: null | HTMLElement) => void,
+  enableMeritActivityType?: boolean,
 ) => {
-  const {
-    featureFlags: { enableMeritActivityType },
-  } = useFeatureFlags();
-
   const getAction = (props: ActivityAddProps) => {
     onAddActivity(props);
     setAnchorEl(null);
--- a/src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
+++ b/src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
@@ -1,5 +1,6 @@
 import i18n from 'i18n';
 import { PerfTaskType } from 'shared/consts';
+import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 
 import { ActivityAddProps, PerformanceTasks } from '../Activities.types';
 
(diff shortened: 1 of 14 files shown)


src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
```diff
@@ -1,22 +1,42 @@
...
+    featureFlags: { enableMeritActivityType },
+  } = useFeatureFlags();
+
   const getAction = (props: ActivityAddProps) => {
     onAddActivity(props);
     setAnchorEl(null);
   };
 
   return [
+    ...(enableMeritActivityType
```
ChaconC, line 21: can you add this activity at the bottom of the list

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -579,6 +579,7 @@ export const getNewPerformanceTask = ({
     [PerfTaskType.Touch]: getGyroscopeOrTouchItems(GyroscopeOrTouch.Touch),
     [PerfTaskType.ABTrailsMobile]: getABTrailsItems(DeviceType.Mobile),
     [PerfTaskType.ABTrailsTablet]: getABTrailsItems(DeviceType.Tablet),
+    [PerfTaskType.Unity]: getABTrailsItems(DeviceType.Mobile),
```
ChaconC: This is essentially adding a `ABTrails` task with a different name. This task should try to implement its own task properties, which will fail to get added until the BE is updated.
CleanShot 2024-08-05 at 14 42 54

src/resources/app-en.json
```diff
@@ -944,13 +944,15 @@
     "abTrailsMobile": "A/B Trails Mobile",
     "flanker": "Simple & Choice Reaction Time Task Builder",
     "gyroscope": "CST Gyroscope",
-    "touch": "CST Touch"
+    "touch": "CST Touch",
+    "unity": "New Unity Task (Beta)"
```
ChaconC, line 1000: While we don't have a name yet I'd avoid "New" and keep it as "Unity Task (Beta)" or (POC)

ChaconC (changes requested):

src/resources/app-en.json
ramirlm: This is per design @ChaconC 
CleanShot 2024-08-05 at 17 58 23@2x

Also, notice the layout of the select is different, but that's also out of scope for this ticket.

ramirlm: The `(Beta)` is also gray, but I'll ask Mike to keep it as it is, since it's a POC

src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
ramirlm: Sure

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
ramirlm: So can we keep this way with a TODO to change it after the BE is implemented?

src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
ChaconC: Nevermind, this is also as per design

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
ChaconC: Because this is behind a Feature Flag I'd prefer to have the full implementation in this ticket, similar to how `ParagraphItem` was handled, meaning you can't save Applets which include a `Unity` task

src/modules/Builder/features/Activities/ActivitiesHeader/ActivitiesHeader.utils.tsx
```diff
@@ -1,22 +1,42 @@
...
+import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 
 import { ActivityAddProps, PerformanceTasks } from '../Activities.types';
 
 const { t } = i18n;
 
 const getMenuItemTitle = (perfTask: string) => `performanceTasks.${perfTask}`;
 
-export const getPerformanceTasksMenu = (
+export const GetPerformanceTasksMenu = (
```
ChaconC: I understand why you renamed this, but this is neither a Component nor a Hook, violating [rules-of-hooks](https://react.dev/reference/rules/rules-of-hooks). You could instead pass the condition as a parameter.

ChaconC (commented):
Thanks for making the changes, as far as POC goes this should be enough, with the next step being uploading a config file (TBD).

ChaconC (approved):
Please be sure to include the ticket in the title (M2-xxx), and use squash-merge

PR #1884, opened 2024-08-08
feat: speed up data export (M2-7496)

IhorNesterenko99:
- [x] OSS packages added to MindLogger [open source credit page](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/263127186/Admin+Panel+Applet+Builder+Library)

### 📝 Description

🔗 [Jira Ticket M2-7496](https://mindlogger.atlassian.net/browse/M2-7496)

This PR introduces significant performance improvements to the data export process for applet activities with a large number of answers, reducing export time by over 50%.

**Key Changes:**

   - **Concurrent Requests:** Data export time is improved by making concurrent requests to the server and processing data in batches without waiting for previous responses to decrypt.
   - **Web Workers:** Web Workers are utilized to handle decryption in separate threads, enhancing the overall export process.
   - **Feature Flag 'enableDataExportSpeedUp'**: Introduced a feature flag, which is set to "ON" in the DEV, TEST, and UAT environments. If the flag is "OFF", the application will revert to the previous data export process.


src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (!exportDataQueue) return;
+
+    const processingBatch = exportDataQueue[currentPage - 1];
+
+    if (processingBatch) {
+      processData(processingBatch, currentPage);
```
BamMironov: `useEffect` should only be used to sync components with an external system:
https://react.dev/reference/react/useEffect

In this case, there is no synchronization with an external API. Hence, it should work the other way. For instance, a user event should invoke the `processData` (e.g. click).

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+      return worker;
+    });
+
+    workersRef.current = workers;
+
+    return () => {
+      workers.forEach((worker) => worker.terminate());
+    };
+    // create workers only once when the component mounts
```
BamMironov: Please elaborate on why they shouldn't be recreated. AFAIK, React's Strict Mode runs every effect twice in dev. So, if nothing happens in dev mode then there shouldn't; be a problem to recreated those workers

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export const useMultipleDecryptWorkers = ({
+  handleExportPopupClose,
+  appletId,
+  encryption,
+  filters,
+}: MultipleDecryptWorkersProps) => {
+  const [dataIsExporting, setDataIsExporting] = useState(false);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [exportDataQueue, setExportDataQueue] = useState<ExportDataResult[] | null>(null);
```
BamMironov, line 20: Is there a reason to use `useState` for `exportDataQueue` and `currentPage`? None of the gets exported and used in the render

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
BamMironov, line 1: I have a strong feeling that though the code is working its business logic should not rely on React's API. This means that the implementation that makes use of React hooks seems cumbersome and unnecessary.
Most of the code could be moved to a dedicated class(es) or functions that could be easy to read and test.
Some of the useState and useRef usages give the impression of public/private fields of a class that could be used inside public/private methods.

Moreover, adding tests to such classes/function would not be a big deal compared to the hook.
The hook may still be necessary though. However, it should be a bridge between Business Logic layer and React world.

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup_new.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const hasEncryptionCheck = useCheckIfHasEncryption({
+    isAppletSetting,
+    appletData: chosenAppletData,
+    callback: handleDataExportSubmit,
+  });
+
+  const showEnterPwdScreen = !!chosenAppletData && !dataIsExporting && !hasEncryptionCheck;
+
+  const executeAllPagesOfExportData = useCallback(
```
BamMironov: This also, gives me an impression of a separate class/function that could host this responsibility

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup_new.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  useEffect(() => {
+    setActiveModal(showEnterPwdScreen ? Modals.PasswordCheck : Modals.DataExport);
+  }, [showEnterPwdScreen]);
+
+  useEffect(
+    () => () => {
+      if (abortControllerRef.current) {
+        // stop requests to the API if the component unmounts
+        abortControllerRef.current.abort();
```
BamMironov: This is a good example of using `useEffect` as it syncs with the external API (AbortController). However, the right approach would be to hide the aborController in a class instance AND invoke a special method that describes itself. For example:
```
useEffect(
    () => () => {
      ExportDataService.cancel()
    }
  [],
);
```

BamMironov (changes requested):
Please consider the comments below

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
IhorNesterenko99: `exportDataQueue` is used for storing the queue of received responses from the database. When it changes, the corresponding `useEffect` is triggered to process the next part of the queue. The part of `exportDataQueue` that needs processing is determined by `currentPage`. They are not meant to be used in the render process, but rather to manage side effects in the component.

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
IhorNesterenko99: There is no user event to invoke `processData` because the goal is to fetch all the pages of export data sequentially. This happens automatically as each page is received, without requiring user interaction. By using `useEffect`, we ensure that the data fetching and subsequent decryption/processing happen continuously and in parallel. This approach allows us to handle the data processing efficiently, as each page triggers the next step in the sequence without waiting for a user action.

src/modules/Dashboard/features/Respondents/Popups/DataExportPopup/DataExportPopup.hooks.tsx
IhorNesterenko99: Why Web Workers shouldn't be recreated:

- resource efficiency: Creating web workers is a resource-intensive operation. Recreating them unnecessarily would consume more CPU and memory, leading to performance issues.

- task continuity: If workers were recreated during the component's lifecycle, any ongoing tasks assigned to the previous workers would be lost, potentially leading to incomplete processing or errors.

- React strict mode: In development mode, React's Strict Mode intentionally mounts, unmounts, and remounts components to help identify side effects and potential issues. Properly managed workers that function correctly during this process indicate that the code is handling side effects properly. However, recreating workers every time the component re-renders or runs effects can lead to unintended consequences, such as multiple workers processing the same tasks.

- avoiding duplicated Workers: Recreating workers on every render or effect run could lead to multiple instances of workers processing the same data or conflicting with each other, resulting in bugs and unpredictable behavior.

BamMironov (approved):

PR #1865, opened 2024-07-15
feat: Add Phrasal Template item type (M2-7076)

JustinNusca:
### 📝 Description

🔗 [M2-7076](https://mindlogger.atlassian.net/browse/M2-7076): [Action Plan] New Phrasal Template Item Type

This PR updates the builder to allow for the creation of a new Item type, the "Phrasal Template".

When the `enablePhrasalTemplate` feature flag is enabled, a new "Downloadable" group will appear in the "Item Type" dropdown, containing a new item type, "Phrasal Template".

When selected, the UI for the item's configuration will update to allow the admin user to configure the phrases, and the fields within those phrases.

These fields are connected to the existing form state of the builder, and are handled via similar validation and passed as part of the existing payload when the applet is updated after successful validation, when the admin user saves their work.

> [!WARNING] 
> The ability to add/remove phrases, as well as fields within those phrases is technically out of scope for this task, and is to be handled in [M2-7183](https://mindlogger.atlassian.net/browse/M2-7183). That said, by establishing the types correctly and rendering/manipulating fields via `react-hook-form`'s `useFieldArray`, the effort for managing add/remove behaviours was very minimal, and something I opted to wire up as a means of testing form validations in the browser (eg, that there must be at least two fields, that there must be at least one field of a given type, that there must be at least one phrase). 
>
> I left these in place, but they do not complete M2-7183, which includes additional UX like confirmation dialogs, but they are easily disabled if we don't want to allow for this behaviour for the time being.

### 🪤 Peer Testing

> [!NOTE] 
> This feature depends on the `enablePhrasalTemplate` feature flag. This _should_ already be enabled in dev/local environments, but if needed, you can add `features.enablePhrasalTemplate = true` prior to the `return` statement [here in `useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/6dc81667f7dd06c89f4fab6e80236a011f0ed106/src/shared/hooks/useFeatureFlags.ts#L22-L28).

For an applet, with an activity that contains a number of items:

1. From the Dashboard, choose "Edit Applet" from the applet's dropdown menu.
2. Navigate to the "Activities" tab in the Applet builder, and then choose an activity that contains multiple items.
3. Navigate to the "Items" tab in the Activity builder, and choose "Add Item" and/or edit an existing item.
4. In the "Item Type" dropdown, observe that "Phrasal Template" is available as a selectable option, and is located in the "Downloadable" group.
[...]

PR #1886, opened 2024-08-09
feat:  add from response field (M2-7211) 

ionyRojas:
### 📝 Description

🔗 [Jira Ticket M2-7211](https://mindlogger.atlassian.net/browse/M2-7211)

Changes include:

- Add global const for repeated keywords
- add support for sliderRows

### 📸 Screenshots

image

### DIFF
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
@@ -48,6 +48,7 @@ export const DisplayModeOptions: Readonly<
 
 export const KEYWORDS = {
   SENTENCE: 'sentence',
+  DISPLAY_SENTENCE: 'sentence',
   ITEM_RESPONSE: 'item_response',
   LINE_BREAK: 'line_break',
 } as const;
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
@@ -1,8 +1,7 @@
 import { ItemResponseType } from 'shared/consts';
 
-export const DisplayModeOptions: Record<
-  ItemResponseType,
-  { id: string; name: string }[] | undefined
+export const DisplayModeOptions: Readonly<
+  Record<ItemResponseType, { id: string; name: string }[] | undefined>
 > = {
   [ItemResponseType.MultipleSelection]: [
     { id: 'sentence', name: 'Sentence' },
@@ -20,7 +19,6 @@ export const DisplayModeOptions: Record<
     { id: 'bullet_list_option_row', name: 'Bullet List: Option Text, Row Text' },
     { id: 'bullet_list_text_row', name: 'Bullet List: Row Text, Option Text' },
   ],
-  [ItemResponseType.Slider]: undefined,
   [ItemResponseType.ABTrails]: undefined,
   [ItemResponseType.Audio]: undefined,
   [ItemResponseType.AudioPlayer]: undefined,
@@ -34,6 +32,7 @@ export const DisplayModeOptions: Record<
   [ItemResponseType.Photo]: undefined,
   [ItemResponseType.PhrasalTemplate]: undefined,
   [ItemResponseType.SingleSelection]: undefined,
+  [ItemResponseType.Slider]: undefined,
   [ItemResponseType.SliderRows]: undefined,
   [ItemResponseType.StabilityTracker]: undefined,
   [ItemResponseType.Text]: undefined,
@@ -43,3 +42,9 @@ export const DisplayModeOptions: Record<
   [ItemResponseType.TouchTest]: undefined,
   [ItemResponseType.Video]: undefined,
 };
+
+export const KEYWORDS = {
+  SENTENCE: 'sentence',
+  ITEM_RESPONSE: 'item_response',
+  LINE_BREAK: 'line_break',
+} as const;
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
@@ -1,7 +1,9 @@
 import { ItemResponseType } from 'shared/consts';
 
 export const DisplayModeOptions: Readonly<
-  Record<ItemResponseType, { id: string; name: string }[] | undefined>
+  Record<ItemResponseType, { id: string; name: string }[] | undefined> & {
+    default: undefined;
+  }
 > = {
   [ItemResponseType.MultipleSelection]: [
     { id: 'sentence', name: 'Sentence' },
@@ -41,6 +43,7 @@ export const DisplayModeOptions: Readonly<
   [ItemResponseType.TouchPractice]: undefined,
   [ItemResponseType.TouchTest]: undefined,
   [ItemResponseType.Video]: undefined,
+  default: undefined,
 };
 
 export const KEYWORDS = {
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
@@ -30,7 +30,7 @@ export function RenderedField({
     id: string;
     items?: Array<{ id: string; name: string }>;
   }>();
-  const [fromResponse, setFromResponse] = useState<{
+  const [responseFrom, setResponseFrom] = useState<{
     name: string;
     items?: SliderRowsItemResponseValues[];
   }>();
@@ -38,7 +38,7 @@ export function RenderedField({
   const params = useParams();
   const { control, getValues, setValue } = useCustomFormContext();
   const fieldValue = getValues(name as string);
-  const activitiesFromStore = applet.useActivityDataFromApplet(params?.activityId || '');
+  const itemsFromStore = applet.useActivityItemsFromApplet(params?.activityId || '');
   const { question: selectedOptionQuestion } =
     responseOptions?.find(({ name }) => fieldValue?.itemName === name) ?? {};
 
@@ -72,9 +72,9 @@ export function RenderedField({
     const selectedItem = responseOptions.find(({ name }) => name === value);
 
     if (!selectedItem) {
-      const missedItem = activitiesFromStore?.find((activity) => {
-        if (fieldValue?.itemName?.includes(activity?.name)) {
-          return activity;
+      const missedItem = itemsFromStore?.find((item) => {
+        if (fieldValue?.itemName?.includes(item?.name)) {
+          return item;
         }
 
         return null;
@@ -93,7 +93,7 @@ export function RenderedField({
         id: selectedItem?.responseType as string,
         items,
       });
-      setFromResponse(undefined);
+      setResponseFrom(undefined);
       setValue(name, {
         ...fieldValue,
         itemIndex: 0,
@@ -102,10 +102,10 @@ export function RenderedField({
     }
 
     if (
-      selectedItem?.name !== fromResponse?.name &&
+      selectedItem?.name !== responseFrom?.name &&
       selectedItem?.responseType === ItemResponseType.SliderRows
     ) {
-      setFromResponse({
+      setResponseFrom({
         name: selectedItem?.name,
         items: selectedItem?.responseValues?.rows,
       });
@@ -133,7 +133,7 @@ export function RenderedField({
     return selectedDisplayMode?.name;
   };
 
-  const getFromResponseRenderedValue = (value: unknown) => {
+  const getResponseFromRenderedValue = (value: unknown) => {
     if (!value && typeof Number(value) !== 'number') {
       return (
         <StyledBodyLarge color={variables.palette.outline}>
@@ -142,7 +142,7 @@ export function RenderedField({
       );
     }
 
-    const selectedFromResponse = fromResponse?.items?.[Number(value)];
+    const selectedFromResponse = responseFrom?.items?.[Number(value)];
 
     return selectedFromResponse?.label;
   };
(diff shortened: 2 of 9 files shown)


farmerpaul:
@ionyRojas I'm happy to review this, but it still includes a bunch of commits from `jn/feature/M2-7076-add-phrasal-template-item-type`. Can you simplify/rebase this so that it only includes the commits relevant to this branch?

ionyRojas:
give me a while, I"m fixing some conflicts and I'll upload this again @farmerpaul

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
```diff
@@ -43,3 +42,9 @@ export const DisplayModeOptions: Record<
   [ItemResponseType.TouchTest]: undefined,
   [ItemResponseType.Video]: undefined,
 };
+
+export const KEYWORDS = {
+  SENTENCE: 'sentence',
+  ITEM_RESPONSE: 'item_response',
+  LINE_BREAK: 'line_break',
+} as const;
```
farmerpaul: Appreciate your creating constants for these strings!

One nit: We tend to use enums in the code for this sort of thing. Constants are fine, too, but this just seemed like a perfect opportunity to use an enum.

Secondly, I noticed you're using the `SENTENCE` value for two distinct purposes in the code: as an **item type** (`sentence`/`item_response`/`line_break`), as well as a **display mode value** (`sentence`/`bullet_list`/`sentence_option_row`/`sentence_row_option`/etc.).

It's an unfortunate coincidence that the keyword `sentence` is being used by the BE for two semantically distinct properties, and I think it's important we don't confuse the two purposes of that keyword. So I think two separate, meaningfully named enums would make more sense:

```suggestion
export enum FieldType {
  Sentence = 'sentence',
  ItemResponse = 'item_response',
  LineBreak = 'line_break',
}

export enum DisplayMode {
  Sentence = 'sentence',
  SentenceOptionRow = 'sentence_option_row',
  SentenceRowOption = 'sentence_row_option',
  BulletListOptionRow = 'bullet_list_option_row',
  BulletListTextRow = 'bullet_list_text_row',
}
```

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -14,21 +14,26 @@ import {
...
   responseOptions = [],
-  type = 'sentence',
+  type = KEYWORDS.SENTENCE,
   ...otherProps
 }: Omit<PhrasalTemplateFieldProps, 'canRemove' | 'onRemove'>) {
   const [displayMode, setDisplayMode] = useState<{
     id: string;
     items?: Array<{ id: string; name: string }>;
   }>();
+  const [fromResponse, setFromResponse] = useState<{
```
farmerpaul: Nit: The dropdown has the title "Response From", not "From Response" – so I think it would be good to stay aligned in the identifier naming for consistency:
```suggestion
  const [responseFrom, setResponseFrom] = useState<{
```

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
```diff
@@ -20,7 +19,6 @@ export const DisplayModeOptions: Record<
     { id: 'bullet_list_option_row', name: 'Bullet List: Option Text, Row Text' },
     { id: 'bullet_list_text_row', name: 'Bullet List: Row Text, Option Text' },
```
farmerpaul, line 22: I just noticed that none of these values are being internationalized. Sorry I didn't catch this on my last review! It's essential we avoid hard-coding English-language strings anywhere in the code except in the language files, `app-en.json` and `app-fr.json`, and they must always be accessed via i18next functions (`useTranslation()`, etc.).

So I would use the `id` as the key that we look these values up via i18next. There shouldn't be a `name` attribute here at all, because i18next should be returning those strings via that `id` key.

I'll provide some guidance here as I get that it might not be completely obvious to you. Some minor refactoring is needed to support this:

```ts
export const DisplayModeOptions: Readonly<Record<ItemResponseType, string[] | undefined>> = {
  [ItemResponseType.MultipleSelection]: ['sentence', 'bullet_list'],
  [ItemResponseType.SingleSelectionPerRow]: [
    'sentence_option_row',
    'sentence_row_option',
    'bullet_list_option_row',
    'bullet_list_text_row',
  ],
  …
  // and so on…
};
```

Then in our language files:

```js
// app-en.json

  "phrasalTemplateItem": {
    …
    "displayModes": {
      "sentence": "Sentence",
      "bullet_list": "Bullet List",
      "sentence_option_row": "Sentence: Option Text, Row Text",
      "sentence_row_option": "Sentence: Row Text, Option Text",
      "bullet_list_option_row": "Bullet List: Option Text, Row Text",
      "bullet_list_text_row": "Bullet List: Row Text, Option Text"
    }
  },
```

```js
// app-fr.json

  "phrasalTemplateItem": {
    …
    "displayModes": {
      "sentence": "Phrase",
      "bullet_list": "Liste à puces",
      "sentence_option_row": "Phrase : Texte de l'Option, Texte de la Ligne",
      "sentence_row_option": "Phrase : Texte de la Ligne, Texte de l'Option",
      "bullet_list_option_row": "Liste à puces : Texte de l'Option, Texte de la Ligne",
      "bullet_list_text_row": "Liste à puces : Texte de la Ligne, Texte de l'Option"
    }
  },
```

And finally, `PhrasalTemplateRenderField` will need a few changes:

```tsx
export function RenderedField({ … }) {
  const [displayMode, setDisplayMode] = useState<{
    id: string;
    items?: Array<string>; // 👈 simplify this
  }>();

  …

  const getDisplayModeRenderedValue = (value: unknown) => {
    if (!value) {
      return (
        <StyledBodyLarge color={variables.palette.outline}>
          {t('phrasalTemplateItem.fieldResponsePlaceholder')}
        </StyledBodyLarge>
      );
    }

    // 👇 much simpler, don't need to look up the value in the array
    return t(`phrasalTemplateItem.displayModes.${value}`);
  };

  …
          {displayMode?.items && (
            <SelectController
              …
              options={displayMode?.items.map((value) => ({ // 👈 this now just an array of strings
                labelKey: value, // 👈 these are used for React unique `key` prop and are not meant to be English-language labels
                tooltip: <StyledMdPreview modelValue={(name as unknown as string) ?? ''} />,
                value, // 👈 update this
              }))}
              …
            />
          )}
  …  
}
```

Once you've made these changes, make sure to test your code to confirm translation is working correctly by switching the language in the app:

mbanting:
@ionyRojas reminder to use the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) convention of using `feat` 
It's a small nitpick but we may be using some tooling in the future that will look for that prefix rather than `Feature`
I'll update the title for this one.

If it's not ready yet, do you mind converting this back to draft. Looks like resolving the conflicts is taking longer than a quick update. Once it's ready for a review again you can mark it as such. Thanks!

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -102,8 +133,22 @@ export function RenderedField({
     return selectedDisplayMode?.name;
   };
 
+  const getFromResponseRenderedValue = (value: unknown) => {
```
farmerpaul: Same comment as earlier regarding identifier naming consistency:
```suggestion
  const getResponseFromRenderedValue = (value: unknown) => {
```

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -68,7 +73,7 @@ export function RenderedField({
 
     if (!selectedItem) {
       const missedItem = activitiesFromStore?.find((activity) => {
```
farmerpaul: I understand that Justin was the one to create the `useActivityDataFromApplet` Redux selector, in the previous PR (#1865). However, referring to this array value as "activities" was really confusing to me, because that selector doesn't actually return a list of _activities_. It returns a list of _items_ (as in, questions) for the given activity. I was scratching my head for a while trying to understand this code, confused why we were comparing an item name with an activity name, and returning an activity from this `getRenderedValue` function, until I grasped what was going on.

To improve this, I'd like to request the following improvement to semantics:
1. Rename the `useActivityDataFromApplet` Redux selector to `useActivityItemsFromApplet`
2. Rename the variable `activitiesFromStore` to `itemsFromStore`
3. In this `find` callback, rename the parameter `activity` to `item`

Thank you! This will just help keep my brain (and other devs' brains) from exploding when reading the code. 😉

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -78,10 +83,36 @@ export function RenderedField({
       return missedItem?.name;
     }
 
-    if (selectedItem?.responseType !== displayMode?.id) {
+    if (
+      selectedItem?.responseType !== displayMode?.id &&
+      selectedItem?.responseType !== ItemResponseType.SliderRows
+    ) {
+      const items = DisplayModeOptions[selectedItem?.responseType || 'default'];
```
farmerpaul, line 90: `default` isn't an acceptable property of the `DisplayModeOptions` object – that key doesn't exist. TypeScript is currently allowing this code to pass, presumably due to the extra expression complexity, but you can see that `default` is actually invalid if you were to try to access it directly:

CleanShot 2024-08-12 at 15 24 42@2x

Instead, I would do this:

```suggestion
      const items = selectedItem ? DisplayModeOptions[selectedItem.responseType] : undefined;
```

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -78,10 +83,36 @@ export function RenderedField({
...
+        itemIndex: 0,
+        displayMode: items ? fieldValue?.displayMode : KEYWORDS.SENTENCE,
+      });
+    }
+
+    if (
+      selectedItem?.name !== fromResponse?.name &&
+      selectedItem?.responseType === ItemResponseType.SliderRows
+    ) {
+      setFromResponse({
```
farmerpaul: I just want to note that this function is a render function – in general, it's important that any functions that are a part of the render thread _do not set state_. Doing so runs the risk of causing infinite loops. State should only be set in effect callbacks or event handler callbacks.

I've done some testing, however, and it seems that this code is not causing any problems in the UI, so rather than refactor what you've done using effects, etc., I think it's fine to leave it as is. However in the future, we should avoid violating this rule of React state management to be safe.

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
```diff
@@ -146,9 +191,27 @@ export function RenderedField({
...
+          {fromResponse?.items && (
+            <SelectController
+              name={`${name}.itemIndex`}
+              control={control}
+              defaultValue=""
+              fullWidth
+              options={fromResponse?.items.map(({ label }, index) => ({
+                labelKey: label as string,
+                tooltip: <StyledMdPreview modelValue={(name as unknown as string) ?? ''} />,
+                value: `${index}` ?? '',
```
farmerpaul: This nullish coalescing operation will never be used because a zero-length string never evaluates to nullish (it can be falsey, but never nullish), so you can remove it:
```suggestion
                value: `${index}`,
```

farmerpaul (changes requested):
Hi @ionyRojas, I've done some testing, and I haven't run into any issues, so nice work. I did have some comments on the code – some localization concerns, and suggestions for making the code easier to follow. Sorry it took me so long to review – I just found the naming of things to throw me off a few times; hopefully my suggestions will help future devs be able to understand it as well. Thank you! 🙏🏻

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
ionyRojas: good comment I understand your point but there is a point of why in TS is preferable to use a const with the `as const` word instead of enum... for more information look at this https://www.youtube.com/watch?v=jjMbPt_H3RQ
but in general the recommendation is to avoid using enum inside ts it's transformed basically into a IIFE so my recommendation is to avoid using enums in typescript
I made a little playground with that, that's the same that Matt Pocock mentioned on his video
image

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
ionyRojas: hmm I forgot to add the default, is better to add the default because it's acting as a switch statement and the idea is stop if-ting everything. I'll add the default as null as the other keywords

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
farmerpaul: Sure, `as const` is fine with me. Thanks for sharing the video!

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateRenderField.tsx
ionyRojas: if that's not generating an infinite loop or reloading your page on every render or generating a state mutation or generating re-renderings, it means it's not breaking any rule ;) and that's why it's working

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
ionyRojas: I tried this change but I'm reversing this because we need the ID to be able to render the deleted items, when you lose the reference you need the ID for that array. That's why it was added

farmerpaul (dismissed):

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplateField/PhrasalTemplateField.const.ts
farmerpaul: I just noticed during testing that I'm only seeing the French version of the display mode in the _selected value_ of the dropdown, but the listed dropdown options are still untranslated:

CleanShot 2024-08-15 at 16 31 43@2x

Can you fix this?

farmerpaul (approved):

PR #1894, opened 2024-08-20
feat: add activity/flow auto-assign checkbox (M2-7390)

qiushihe:

### 📝 Description

🔗 [Jira Ticket M2-7390](https://mindlogger.atlassian.net/browse/M2-7390)

This PR adds a "auto assign" checkbox to the "About Activity" and the "About Activity Flow" screens.

### 📸 Screenshots

activity

activity-flow

### 🪤 Peer Testing

1. Create an applet
2. Add an activity, keeping the "auto assign" checkbox checked
3. Add an activity flow, keeping the "auto assign" checkbox checked
4. Save and publish applet
5. Edit the just created applet
6. Edit the activity
7. The "auto assign" checkbox should still be checked
8. Uncheck the "auto assign" checkbox
9. Edit the activity flow
10. The "auto assign" checkbox should still be checked
11. Uncheck the "auto assign" checkbox
12. Save and publish applet
13. Edit the applet
14. Edit the activity
15. The "auto assign" checkbox should now be unchecked
16. Edit the activity flow
17. The "auto assign" checkbox should now be unchecked

### ✏️ Notes

N/A

### DIFF
--- a/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
+++ b/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
@@ -174,6 +174,21 @@ export const ActivityAbout = () => {
       onCustomChange: handleIsReviewableChange,
       'data-testid': 'builder-activity-about-reviewable',
     },
+    {
+      name: `${fieldName}.autoAssign`,
+      defaultControllerValue: true,
+      label: (
+        <StyledBodyLarge sx={{ position: 'relative' }}>
+          <span>{t('autoAssignActivity')}</span>
+          <Tooltip tooltipTitle={t('autoAssignTooltip')}>
+            <span>
+              <StyledCheckboxTooltipSvg id="more-info-outlined" />
+            </span>
+          </Tooltip>
+        </StyledBodyLarge>
+      ),
+      'data-testid': 'builder-activity-about-auto-assign',
+    },
   ];
 
   return (
@@ -210,10 +225,19 @@ export const ActivityAbout = () => {
       </StyledTitleMedium>
       <StyledFlexColumn>
         {checkboxes.map(
-          ({ name, label, isInversed, disabled, 'data-testid': dataTestid, onCustomChange }) => (
+          ({
+            name,
+            label,
+            isInversed,
+            disabled,
+            'data-testid': dataTestid,
+            onCustomChange,
+            defaultControllerValue,
+          }) => (
             <CheckboxController
               key={name}
               control={control}
+              defaultControllerValue={defaultControllerValue}
               name={name}
               label={label}
               disabled={disabled}
--- a/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
+++ b/src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
@@ -176,7 +176,6 @@ export const ActivityAbout = () => {
     },
     {
       name: `${fieldName}.autoAssign`,
-      defaultControllerValue: true,
       label: (
         <StyledBodyLarge sx={{ position: 'relative' }}>
           <span>{t('autoAssignActivity')}</span>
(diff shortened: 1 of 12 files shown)


src/shared/components/FormComponents/CheckboxController/CheckboxController.types.ts
```diff
@@ -8,6 +8,7 @@ type FormCheckboxProps = {
   onCustomChange?: (event: ChangeEvent<HTMLInputElement>) => void;
   'data-testid'?: string;
   sxLabelProps?: SxProps;
+  defaultControllerValue?: boolean;
```
qiushihe: There seems to be a lot of interference from both `CheckboxProps` and `UseControllerProps<T>` regarding the type of the `defaultValue` attribute. I couldn't get a "optional boolean" `defaultValue` attribute to register properly, even if I put `& { defaultValue?: boolean }` at the end of `export type InputControllerProps = ...`. It just keeps getting overridden to not accept `boolean` values.

So I decided to add an `defaultControllerValue` attribute instead.

Also, `CheckboxController` actually only need very few of these props, and it spread the rest of props (including `defaultValue`) into an inner `<Checkbox />` component. But this `defaultControllerValue` need to be set on the `<Controller />` component instead. So I feel adding the explicit `defaultControllerValue` is the most clear and clean implementation, given what we have to work with.

src/modules/Builder/features/ActivityAbout/ActivityAbout.tsx
```diff
@@ -210,10 +225,19 @@ export const ActivityAbout = () => {
...
+            onCustomChange,
+            defaultControllerValue,
+          }) => (
             <CheckboxController
               key={name}
               control={control}
+              defaultControllerValue={defaultControllerValue}
               name={name}
               label={label}
               disabled={disabled}
```
farmerpaul: Could this be simplified to:

```tsx
          (props) => <CheckboxController key={name} control={control} {...props} />,
```

src/shared/components/FormComponents/CheckboxController/CheckboxController.types.ts
farmerpaul: I'm wondering why we need to provide a default value on the `Controller` level? I've never actually done it that way with react-hook-form – always provided default values for the form at the `useForm()` step via its `defaults` parameter, before any `Controller` components are even rendered.

I delved into the `useForm` [call for the Builder](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/M2-7390/src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx#L70-L76), dug down a bit, and my impression is that new activities and flows have their default values decided [here](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/M2-7390/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx#L312-L360) and [here](https://github.com/ChildMindInstitute/mindlogger-admin/blob/feature/M2-7390/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx#L639-L646), respectively.

Could you set defaults that way, rather than making changes to `CheckboxController`?

farmerpaul (changes requested):
Hey @qiushihe, nice work, peer testing all passed. I did have a question about how we're assigning the default value, curious to hear your thoughts on this.

src/shared/components/FormComponents/CheckboxController/CheckboxController.types.ts
qiushihe: I think this may be one of those "it works either way" kind of thing. Because `defaultValue` is a prop on `Controller` and it evidently works. That said, I was in fact looking for a way to do the way you linked but I couldn't find any examples 😖. So I'll give this another stab and try to make this part consistently with how things are currently done elsewhere.

qiushihe: Also, I see the difference in approaches now. What I did was to tell the form/UI controller: "if there is no value for auto-assign, then use `true` as the default"; whereas the way we're currently doing things with what you linked is to create a new object with explicitly set default values.

qiushihe:
@farmerpaul I made the changes you requested. Please take another look. Thanks!

farmerpaul (approved):
Lookin' great, thanks @qiushihe!!

PR #1903, opened 2024-08-30
Revert 1896 fix/m2 7626 data export  export with lot of media

ChaconC:
Reverts #1896 M2-7626 as it introduced an unwanted behavior during export, causing a large number of files to be downloaded instead of a single archive.


mbanting (changes requested):
@ChaconC the PR being reverted #1896 only had 4 files changed. Is the intent to isolate reversion to just that PR? if so can `git revert b1ba62bcce560ea00117592938e06b9bbc6e5614` should do the trick.

PR #1904, opened 2024-08-30
Revert "fix: data export with large number of media files (M2-7626) (…

ChaconC:
Reverts https://github.com/ChildMindInstitute/mindlogger-admin/pull/1896 [M2-7626](https://mindlogger.atlassian.net/browse/M2-7626) as it introduced an unwanted behavior during export, causing a large number of files to be downloaded instead of a single archive.

This reverts commit b68679f0b908a5ca2c605b40941b72c690ba5245.

[M2-7626]: https://mindlogger.atlassian.net/browse/M2-7626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mbanting (approved):
Compared with #1896 and looks accurate

PR #1909, opened 2024-09-04
feat: [Subscales] Updates to Age/Gender Collection Screens (CAMHI) (M2-7587)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-7587](https://mindlogger.atlassian.net/browse/M2-7587)

This PR makes the following updates to the activity builder section of the Admin App:
- The subscales configuration screen now has a setting allowing the admin to choose the type of field that collects the age from the participant: a text field, or a select element
- Number validation has been added to the age text field, and the input length has been limited to 3 characters
- New copy has been added to the age and sex question screens according to the language specified in the ticket. This copy originally included a hyperlink to terms of service, but I've recommended it be removed.

#### Admin App

image

#### Web App

image

image

image

### 🪤 Peer Testing

- Create an activity and add a subscale configuration to it.
- Upload a lookup table for at least one of the subscales
- Confirm that the age field type setting is now visible at the bottom of the subscale configuration page
- Confirm that the items list now has `gender_screen` and `age_screen`  entries that look like this
  image
- Toggle the age field setting, and launch the activity in the participant web app
- Confirm that the age field type setting is being respected in the web app
- Confirm that the new copy is showing up on the age and sex question screens

### ✏️ Notes

N/A


src/modules/Builder/features/ActivitySettings/SubscalesConfiguration/SubscalesConfiguration.const.ts
```diff
@@ -77,11 +78,27 @@ export const genderItem = getNewActivityItem({
...
 });
+
+export const ageDropdownItem = getNewActivityItem({
+  allowEdit: false,
+  name: LookupTableItems.Age_screen,
+  question: t('ageQuestion'),
+  config: defaultNumberSelectionConfig,
+  responseType: ItemResponseType.NumberSelection,
+  responseValues: {
+    minValue: 1,
```
farmerpaul: The AC states a range of 0-50 🤷🏻‍♂️ 
```suggestion
    minValue: 0,
```

src/modules/Builder/features/ActivitySettings/SubscalesConfiguration/SubscalesConfiguration.tsx
```diff
@@ -181,6 +192,34 @@ export const SubscalesConfiguration = () => {
             noDataPlaceholder={t('noElementsYet')}
             data-testid={`${dataTestid}-elements-associated-with-subscales`}
           />
+          {hasLookupTable && (
+            <FormControl
+              sx={{ mt: theme.spacing(2), flexDirection: 'row', alignItems: 'center', gap: 1.6 }}
```
farmerpaul:
```suggestion
              sx={{ mt: 2, flexDirection: 'row', alignItems: 'center', gap: 1.6 }}
```

farmerpaul (approved):
LGTM, just noticed one tiny AC thing.

src/modules/Builder/features/ActivitySettings/SubscalesConfiguration/SubscalesConfiguration.const.ts
sultanofcardio: Oh you're right, but that feels a bit weird. Let me clarify that real quick

farmerpaul: Also FWIW, `0` is also an acceptable value when age is a text response type. Maybe the subject could be a baby less than 1 year old?

sultanofcardio: Confirmed that it should be 0. Thanks for catching this Paul 🙏

sultanofcardio: > Also FWIW, 0 is also an acceptable value when age is a text response type.

Ah yeah, I noticed that after you pointed this out so I would've had to address that if it was supposed to be 1. But I guess that's not necessary

AlejandroCoronadoN (approved):
It worked on the Admin panel and the code looks good.

PR #1915, opened 2024-09-06
fix: eslint config for prettier

qiushihe:

### 📝 Description

🔗 [Jira Ticket M2-7785](https://mindlogger.atlassian.net/browse/M2-7785)

This PR fixes eslint config for prettier.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

You can verify the effectiveness of these changes by:

1. Check out `develop`
2. Make some obvious ugly code
3. Run `lint:fix` -- this should have no effect
4. Check out the branch of this PR
5. Make the same obvious ugly code
6. Run `lint:fix` -- this should fix your ugly code

### ✏️ Notes

N/A

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,7 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "plugins": ["import", "react-hooks", "@typescript-eslint", "unused-imports"],
+  "plugins": ["import", "react-hooks", "@typescript-eslint", "unused-imports", "prettier"],
   "env": {
     "node": true
   },
@@ -9,7 +9,7 @@
     "eslint:recommended",
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
-    "prettier"
+    "plugin:prettier/recommended"
   ],
   "rules": {
     "arrow-body-style": ["error", "as-needed"],


farmerpaul:
I'm ambivalent about switching over to this approach. I did ask @sultanofcardio about the desired approach for integrating prettier back in February (he was the main one tasked with improving linting/styling consistency between the apps when we first started on the project). He said that although this change enables Prettier as an ESLint rule (making formatting issues become lint errors), Prettier discourages this approach and recommends running prettier separately.

It's also worth noting that we still lack consistency here among the other apps:
- The respondent Web App uses `plugin:prettier/recommended`
- The respondent Mobile App uses `prettier`

My only real preference would be to pick one and use it consistently across the 3 apps. Which approach we go with doesn't matter so much to me.

If we opt to use `prettier` instead of `plugin:prettier/recommended`, then you can make sure all your code is prettified on save in VS Code by setting prettier as your default formatter:

```json
  "editor.formatOnSave": true,
  "[javascript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[typescriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[json]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
```

sultanofcardio:
@farmerpaul I'm also indifferent about this tbh. I think if most of us prefer this, we should go ahead with this change. Just make sure we get some reviewers from pod 6.

Re consistency, I think that should be much easier to achieve at this point

farmerpaul (approved):
I'm fine with this change, so long as Pod 6 is cool with it? /cc @ChaconC

mbanting (approved):
I added @ChaconC as a reviewer too, but I'm good with this. I assume we'll make a similar update on [mobile](https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/dev/.eslintrc.js)?

qiushihe:
And just adding my thought in this:

First of all, for this part:
> Prettier discourages this approach and recommends running prettier separately.

I honestly don't understand why they would even recommend that. If someone want to do the "format on save" approach, they can still do that. They can still configure VScode or whatever to run `prettier` on save. That should have nothing to do with how the prettier-eslint integration is configured; nor should that have anything to do with the fact that `prettier` is integrated with `eslint` at all. I also feel adding `eslint` integration with `prettier` takes away nothing ('cause you can still run `prettier` separately), so I don't see any harm in the integration.

Secondly:
* If we decide we want to follow `prettier`'s recommendation and not integrating it with `eslint` ...
  * ... then we should just remove `eslint-plugin-prettier` and all related configuration
* If we decide we want to integration `prettier` with `eslint` ...
  * ... then we should configure such integration properly: The current configuration is simply incorrect. The correct value for the `extend` array is `plugin:prettier/recommended` (see: https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#configuration-legacy-eslintrc)

In conclusion, I think our current stance is that we want to continue to integrate `prettier` with `eslint`? So, if that's the case, then we should fix the incorrect configuration, hence this PR 💁‍♀️

I also recorded a loom saying basically the same thing: https://www.loom.com/share/75ca2108296e418d8221682be10db20a

sultanofcardio:
@qiushihe what you're seeing is not incorrect. Prettier is not currently set up as an ESLint plugin. Including `"prettier"` in the `extends` array is just making use of [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier), which is an ESLint config that turns off other ESLint rules that might conflict with Prettier. This PR enables Prettier as an ESLint plugin, and turns on the set of recommended ESLint rules that ship with [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) (including those from `eslint-config-prettier`).

I think we agree that this is fine

qiushihe:
~~So for this part:~~
> ~~Including "prettier" in the extends array is just making use of eslint-config-prettier, which is an ESLint config that turns off other ESLint rules that might conflict with Prettier.~~

~~Do you know where that’s documented in eslint’s or prettier’s documentations? I didn’t know that’s a thing.~~

nvm I found it: https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#installation

mbanting:
> @qiushihe what you're seeing is not incorrect. Prettier is not currently set up as an ESLint plugin. Including `"prettier"` in the `extends` array is just making use of [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier), which is an ESLint config that turns off other ESLint rules that might conflict with Prettier. This PR enables Prettier as an ESLint plugin, and turns on the set of recommended ESLint rules that ship with [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) (including those from `eslint-config-prettier`).
> 
> I think we agree that this is fine

Great explanation. Thank you @sultanofcardio

PR #1924, opened 2024-09-20
feat: Unassign Activities (M2-7277, M2-7547)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-7277](https://mindlogger.atlassian.net/browse/M2-7277): Unassign Activity or Flow
🔗 [Jira Ticket M2-7547](https://mindlogger.atlassian.net/browse/M2-7547): Unassign Tooltip/Copy for Auto-Assigned Activities

This PR adds the ability to unassign respondents and target subjects from assigned activities/flows on the Participant Details Page (PDP):

- Adds the "Unassign Activity" menu item to assigned activity/flow cards on the PDP
- Adds the "Unassigned" section to the bottom of the PDP to list unassigned activities/flows
- When "Unassign Activity" is clicked, displays the [Unassign dialogue](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=11807-162282&node-type=frame&t=z8cp3iGbC7PeNhEy-11) (if the activity/flow has only one assignment) or the [Unassign panel](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=11807-265109&node-type=frame&t=z8cp3iGbC7PeNhEy-11) (if the activity/flow has multiple assignments).
- Confirming the unassign action calls the new DELETE `/assignments/applet/{applet_id}` endpoint to remove the assignment(s) and re-queries the PDP

> [!NOTE]
> The [Unassign dialogue](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=12187-395534&node-type=frame&t=z8cp3iGbC7PeNhEy-11) & [Unassign panel](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=11807-276493&node-type=frame&t=z8cp3iGbC7PeNhEy-11) are based on the new Unassign flow from the [new (revised) PDP designs](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=11796-234679&node-type=frame&t=z8cp3iGbC7PeNhEy-11). However, the changes made to the PDP itself in this PR to identify unassigned activities (adding the "Unassigned" section) are based on the [original designs](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=9135-121172&t=z8cp3iGbC7PeNhEy-4). Actual revisions to the PDP to use the newest designs were out of scope for this ticket, and will be happening subsequently this sprint.

> [!TIP]
> Related changes have been grouped by commit, with detailed descriptions of why certain changes were made in commit messages. It will probably be easier to review [commit-by-commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1924/commits).

### 📸 Screenshots

https://github.com/user-attachments/assets/a99d40f2-9056-49bb-91f7-1d18c25633cc

### 🪤 Peer Testing

> [!NOTE] 
> This feature depends on the `enableActivityAssign` feature flag. This should already be enabled in dev/local environments. To control its value manually, add `features.enableActivityAssign = [true|false]` prior to the `return` statement [here in `useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/e6ff173fc6aaf31da3e81d92ab4f59caa8eef6da/src/shared/hooks/useFeatureFlags.ts#L25-L26).

#### Test with feature flag enabled ✅

1. Add `features.enableActivityAssign = true` to [`useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/e6ff173fc6aaf31da3e81d92ab4f59caa8eef6da/src/shared/hooks/useFeatureFlags.ts#L25-L26).
2. Please refer to the AC of [M2-7277](https://mindlogger.atlassian.net/browse/M2-7277) and [M2-7547](https://mindlogger.atlassian.net/browse/M2-7547) to ensure behaviour is in alignment.

#### Test with feature flag disabled 🚫

1. Add `features.enableActivityAssign = false` to [`useFeatureFlags`](https://github.com/ChildMindInstitute/mindlogger-admin/blob/e6ff173fc6aaf31da3e81d92ab4f59caa8eef6da/src/shared/hooks/useFeatureFlags.ts#L25-L26).
2. Load the PDP and ensure there are is regression in behaviour.

### ✏️ Notes

- Refactored the internal banners hook previously used by `ActivityAssignDrawer` to a shared hook called `useModalBanners`, which gives any component (drawer or otherwise) the convenience to render its custom set of banners internally.
- Hid a useless Search field on the PDP that was never actually implemented and should have been commented out with a TODO link long ago.
- Removed some console warnings by avoiding use of `:first-child` in CSS.

[M2-7277]: https://mindlogger.atlassian.net/browse/M2-7277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Dashboard/components/ActivityUnassignDrawer/ActivityUnassignDrawer.schema.ts
```diff
@@ -0,0 +1,8 @@
+import * as yup from 'yup';
+
+export const useActivityUnassignFormSchema = () =>
+  yup
+    .object({
+      selectedAssignments: yup.array().min(1).required(),
+    })
+    .required();
```
sultanofcardio: Could this be a static object instead of a function?

src/modules/Dashboard/components/ActivityUnassignDrawer/ActivityUnassignDrawer.utils.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <Trans i18nKey="activityUnassign.confirmationBodyMultipleRespondents">
+      <strong>
+        <>{{ selectedCount: selected.length }}</>
+      </strong>
+      <strong></strong>
+      <strong>
+        <>{{ target }}</>
+      </strong>
+    </Trans>
+  );
```
sultanofcardio: It's not mentioned in the AC, but there's a fourth scenario that I think we should consider here: the case where we have multiple respondents and multiple subjects. Consider the scenario where two participants are both assigned to complete the same activity about each other. We can end up with this

image

In this case, the content of the dialog depends on which respondent is selected first.

https://github.com/user-attachments/assets/a8cd3559-1af0-416f-9c64-ab3b1ec70da6

Given that it's not specified in the AC, we may want to raise this first but I think something along these lines makes sense:
> If you proceed, the [#] selected Respondents will no longer be required to complete this activity about the [#] selected Subjects. Any data previously collected will not be lost.

sultanofcardio: Also I think using `participantContext` here is blocking a valid **Single subject, multiple respondents** scenario.

image

Here there are two subjects, but they are the same. The dialog ends up treating them as two different subjects

image image

Same as above, the respondent is determined based on which one is selected first

src/modules/Dashboard/components/ActivityUnassignDrawer/ActivityUnassignDrawer.types.ts
```diff
@@ -0,0 +1,27 @@
...
+const ParticipantContext = ['respondent', 'target'] as const;
+export type ParticipantContext = (typeof ParticipantContext)[number];
+
+export type ActivityUnassignDrawerProps = Pick<DrawerProps, 'open'> & {
+  onClose: (shouldRefetch?: boolean) => void;
+  appletId?: string;
+  activityOrFlow?: AssignedActivity | AssignedHydratedActivityFlow;
+  /**
+   * Whether initiating Unassign from the context of the participant being the respondent or the
+   * the target subject.
```
sultanofcardio:
```suggestion
   * Whether initiating Unassign from the context of the participant being the respondent or
   * the target subject.
```

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
```diff
@@ -73,27 +77,34 @@ export function useFlowGridMenu({
...
-          isDisplayed: canAssign,
+          isDisplayed: canAssign && !autoAssign,
         },
+        {
+          'data-testid': `${testId}-flow-unassign`,
+          action: () => {
+            if (flowId) onClickUnassign?.(flowId);
+          },
+          icon: <Svg id="clear-calendar" />,
+          title: t('unassignActivity'),
```
sultanofcardio: Should this say "Unassign Flow" instead? Given that the edit menu entry says "Edit flow" I'm thinking we should be consistent (we can also change the assign menu entry above as well)

src/resources/app-en.json
```diff
@@ -72,6 +72,23 @@
     "emailsSent": "Emails have been sent",
     "done": "Done"
   },
+  "activityUnassign": {
+    "title": "Unassign Activity",
```
sultanofcardio: Should the title of the panel be smart about what's it's showing? Currently it says Unassign Activity even if the thing in question is a flow

image

src/resources/app-en.json
```diff
@@ -72,6 +72,23 @@
...
+    "selectAssignments": "Select Assignments",
+    "respondent": "Respondent",
+    "subject": "Subject",
+    "subjectSelf": "Self",
+    "unassign": "Unassign",
+    "confirmationTitle": "Unassign ‘{{name}}’",
+    "confirmationButton": "Unassign",
+    "confirmationBodySingle": "If you proceed, <0>{{respondent}}</0> will no longer be required to complete this activity about <0>{{target}}</0>. Any data previously collected will not be lost.",
+    "confirmationBodyMultipleTargets": "If you proceed, <0>{{respondent}}</0> will no longer be required to complete this activity about the <0>{{selectedCount}} selected Subjects</0>. Any data previously collected will not be lost.",
+    "confirmationBodyMultipleRespondents": "If you proceed, the <0>{{selectedCount}} selected Respondents</0> will no longer be required to complete this activity about <0>{{target}}</0>. Any data previously collected will not be lost.",
```
sultanofcardio: Same thing with the copy here. It always uses the word activity, even when dealing with a flow. Being dynamic about this is arguably minutia so lmk if you think this is worth making it dynamic

sultanofcardio (commented):
This is a quite a large change so I haven't had a chance to go over all the code yet. From a functionality POV I think we're most of the way there, so great job on that as usual 👌 

This is what I've been able to review so far

src/modules/Dashboard/components/ActivityUnassignDrawer/ActivityUnassignDrawer.schema.ts
farmerpaul: Yup, totes! I was just following the same pattern used elsewhere that react-hook-form is used, but this can totally be static and probably doesn't even need to be in its own file.

src/modules/Dashboard/components/ActivityUnassignDrawer/ActivityUnassignDrawer.utils.tsx
farmerpaul: Good eye, Kenroy.

**Re first comment:**

I should have called this out in the PR description. Here's the situation:

It's only in the context of this now-outdated PDP design that it's possible to initiate Unassign where there may be a combination of multiple respondents _and_ multiple target subjects in the list of assignments. It will not be possible in the new designs, because each "Unassign" trigger is from the context of a subject as the respondent, _or_ the subject as the target subject – no longer both.

However, now that you point it out, I think it's prudent to be flexible anyway, even if won't be possible with my upcoming PDP revisions. For example, maybe eventually (and perhaps soon even) we'll be able to perform "Unassign" outside of the context of a given subject (say, for an activity or flow in general), in which case it should be able to handle this case.

**Re second comment:**

I think the reason it's doing this is because `respondent` is hardcoded for the `participantContext` prop, so currently, it's always showing the wrong copy when there are multiple subjects in this temporary version of the PDP. I'll make more resilient so the copy actually reflects the list of assignments, it'll be a simple enough fix.

src/modules/Dashboard/components/FlowGrid/FlowGrid.hooks.tsx
farmerpaul: Agreed :+1:

src/resources/app-en.json
farmerpaul: I mean, that's what design has, even in their example where it's a Flow that was selected. It's the same with the "Assign Activity" panel – but in that case, it's unknown which activity/flow or activities/flows you may be wanting to assign. It's only predictable which word to choose in the context of Unassign. Maybe I'll ask in Figma.

src/resources/app-en.json
farmerpaul: I think it's worth it!

src/resources/app-en.json
farmerpaul: Got approval for this change!

sultanofcardio (approved):
Great work Paul

PR #1928, opened 2024-09-24
fix: Review and fix banner messages to display accordingly (M2-7580)

LashaunnaS:

### 📝 Description

🔗 [Jira Ticket M2-7580](https://mindlogger.atlassian.net/browse/M2-7580)

- QA reported an inconsistency with how the assign activity banners display w/in the assign activity side drawer.
- The issue seemed to stem from an `if` statement short circuiting prior to a more immediate condition being returned.

Changes include:

- Display the correct banner message when assigning an activity using the Activity dataviz/summary card for a Limited Accounts.

### 📸 Screenshots

Before:

https://github.com/user-attachments/assets/10434cad-4c4d-496c-a114-a3e18347f608

After:

https://github.com/user-attachments/assets/3955da65-541a-43ba-be77-353295448ad7

### 🪤 Peer Testing

- Please view reproduction steps in [Ticket M2-7580](https://mindlogger.atlassian.net/browse/M2-7580)

### ✏️ Notes

- Source of truth for banner rendering logic can be found [M2-7135](https://mindlogger.atlassian.net/browse/M2-7135)

[M2-7135]: https://mindlogger.atlassian.net/browse/M2-7135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
+++ b/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
@@ -386,12 +386,12 @@ export const ActivityAssignDrawer = ({
             hasActivity: activityIds.length || flowIds.length,
             isTeamMember,
           });
-        } else if (activityIds.length || flowIds.length) {
-          addBanner('ActivityAutofillBanner');
         } else if (assignments[0]?.targetSubjectId) {
           addBanner('SubjectAutofillBanner', {
             hasActivity: activityIds.length || flowIds.length,
           });
+        } else if (activityIds.length || flowIds.length) {
+          addBanner('ActivityAutofillBanner');
         }
       }, 300);
     } else {
--- a/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
+++ b/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
@@ -386,12 +386,12 @@ export const ActivityAssignDrawer = ({
             hasActivity: activityIds.length || flowIds.length,
             isTeamMember,
           });
+        } else if (activityIds.length || flowIds.length) {
+          addBanner('ActivityAutofillBanner');
         } else if (assignments[0]?.targetSubjectId) {
           addBanner('SubjectAutofillBanner', {
             hasActivity: activityIds.length || flowIds.length,
           });
-        } else if (activityIds.length || flowIds.length) {
-          addBanner('ActivityAutofillBanner');
         }
       }, 300);
     } else {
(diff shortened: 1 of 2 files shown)


src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
```diff
@@ -386,12 +386,12 @@ export const ActivityAssignDrawer = ({
...
             isTeamMember,
           });
+        } else if (activityIds.length || flowIds.length) {
+          addBanner('ActivityAutofillBanner');
         } else if (assignments[0]?.targetSubjectId) {
           addBanner('SubjectAutofillBanner', {
             hasActivity: activityIds.length || flowIds.length,
           });
-        } else if (activityIds.length || flowIds.length) {
-          addBanner('ActivityAutofillBanner');
```
farmerpaul: This isn't quite doing the trick. The issue as reported in the bug description is that the message `The Participant & Activity have been auto-filled. 1 row needs attention to continue.` is what's desired when there is an activity/flow defined, _and_ a target subject. Your change above actually causes the message `Your Activity was auto-filled, add Respondents to continue.` to be shown under this condition now, which isn't quite right. So I think the above change should be reverted (I think the original logic was correct).

Looking at the code as it was before the change, that condition _should_ have been captured by the `addBanner('SubjectAutofillBanner'…)` call above, with `hasActivity: true`. But if you dig into `SubjectAutofillBanner`, you'll see that it's not even handling the `hasActivity` prop at all – that's a flop on my part!

That component just needs to be tweaked so that the `subjectActivityAutofill` language key is used when `hasActivity` is true, and otherwise behave the same as before:

```tsx
      {hasActivity ? t('subjectActivityAutofill') : t('subjectAutofill')}
```

farmerpaul (changes requested):
Just identified a bit of a discrepancy between the AC and your implementation – let me know if you have any questions!

**Edit:** Also, if you could update the PR title to more closely follow our [Conventional Commits format](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/7503873/Configuration+Management+Plan#Pull-Requests), that'd be swell. In [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), the prefix for bug fixes is simply `fix: …`, and as a habit, we've been adding the Jira ticket number at the end in round parentheses: `… (M2-7580)` I realize this may be different from past projects. 🙏🏻

src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
LashaunnaS: Thanks for catching that, I did get those mixed up.

farmerpaul (approved):
Thank you!! Ready for QA.

PR #1935, opened 2024-10-04
feat: Configure Report To Select Scores or Raw Scores (CAMHI) (M2-7738)

sultanofcardio:

### 📝 Description

> [!NOTE]
> It may be helpful to look at [TL;DR on Subscales, T Scores, and Lookup Tables](https://www.notion.so/cmi-product/TL-DR-on-Subscales-T-Scores-and-Lookup-Tables-32c820fdc84e4d4c8bb66917b3c8381a) before reviewing this PR if you're unfamiliar with subscales

🔗 [Jira Ticket M2-7738](https://mindlogger.atlassian.net/browse/M2-7738)

This PR adds support for linking subscales to report scores, allowing them to use T-Scores from a subscale's lookup table for their calculations. 

> [!NOTE]
> Only subscales that contain both a lookup table and at least one non-subscale item can be linked to a report score. It is valid for subscales to contain _only_ nested subscales, but those are not valid for linking

To link a subscale to a report score, the admin must first define an eligible subscale, and then indicate the type of score to be used on the report

image

> This subscale has non-subscale items and a lookup table

image

> Note that the _Score Calculation Type_ and the _Items used to calculate score_ are fixed to match those of the selected subscale and cannot be edited

### 📸 Screenshots

In addition to the screenshots above, here are some other cases to consider

#### No eligible subscales

When there are no eligible subscales, the radio group does not appear and the score type is fixed to raw score

image

#### Raw score by default

When there are eligible subscales, raw score is selected by default and the _Score Calculation Type_ and the _Items used to calculate score_ remain editable

image

#### Tooltips

Tooltips are included to tell the user about each score type

image image

### 🪤 Peer Testing

> [!NOTE]
> Some scenarios may require the API modifications from https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1616 to be tested

Create an activity and use it to test the following conditions

#### Hidden by default

1. Ensure the activity does not contain any eligible subscales
2. Create a report score
3. **Confirm** that the radio group is not visible

#### Tooltip copy

1. Create an eligible subscale
2. Create a report score and set the score type to "Score"
3. Confirm the copy of each tooltip on the info icons match the acceptance criteria

#### Default score type

1. Ensure the activity does not contain any eligible subscales
5. Create a report score and fill in the required fields
6. Save the applet
7. **Confirm** that the `scoringType` field in the request body is `raw_score`
8. Create an eligible subscale
9. **Confirm** that the radio group is now visible on the report score and raw score is selected

#### Only eligible subscales

1. Create at least one of each of the following subscales
    - One with a lookup table and non-subscale items (eligible)
    - One with a lookup table and only subscale items (ineligible)
    - One without a lookup table (ineligible)
2. Create a report score and set the score type to "Score"
4. **Confirm** that the subscale dropdown list only contains the eligible subscale

#### Link to subscales page

1. Create an eligible subscale
2. Create a report score and set the score type to "Score"
3. Select the eligible subscale from the dropdown list
4. Click the "View Subscale Configuration" button
5. Confirm navigation to the "Subscales Configuration" screen

> [!NOTE]
> The selected subscale can't be directly navigated to, so the one that appears after navigation may be a different one

#### Score range and items

1. Create an eligible subscale
2. Create a report score and set the score type to "Score"
3. Select the eligible subscale from the dropdown list
4. **Confirm** that the _Possible Range of Scores_ is updated to match the min and max scores between the scores on each item and those in the subscale lookup table
5. **Confirm** that the _Items used to calculate score_ are no longer editable and match the items in the selected subscale

#### Linked subscale is saved

1. Create an eligible subscale
2. Create a report score and set the score type to "Score"
3. Select the eligible subscale from the dropdown list
4. Fill in the rest of the required fields
5. Save the applet
6. **Confirm** that the `subscaleName` field in the request body contains the name of the selected subscale

> [!WARNING]
> If https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1616 is not yet complete, then the data may not be persisted

#### Only eligible items

1. Create an eligible subscale that also contains another subscale
2. Create a report score and set the score type to "Score"
3. Select the eligible subscale from the dropdown list
4. Fill in the rest of the required fields
5. Save the applet
6. **Confirm** that the success banner is shown and the `itemsScore` field in the request body contains only the name of the non-subscale item(s)

> [!WARNING]
> If https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1616 is not yet complete, then the data may not be persisted

##### Name change

1. Create an eligible subscale
2. Create a report score and set the score type to "Score"
4. Select the eligible subscale from the dropdown list
6. Fill in the rest of the required fields
8. Change the name of the selected subscale
9. Without navigating away from the subscales configuration screen, Save the applet
10. **Confirm** that the `subscaleName` field in the request body contains the name of the selected subscale

> [!WARNING]
> If https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1616 is not yet complete, then the data may not be persisted

##### Remove lookup table

1. Repeat steps 1-4 from the name change scenario
2. Remove the lookup table from the selected subscale
3. Without navigating away from the subscales configuration screen, Save the applet
4. Confirm that an error is shown and the applet fails to be saved

##### Delete selected subscale

1. Repeat steps 1-4 from the name change scenario
2. Delete the selected subscale
3. Without navigating away from the subscales configuration screen, Save the applet
4. Confirm that an error is shown and the applet fails to be saved

### ✏️ Notes

1. Automated tests have not yet been added to account for the updated page UI
2. I've updated the `onChange` prop on the `RadioGroupController` and the `onCustomChange` prop in the `InputController` to accept the original `onChange` function from `react-hook-form` as a parameter, so that it may optionally be called by the provider


farmerpaul (commented):
Hey @sultanofcardio, I've run through the peer testing so far. All tests pass according to the expectations you've outlined in each of the scenarios, however I did discover a client-side error happening after each of two of these scenarios:

- **Remove lookup table**
- **Delete selected subscale**

In each of these cases, although I'm correctly led to fix the validation error on the Scores & Reports tab, once I fix it (by selecting a different eligible subscale, for instance), I'm unable to save changes. Clicking **Save & Publish** after I've addressed this validation error results in this client-side error, and trying to save from then on doesn't seem work until I reload the browser:

CleanShot 2024-10-07 at 14 41 30@2x

sultanofcardio:
Thanks for finding this @farmerpaul 🙏  It should be fixed with https://github.com/ChildMindInstitute/mindlogger-admin/commit/c30a70cc556bb960763002641f85463247e6586a

farmerpaul (approved):
That error is happening less now! However, I can still cause it to happen, but thankfully, it's not related to your work – it's a pre-existing bug reproducible in `develop`. It seems that when you have a subscale that references another subscale, and you delete the referenced subscale (but not the one that was referencing it) and try to Save & Publish, you get that same `TypeError: Cannot read properties of undefined (reading 'name')` and you can no longer save your changes. If you think it's too much effort to come up with a fix for this, I'll just log this as a bug in Jira. But if it's a quick fix, would be nice to take care of it.

I had a look through the code and nothing seems off to me, so happy to approve!

mbanting (approved):
Peer tested this alongside peer testing M2-8097.
Code LGTM. Great job @sultanofcardio

PR #1939, opened 2024-10-08
feat: Add "Remove Phrase" confirmation dialog (M2-7163)

JustinNusca:
### 📝 Description

🔗 [M2-7163](https://mindlogger.atlassian.net/browse/M2-7163): [FE][Phrase Builder] Delete existing phrases

This PR updates the behaviour for removing an existing phrase from a "Phrase Builder" activity item.

Now, when the "Remove Phrase" button is pressed, a confirmation dialog appears, which the user must acknowledge before the phrase can be removed.

Previously, the "Remove Phrase" button was disabled if only a single phrase existed in the item. Now, the button is enabled, and pressing it shows the same confirmation dialog. When confirmed, the phrase is reset to the "default" state (ie, it contains empty fields for `sentence`, `item_response`, and `sentence` values, in that order).

### 📸 Screenshots

| Before | After |
|-|-|
| before | after |

### 🪤 Peer Testing

From within the applet builder, for an applet with one or more activities:

* Create a new Item, and select an Item Type of "Phrase Builder"
* Create multiple Phrases for the new Item. (Optional: Save your changes to the applet before proceeding).
* Press the delete button to remove a phrase, and then press "Yes, Remove" in the confirmation dialog.
    * If there are multiple remaining phrases, the phrase will be removed.
    * If there is only one remaining phrase, that phrase's content will be reset to default.
* After removing phrases, pressing save (while the edited item is in a valid state) will persist changes.

[M2-7163]: https://mindlogger.atlassian.net/browse/M2-7163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Builder/components/RemovePhrasePopup/RemovePhrasePopup.tsx
+++ b/src/modules/Builder/components/RemovePhrasePopup/RemovePhrasePopup.tsx
@@ -1,0 +1,31 @@
+import { Button } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+
+import { Modal, ModalProps } from 'shared/components';
+import { StyledFlexAllCenter, StyledFlexTopStart } from 'shared/styles';
+
+export const RemovePhrasePopup = ({
+  onRemove,
+  ...otherProps
+}: Omit<ModalProps, 'title' | 'children'> & { onRemove?: () => void }) => {
+  const { t } = useTranslation('app');
+
+  return (
+    <Modal
+      footer={
+        <StyledFlexAllCenter sx={{ width: '100%' }}>
+          <Button color="error" onClick={onRemove} variant="contained">
+            {t('phrasalTemplateRemovePopup.confirm')}
+          </Button>
+        </StyledFlexAllCenter>
+      }
+      hasActions={false}
+      title={t('phrasalTemplateRemovePopup.title')}
+      {...otherProps}
+    >
+      <StyledFlexTopStart as="p" sx={{ px: 3.2, m: 0, pt: 2.4 }}>
+        {t('phrasalTemplateRemovePopup.areYouSure')}
+      </StyledFlexTopStart>
+    </Modal>
+  );
+};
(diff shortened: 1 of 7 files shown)


src/modules/Builder/components/RemovePhrasePopup/RemovePhrasePopup.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <Modal
+      footer={
+        <StyledFlexAllCenter sx={{ width: '100%' }}>
+          <Button color="error" onClick={onRemove} variant="contained">
+            {t('phrasalTemplateRemovePopup.confirm')}
+          </Button>
+        </StyledFlexAllCenter>
+      }
+      hasActions={false}
```
farmerpaul: It's not necessary to do a custom footer – the base props centre-align the only button by default:

```suggestion
      onSubmit={onRemove}
      buttonText={t('phrasalTemplateRemovePopup.confirm')}
      submitBtnColor="error"
```

src/modules/Builder/components/RemovePhrasePopup/RemovePhrasePopup.tsx
```diff
@@ -0,0 +1,31 @@
...
+      title={t('phrasalTemplateRemovePopup.title')}
+      {...otherProps}
+    >
+      <StyledFlexTopStart as="p" sx={{ px: 3.2, m: 0, pt: 2.4 }}>
+        {t('phrasalTemplateRemovePopup.areYouSure')}
+      </StyledFlexTopStart>
```
farmerpaul: Use the standard content container for popups, `StyledModalWrapper` (in `shared/styles`), which applies the same padding app-wide:
```suggestion
      <StyledModalWrapper>
        {t('phrasalTemplateRemovePopup.areYouSure')}
      </StyledModalWrapper>
```

farmerpaul (approved):
Looks good, a couple suggestions, but preapproving! ✅ 

Also, I looked at the ticket in Jira and it includes a screenshot that isn't related to this work – the Trash icon for individual phrase fields. Phrase field deletion was completed in M2-7183, not in this ticket, so I've gone ahead and deleted that screenshot from M2-7163 so as to minimize confusion for QA.

PR #1940, opened 2024-10-10
feat: Add Activity Status Tags (M2-7825)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-7825](https://mindlogger.atlassian.net/browse/M2-7825)

This PR adds the `ActivityStatusChip` component used to render the Active/Inactive/Hidden/Unavailable tag beside Activities and Flows on the PDP screen.

### 📸 Screenshots

| CleanShot 2024-10-07 at 17 14 09@2x | CleanShot 2024-10-07 at 17 13 49@2x |

### 🪤 Peer Testing

> [!TIP]

1. For a given participant, ensure you have activities or flows created with these characteristics:
    - One that is manually assigned to the participant as the subject
    - One that is auto-assigned to the participant as the subject
    - One that is not assigned to the participant, but has submissions for them as the subject (e.g., via Take Now)
    - One that is hidden
2. From that participant's PDP, ensure each of the activities as the appropriate status tag based on the AC of the [ticket](https://mindlogger.atlassian.net/browse/M2-7825).


farmerpaul:
@sultanofcardio FYI I requested that @kattyode test this PR ahead of getting any engineers' reviews, since we're going to losing her help soon and time is running out. This is a pretty simple PR that I don't expect a ton of engineer feedback on.

JustinNusca (approved):
👍 LGTM — looks correct and is showing up in the appropriate state for the described conditions.

sultanofcardio (approved):

src/modules/Dashboard/features/Participant/Assignments/ActivityListItem/ActivityListItem.tsx
```diff
@@ -25,6 +25,7 @@ export const ActivityListItem = ({
       </StyledActivityThumbnailContainer>
       <StyledActivityName>{name}</StyledActivityName>
       {isFlow && <FlowChip size="small" />}
+      <ActivityStatusChip status={status} />
```
sultanofcardio, line 28: The notes to engineering on the ticket says:

> Please wrap this in a new feature flag. the  enableActivityAssign feature flag will be disabled once Assign R1 is ready.

I'm not sure if this is accurate, but if it is then we'd need a new feature flag here. We should clarify that and have it removed from the description if it's not applicable

farmerpaul: This comment was added indiscriminately to all tasks related to PDP revisions by product, without an understanding of the redundancy of doing so from an engineering perspective. The entire revised PDP screen is already behind this feature flag, so there's no extra work needed to ensure that the Activity Status tag portion of this work is also behind the flag – it already is. 🙂

I'll clarify this in the task description, and make the same change to the remaining tasks where this statement is adding confusion.

farmerpaul: Ah, looking more closely I think this work was originally intended for MI Assign R2 – but it got pulled into R1. So it will assume the same `enableActivityAssign` feature flag as the rest of the work I've been doing.

PR #1941, opened 2024-10-11
feat: Add By Participant view (M2-7798)

farmerpaul:
- [ ] ~~Tests for the changes have been added~~ **(tests to be added in a separate task)**

### 📝 Description

🔗 [Jira Ticket M2-7798](https://mindlogger.atlassian.net/browse/M2-7798)

> [!NOTE]
> I opted to target `develop` for this PR in the interests of making an Amplify environment available to QA, but that means this PR also includes the commit from the Activity Status Tags branch (see PR #1940), which it's dependent on. **Please consider only the changes in the [second commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1941/commits/a1656b18649e5a009ecc605e8b3115faaed4c898) during code review.**
>
> I've also moved M2-7798 to `To Be Tested` this morning in the interests of getting QA's attention, even though this is awaiting 👀 from engineering. I'll move it back to `Under Code Review` if it passes QA and engineers haven't had a chance to review yet.

This PR adds the By Participant view to the PDP, including:

- Support to call new API endpoint for By Participant view, `/activities/applet/${appletId}/respondent/${subjectId}`
- Empty states for By Participant view (full account, limited account)
- Adjustments to `ActivityListItem` component to support the new view

Also fixed some menu item logic to detect assignments correctly from the new view and handle disabled menu items properly for hidden activities/flows.

#### Full Account/Team Member Empty State

CleanShot 2024-10-09 at 15 06 52@2x

#### Limited Account Empty State

CleanShot 2024-10-09 at 15 04 11@2x

#### List of Activities/Flows for which Participant is Respondent

CleanShot 2024-10-09 at 15 04 31@2x

### 🪤 Peer Testing

> [!TIP]

1. For a given Participant, ensure you have activities or flows created with these characteristics:
    - One that is manually assigned to the Participant as the Respondent
    - One that is auto-assigned to the Participant as the Respondent
    - One that is not assigned to the Participant, but has submissions for them as the Respondent (e.g., via Take Now, or via a previous assignment that as since been removed)
    - One that is hidden
2. From that Participant's PDP, ensure each of the activities/flows is listed in the By Participant view according to designs, taking into account the out-of-scope portions of the design as described in the AC of the [ticket](https://mindlogger.atlassian.net/browse/M2-7798).
3. Create a new Limited Account and a new Full Account, and ensure there are no activities/flows that are auto-assigned.
4. Open the PDP > By Participant view for the Limited Account, and ensure the Limited Account empty state is shown.
5. Open the PDP > By Participant view for the Full Account, and ensure the Full Account empty state is shown.


src/modules/Dashboard/features/Participant/Assignments/ByParticipant/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,27 @@
...
+        </StyledHeadline>
+      </StyledFlexColumn>
+      {!isLimitedAccount && (
+        <Button variant="contained" color="primary" onClick={onClickAssign} sx={{ mt: 2.4 }}>
+          {t('assignActivityButton')}
+        </Button>
+      )}
+    </StyledFlexAllCenter>
+  );
+};
```
JustinNusca: This is almost a complete duplicate of the existing `EmptyState` component for the other tab. 

It's not the biggest deal, especially because there are just two usages currently, but do you think it would be easier to generic-ise that one for re-use here, by modifing the props to pass in a headline, lede, and buttonProps?

sultanofcardio (approved):

src/modules/Dashboard/features/Participant/Assignments/ActivityListItem/ActivityListItem.tsx
```diff
@@ -3,18 +3,18 @@ import {
...
 import { StyledActivityListItem, StyledActivityName } from './ActivityListItem.styles';
 
 export const ActivityListItem = ({
-  activityOrFlow: { isFlow, images, name },
+  activityOrFlow: { isFlow, images, name, status },
   onClick,
   children,
 }: ActivityListItemProps) => (
-  <StyledActivityListItem as="button" onClick={onClick}>
+  <StyledActivityListItem as={onClick ? 'a' : 'div'} onClick={onClick}>
```
JustinNusca, line 17: I don't think `<a>` is the appropriate element to use here — we're passing an `onClick`, not an `href`.

Because this will render an anchor tag without an `href` attribute, which will not receive focus and users who aren't using a pointer-type device will be unable to press on the list item.

I'm _guessing_ you went this route, rather than keeping it as a button, since nesting buttons isn't valid HTML and browsers will try to correct this. Technically, interactive elements inside an anchor tag are _also_ invalid, even if browsers are more forgiving with how they'll render this. The HTML police might not be coming around to investigate, but it's something to keep in mind.

I haven't looked in detail into how you've adapted these into accordions, which likely complicates this further, but I wonder if we should just move the `onClick` from the root node to the first child, which wraps all of the static content on the left, and avoid nesting the buttons passed in as children on the right. This also avoids misclicks — missing a button by a few pixels and navigating to the specific page, rather than opening the action menu, which happened to me at least a few times while testing 😅.

farmerpaul: I hear ya. I wasn't concerned that the wrapper wasn't accessible because its click action is just a duplicate of the action of the View Data button, which is accessible; avoiding having the container gain keyboard focus makes navigating by keyboard quicker and more intuitive (one less redundant tab press). But I didn't fully clock that issue with nested interactive elements – and in general, I'd like to keep the HTML police on my good side! Especially as violations increase the likelihood of eventually polluting the DevTools console with annoying warnings.

Will come up with a fix!

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/EmptyState/EmptyState.tsx
farmerpaul: Done!

src/modules/Dashboard/features/Participant/Assignments/ActivityListItem/ActivityListItem.tsx
farmerpaul: It will be easier to implement this as part of #1943 due to the overlap between the two PRs around card clickability. So I'll merge this as-is so as to avoid a lot of code conflict wrangling.

PR #1943, opened 2024-10-11
feat: Add By Participant Expanded Views (M2-7921)

farmerpaul:
- ~~[ ] Tests for the changes have been added~~ **(unit tests to be added in separate task)**

### 📝 Description

🔗 [Jira Ticket M2-7921](https://mindlogger.atlassian.net/browse/M2-7921)

This PR adds the expanded Subjects view to each Activity/Flow on the PDP's new By Participant tab. It includes these changes:

- Adding API support for the `/subjects/respondent/{subject_id}/activity-or-flow/{activity_or_flow_id}` endpoint, consumed by this expanded view
- Adding the `ExpandedView` sub-component of the `ByParticipant` screen component
- Adjusting the PDP's generic `ActivityListItem` component to be able to support custom expanded view content
- In the `ByParticipant` component, passing an instance of `ExpandedView` to each `ActivityListItem`, and handling fetching/refetching of data from the new endpoint to populate the `ExpandedView` and handle loading states gracefully

### 📸 Screenshots

https://github.com/user-attachments/assets/c7d0207c-7397-44da-842d-3cd97efea4d8

### 🪤 Peer Testing

> [!TIP]

Review the AC of the [Jira ticket](https://mindlogger.atlassian.net/browse/M2-7921) to ensure behaviour of expanded view is in alignment. Specifically be sure to test this screen where:

- The Participant whose PDP you're on is manually assigned as a respondent to activities/flows about either themselves or other Participants
- The Participant whose PDP you're on is auto-assigned to certain activities/flows
- The Participant whose PDP you're on is **not** assigned to certain activities/flows, but has submitted answer data as a respondent in the past, about either themselves or other Participants
- Ensure that after creating assignments or removing assignments by clicking Assign/Unassign from any of the action menus on this screen causes data to be refreshed so that the screen is always up-to-date

### DIFF
--- a/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
+++ b/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
@@ -72,7 +72,7 @@ const ByParticipant = () => {
 
     // Refresh target subject data for any expanded views
     Object.entries(expandedViewsData).forEach(([id, viewData]) => {
-      if (viewData) handleRefetchExpandedView(id);
+      if (viewData) void handleRefetchExpandedView(id);
     });
   }, [handleRefetchActivities, expandedViewsData, handleRefetchExpandedView]);
 
@@ -103,7 +103,7 @@ const ByParticipant = () => {
       if (!respondentSubject) return;
 
       if (isExpanded) {
-        handleRefetchExpandedView(activityOrFlowId);
+        void handleRefetchExpandedView(activityOrFlowId);
       } else {
         // If expanded view is closed, remove data to free up memory and minimize refetches
         // (after delay to account for transition)
--- a/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
+++ b/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
@@ -28,7 +28,7 @@ const ByParticipant = () => {
   const isLoadingRespondentSubject = useSubjectStatus() !== 'success';
   const { result: respondentSubject } = useSubject() ?? {};
   const [expandedViewsData, setExpandedViewsData] = useState<
-    Record<string, TargetSubjectsByRespondent | undefined>
+    Record<string, TargetSubjectsByRespondent>
   >({});
   const [expandedViewsLoading, setExpandedViewsLoading] = useState<Record<string, boolean>>({});
 
@@ -71,9 +71,7 @@ const ByParticipant = () => {
     handleRefetchActivities();
 
     // Refresh target subject data for any expanded views
-    Object.entries(expandedViewsData).forEach(([id, viewData]) => {
-      if (viewData) void handleRefetchExpandedView(id);
-    });
+    Object.entries(expandedViewsData).forEach(([id]) => void handleRefetchExpandedView(id));
   }, [handleRefetchActivities, expandedViewsData, handleRefetchExpandedView]);
 
   const {
@@ -108,7 +106,11 @@ const ByParticipant = () => {
         // If expanded view is closed, remove data to free up memory and minimize refetches
         // (after delay to account for transition)
         setTimeout(() => {
-          setExpandedViewsData((prev) => ({ ...prev, [activityOrFlowId]: undefined }));
+          setExpandedViewsData((prev) => {
+            const { [activityOrFlowId]: _, ...rest } = prev;
+
+            return rest;
+          });
         }, 300);
       }
     },
@@ -159,7 +161,7 @@ const ByParticipant = () => {
                   data-test-id={`${dataTestId}-${index}`}
                 />
               }
-              isLoadingExpandedView={!!expandedViewsLoading[activity.id]}
+              isLoadingExpandedView={expandedViewsLoading[activity.id]}
             >
               <ActionsMenu
                 anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
--- a/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
+++ b/src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
@@ -1,25 +1,36 @@
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { useAsync } from 'shared/hooks';
-import { getAppletRespondentSubjectActivitiesApi } from 'api';
+import {
+  getAppletRespondentSubjectActivitiesApi,
+  getTargetSubjectsByRespondentApi,
+  ParticipantActivityOrFlow,
+  TargetSubjectsByRespondent,
+} from 'api';
 import { users } from 'redux/modules';
 import { ActionsMenu, Spinner } from 'shared/components';
+import { RespondentDetails } from 'modules/Dashboard/types';
 
 import { AssignmentsTab, useAssignmentsTab } from '../AssignmentsTab';
 import { ActivitiesList } from '../ActivitiesList';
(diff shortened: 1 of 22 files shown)


src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
```diff
@@ -1,5 +1,181 @@
...
+              expandedView={
+                <ExpandedView
+                  activityOrFlow={activity}
+                  targetSubjects={expandedViewsData[activity.id]}
+                  getActionsMenu={getActionsMenu}
+                  onClickViewData={handleClickNavigateToData}
+                  data-test-id={`${dataTestId}-${index}`}
+                />
+              }
+              isLoadingExpandedView={!!expandedViewsLoading[activity.id]}
```
sultanofcardio, line 153: This value is already a boolean
```suggestion
              isLoadingExpandedView={expandedViewsLoading[activity.id]}
```

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
```diff
@@ -1,5 +1,181 @@
...
+    (isExpanded: boolean, activityOrFlowId: string) => {
+      if (!respondentSubject) return;
+
+      if (isExpanded) {
+        handleRefetchExpandedView(activityOrFlowId);
+      } else {
+        // If expanded view is closed, remove data to free up memory and minimize refetches
+        // (after delay to account for transition)
+        setTimeout(() => {
+          setExpandedViewsData((prev) => ({ ...prev, [activityOrFlowId]: undefined }));
```
sultanofcardio, line 111: Setting the entry to undefined will still keep it in the object, and later on `Object.entries.forEach` will still execute a re-fetch for it. Remove it instead
```suggestion
          setExpandedViewsData((prev) => {
            const { [activityOrFlowId]: _, ...rest } = prev;

            return rest;
          });
```

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
```diff
@@ -1,5 +1,181 @@
...
+
+    onClickNavigateToData(activityOrFlow, targetSubject.id);
+  };
+
+  const handleClickToggleExpandedView = useCallback(
+    (isExpanded: boolean, activityOrFlowId: string) => {
+      if (!respondentSubject) return;
+
+      if (isExpanded) {
+        handleRefetchExpandedView(activityOrFlowId);
```
sultanofcardio: Adding `void` before the async function call indicates that we are intentionally ignoring the result/error
```suggestion
        void handleRefetchExpandedView(activityOrFlowId);
```

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
```diff
@@ -1,5 +1,181 @@
...
+
+    fetchActivities({ appletId, subjectId: respondentSubject.id });
+  }, [appletId, fetchActivities, respondentSubject]);
+
+  const handleRefetchAll = useCallback(() => {
+    handleRefetchActivities();
+
+    // Refresh target subject data for any expanded views
+    Object.entries(expandedViewsData).forEach(([id, viewData]) => {
+      if (viewData) handleRefetchExpandedView(id);
```
sultanofcardio:
```suggestion
      if (viewData) void handleRefetchExpandedView(id);
```

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
sultanofcardio: I also think we should differentiate between refetching after the assign drawer closes, and before expanding the card. More often than not, the entries in this list will be the same after expanding the card but since we don't cache the list, it blocks the card from expanding.

Considering that, and the fact that this can be an expensive call depending on the number of results, I think we should cache the results of the call to `getTargetSubjectsByRespondentApi` for some time so that it doesn't perform a fetch on every expand. The cache can be ignored when fetching after the assign drawer closes

sultanofcardio (approved):
Looking good. I left a few comments for your consideration

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
farmerpaul: Since we don't have a pre-existing pattern for caching of endpoints in the Admin App (that I have been able to find, at least), rather than attempt to pull off a new caching system helter-skelter, I spoke to @mbanting and we agreed to split this performance optimization into a post-MI Assign launch fast-follow.

farmerpaul: Created [M2-8051](https://mindlogger.atlassian.net/browse/M2-8051).

src/modules/Dashboard/features/Participant/Assignments/ByParticipant/ByParticipant.tsx
```diff
@@ -29,35 +40,86 @@ const ByParticipant = () => {
...
+    (isExpanded: boolean, activityOrFlowId: string) => {
+      if (!respondentSubject) return;
+
+      if (isExpanded) {
+        void handleRefetchExpandedView(activityOrFlowId);
+      } else {
+        // If expanded view is closed, remove data to free up memory and minimize refetches
+        // (after delay to account for transition)
+        setTimeout(() => {
+          setExpandedViewsData((prev) => ({ ...prev, [activityOrFlowId]: undefined }));
```
farmerpaul: Somehow your comment got lost @sultanofcardio, I guess during rebase? Just wanted to respond:

> Setting the entry to undefined will still keep it in the object, and later on Object.entries.forEach will still execute a re-fetch for it. Remove it instead

I accounted for `undefined` entries [above](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1943/files#diff-5fb2ade196f939ddef9b4ecea13b66a6e82119397515bbbb236193b54dd63cd5R75), only refetching if data is present for that entry. I considered your suggested approach and didn't like that the `const { …, ...rest } = …` approach took up 4 extra lines in the file, but it would mean one less conditional during refetch, so I'll do your version! (Also I shouldn't be so motivated by conciseness 😛)

sultanofcardio: It didn't get lost, I later realised that it was accounted for so I deleted the comment. I should've probably replied to it instead 😅

farmerpaul: I'm just as sneaky, rewording my reply to you like 5 times before I was satisfied lol… 🙌

JustinNusca (approved):
Thanks for the effort in making sure the activity list items are accessible. 

In my original comment, I didn't account for one of the buttons having the same functionality as the original `onClick` pop did, so I think there was more wiggle room there than I realized.

Nested interactive elements can be difficult to maintain accessible UX for, and this appears to address everything I mentioned — even the off-hand comment about misclicks 😅.

Not required, but I wonder if a comment in the `ActivityListItem` component file that explicitly explains the `.primary-button` pattern as a means of adding click-ability to the component? The purpose isn't immediately obvious from just reading the CSS, and I can see reasoning for this being forgotten if this ever gets revisited much later.

farmerpaul:
Thanks for the review @JustinNusca! I did add a JSdoc [comment here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1943/files#diff-3187bea639d37a9ccceccba3a217d38c0f5239e19267bd34490f3fdc038476c5R10-R15) to encourage consumers to use that class in the `children` prop, but it occurred to me now that people likely won't see that comment in their editor as they more likely would for other attribute-like props. Good call on adding it to the component definition itself. :+1:

PR #1945, opened 2024-10-15
feat: add Greek option for invitation emails (M2-7665)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-7665](https://mindlogger.atlassian.net/browse/M2-7665)

This PR adds a "Greek" language option for invitation emails.

This PR also refactors some of the language-related code to make a clear distinction between:
* Languages supported by the API; vs ...
* Languages supported by the UI

Previously, both context uses the same enum, which contained English and French. With this PR, we're adding Greek to only the list of languages supported by the API, with the UI continuing to support only English and French.

Regardless if the 2 contexts would eventually support the same list of language or not, the 2 contexts are nonetheless for different purposes, and therefore the 2 enum should not conflate with each other.

### 📸 Screenshots

Add participants screen

Screenshot 2024-10-15 at 2 49 32 PM

Add team member screen (which hits the same endpoint)

Screenshot 2024-10-15 at 2 49 45 PM
  
### 🪤 Peer Testing

1. Create an applet
2. Attempt to invite a participant with a full-account
  * The language dropdown should now contain "Greek" as an option
3. Attempt to add a team member
  * The language dropdown should now contain "Greek" as an option

### ✏️ Notes

The corresponding BE PR is here: https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1624

### DIFF
--- a/src/shared/components/Footer/Language/SelectLanguage/SelectLanguage.test.tsx
+++ b/src/shared/components/Footer/Language/SelectLanguage/SelectLanguage.test.tsx
@@ -3,7 +3,7 @@
 import { screen, fireEvent } from '@testing-library/react';
 
 import { renderWithProviders } from 'shared/utils/renderWithProviders';
-import { Languages } from 'shared/api';
+import { ApiLanguages } from 'shared/api';
 
 import { SelectLanguage } from './SelectLanguage';
 import { languages } from '../Language.const';
@@ -13,7 +13,7 @@ const props = {
   open: true,
   onClose,
   currentLanguage: {
-    value: Languages.EN,
+    value: ApiLanguages.EN,
   },
 };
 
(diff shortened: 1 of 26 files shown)


src/shared/components/Footer/Language/SelectLanguage/SelectLanguage.test.tsx
```diff
@@ -13,7 +13,7 @@ const props = {
   open: true,
   onClose,
   currentLanguage: {
-    value: Languages.EN,
+    value: ApiLanguages.EN,
```
farmerpaul: Since this code is about the footer language selector:
```suggestion
    value: UiLanguages.EN,
```

src/shared/components/Footer/Language/SelectLanguage/SelectLanguage.test.tsx
```diff
@@ -52,7 +52,7 @@ describe('SelectLanguage', () => {
     fireEvent.click(okButton);
 
     expect(onClose).toBeCalledWith({
-      value: Languages.EN,
+      value: ApiLanguages.EN,
```
farmerpaul:
```suggestion
      value: UiLanguages.EN,
```

farmerpaul (approved):
Looks good, just a couple nits.

PR #1949, opened 2024-10-18
fix: Add "web only" chips to web-only items in the Applet editor's Item Type dropdown. (M2-8058)

JustinNusca:
### 📝 Description

🔗 [M2-8058](https://mindlogger.atlassian.net/browse/M2-8058): [FE][Phrase Builder] Add a “Web Only” tag to the Phrase Builder item type in the Admin app

This PR adds "Web Only" chips to the Item Type dropdown in the Applet editor, for item types that are only accessible in the web app. At the moment, this is just the "Phrase Builder" item type. As part of this change, I:
- [Updated the `GroupedSelectSearchController`](https://github.com/ChildMindInstitute/mindlogger-admin/commit/736d996ca632eeb9a1606b510129d8a5c5da60ba).
  - It now shows "Web Only" chips on responseTypes that are only supported on Web.
  - I also updated how dropdown content was rendered to re-use select value JSX content and minimize custom styling.
    - This results in a minor spacing change: Previously, dropdown items had an unintentionally inconsistent height based on the presence of a "web/mobile only" chip. Now, all dropdown items are the same height.
- [Removed `isMobileOnly` configuration values from `ItemConfiguration`, and the `getIsOnlyMobileValue` function from `GroupedSelectSearchController` utils]((https://github.com/ChildMindInstitute/mindlogger-admin/commit/736d996ca632eeb9a1606b510129d8a5c5da60ba)).
   - The above values were inconsistent with each other, which were themselves inconsistent with the definitions in `shared/utils/responseType`. This resulted in issues like the Applet editor reporting that "Phrase Builder" item types were "Web only", but when the activity was viewed in the Dashboard, the "Take Now" option would be disabled, with a tooltip suggesting the Activity must be completed on mobile devices.
- To fix the above issue, I instead updated the "supportedResponseType" arrays and added new "isXOnly" functions to `/shared/utils/responseType`, which should now serve as the source of truth for these evaluations (at least within the context of the admin app).
- [Added/Updated tests for the above scenarios](https://github.com/ChildMindInstitute/mindlogger-admin/commit/01b2ba07a72cf04cdb4f557eb8bf79f3d512ec8b). This includes:
   - Removing `@ts-nocheck`s and fixed eslint warnings in the test files I modified.
   - Removing redundant tests for removed functionality.
   - Adding new tests for `getIsMobileOnly`/`getIsWebOnly` in `shared/utils/responseType`.
   - Adding new tests for "X-Only" chip rendering in `GroupedSelectSearchController`.

### 📸 Screenshots

| Before | After |
|-|-|
| before | after |

### 🪤 Peer Testing

While editing an applet...

1. Add a new item to an activity. Select "Phrase Builder" for the item type.
2. Observe that the Phrase Builder item option includes a "Web Only" chip, both in the select menu and when selected.

If you do not already have an activity that includes a Phrase Builder item, fill in the required details and save your changes.

1. Navigate to the Activities page, and open the "Additional Options"/"…" menu for the activity that contains a Phrase Builder item.
2. Observe that the "Take Now" option is enabled. Previously, it would be disabled, with a tooltip indicating that the activity must be completed on mobile.

[M2-8058]: https://mindlogger.atlassian.net/browse/M2-8058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/GroupedSelectSearchController/GroupedSelectSearchController.styles.ts
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/GroupedSelectSearchController/GroupedSelectSearchController.styles.ts
@@ -1,4 +1,4 @@
-import { styled, MenuItem, ListItem, Select, Box } from '@mui/material';
+import { styled, MenuItem, ListItem, Select } from '@mui/material';
 
 import { theme, variables } from 'shared/styles';
 import { shouldForwardProp } from 'shared/utils';
@@ -84,11 +84,4 @@ export const StyledListSubheader = styled(ListItem)`
   .svg-checkbox-multiple-filled {
     stroke: ${palette.on_surface_variant};
   }
-`;
-
-export const StyledMobileOnly = styled(Box)`
-  margin-left: ${theme.spacing(1)};
-  padding: ${theme.spacing(0.6, 1.2)};
-  background-color: ${variables.palette.on_surface_variant_alfa8};
-  border-radius: ${variables.borderRadius.xs};
 `;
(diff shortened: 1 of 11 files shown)


LashaunnaS (approved):
Adds the new chip ✅
Looks great 👍🏾

PR #1942, opened 2024-10-11
feat: [Duplicate Applet] Enhance Duplication (M2-7729)

sultanofcardio:
- [ ] Tests for the changes have been added

### 📝 Description

🔗 [Jira Ticket M2-7729](https://mindlogger.atlassian.net/browse/M2-7729)

This PR adds a checkbox to the applet duplication popups for including report server configuration. The checkbox is optional and duplication works the same without checking it.

When the checkbox is checked, the duplicated applet will contain these properties from the original applet:

- `reportServerIp`
- `reportPublicKey`
- `reportIncludeUserId`
- `reportIncludeCaseId`
- `reportEmailBody`

> [!NOTE]
> `reportRecipients` is explicitly excluded

### 📸 Screenshots

https://github.com/user-attachments/assets/1d366895-65af-449c-96b2-3b71087a832c

### 🪤 Peer Testing

1. Create an applet with a report server configuration (you can use the [dev server](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/348684325/Credentials#Dev-env))
2. Duplicate the applet both with and without the report server configuration (monitor dev tools network requests)
3. Confirm that the applet can be duplication without checking the box
4. Confirm that requests are sent to the report server `/verify` and `/set-password` endpoints
5. Confirm that the applet duplicated with report server configuration has the properties mentioned above in the applet builder

### ✏️ Notes

N/A

PR #1950, opened 2024-10-18
fix: PDF reports for duplicated applets that include Report Server Configuration (M2-8057)

mbanting:

### 📝 Description

🔗 [Jira Ticket M2-8057](https://mindlogger.atlassian.net/browse/M2-8057)

This fixes an issue with the new Duplicate Applet functionality when including the Report Server Configuration (implemented in [M2-7729](https://mindlogger.atlassian.net/browse/M2-8057)). The PDF report was not generating successfully with these duplicated applets. 

Determined the cause of this was the Admin Panel sending an empty password to the Report Server during replication (via the report server's `set-password` endpoint). This was because the `encryptionDataRef` was getting reset in the finally clause of the `useAsync` call to `duplicateAppletApi`, _before_ the report server configuration setup had a chance to complete.

Changes include:

- Updates `useAsync` of `duplicateAppletApi` by moving cleanup out of the `finally` callback, and instead handles the `encryptionDataRef` cleanup on both success and error callbacks separately, ensuring cleanup takes place after it is no longer needed. 

### 📸 Screenshots

https://www.loom.com/share/c2f01f6aa6cf4b91a9d60197b1753214?sid=02575f6e-3b68-4a9e-919c-2bf375814ced

### 🪤 Peer Testing

1. Setup an Applet with a single activity and item with scoring
2. Configure the Applet's Report Configuration and use one of the existing Report Servers (recommend [UAT](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/348684325/Credentials#UAT-env))
3. Configure the activity's Scores & Reports to Generate Report and Show Score Summary, and configure a Score
4. Save & Publish the Applet
5. Complete the activity (feel free to initiate it via Take Now)
6. Confirm a report was received (this is existing functionality)
7. Go back to Admin and duplicate the applet, including Report Server configuration
8. Go to duplicate applet's Report Configuration and enter an email (since it is not duplicated from original applet)
9. Complete the activity of the duplicated applet (feel free to initiate it via Take Now)
10.  **Expected outcome:** a PDF report is received

### ✏️ Notes

- https://github.com/ChildMindInstitute/mindlogger-admin/pull/1942

[M2-7729]: https://mindlogger.atlassian.net/browse/M2-7729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Dashboard/features/Applet/Popups/DuplicatePopups/DuplicatePopups.tsx
+++ b/src/modules/Dashboard/features/Applet/Popups/DuplicatePopups/DuplicatePopups.tsx
@@ -112,6 +112,8 @@ export const DuplicatePopups = ({ onCloseCallback }: { onCloseCallback?: () => v
     },
   );
 
+  const resetEncryptionData = () => (encryptionDataRef.current = {});
+
   const { execute: executeDuplicate, isLoading: isDuplicateLoading } = useAsync(
     duplicateAppletApi,
     async (res) => {
@@ -122,9 +124,7 @@ export const DuplicatePopups = ({ onCloseCallback }: { onCloseCallback?: () => v
       if (error?.response?.status === ApiResponseCodes.Forbidden) return;
 
       setErrorModalVisible(true);
-    },
-    () => {
-      encryptionDataRef.current = {};
+      resetEncryptionData();
     },
   );
 
@@ -157,6 +157,7 @@ export const DuplicatePopups = ({ onCloseCallback }: { onCloseCallback?: () => v
 
     onCloseCallback?.();
     duplicatePopupsClose();
+    resetEncryptionData();
 
     Mixpanel.track('Applet Created Successfully', {
       'Applet ID': currentAppletId,


farmerpaul (approved):
Hey @mbanting, the code looks good to me. I guess this `useAsync` hook doesn't actually handle async callbacks by awaiting them, else this might not have been an issue – is that right?

I trust that this fix solves the issue, but unfortunately I'm unable to get my local environment to email me a PDF report at all – step 6 in your Peer Testing section – even though I've double- and triple-checked that I have everything configured right. I assume that despite my checks, I still don't have something properly set up to get it to work. The whole reporting and scoring feature of MindLogger has been very difficult for me to wrap my head around.

I'll pre-approve because I trust your judgment!

LashaunnaS (approved):
I was able to successfully reproduce the expected behavior. ✅

mbanting:
> Hey @mbanting, the code looks good to me. I guess this `useAsync` hook doesn't actually handle async callbacks by awaiting them, else this might not have been an issue – is that right?
> 
> I trust that this fix solves the issue, but unfortunately I'm unable to get my local environment to email me a PDF report at all – step 6 in your Peer Testing section – even though I've double- and triple-checked that I have everything configured right. I assume that despite my checks, I still don't have something properly set up to get it to work. The whole reporting and scoring feature of MindLogger has been very difficult for me to wrap my head around.
> 
> I'll pre-approve because I trust your judgment!

Thanks @farmerpaul !

The issue wasn't with useAsync not awaiting correctly—it does handle async callbacks properly (I think). The real issue was that the code needing `encryptionDataRef` was in a `useEffect`, which runs asynchronously and independent of the callbacks, including the finally callback that was resetting the data.

farmerpaul:
> The issue wasn't with useAsync not awaiting correctly—it does handle async callbacks properly (I think). The real issue was that the code needing encryptionDataRef was in a useEffect, which runs asynchronously and independent of the callbacks, including the finally callback that was resetting the data.

What I mean is that `useAsync` doesn't handle asynchronous _success and error callbacks_. It _does_ properly handle the asynchronous nature of the _main callback_ that's passed to the hook, but I think it disregards any asynchronous nature of the success and error callbacks.

So for example, if the `successCallback` happens to be an async one and returns a Promise, that return value is discarded – the hook doesn't `await` it [here](https://github.com/ChildMindInstitute/mindlogger-admin/blob/a55547a273aa2798fa9b264f2481b71cbb860d40/src/shared/hooks/useAsync/useAsync.ts#L65), before finishing the `then()` condition. That means the `finallyCallback` will run even if the `successCallback` is still busy doing things asynchronously.

mbanting:
> > The issue wasn't with useAsync not awaiting correctly—it does handle async callbacks properly (I think). The real issue was that the code needing encryptionDataRef was in a useEffect, which runs asynchronously and independent of the callbacks, including the finally callback that was resetting the data.
> 
> What I mean is that `useAsync` doesn't handle asynchronous _success and error callbacks_. It _does_ properly handle the asynchronous nature of the _main callback_ that's passed to the hook, but I think it disregards any asynchronous nature of the success and error callbacks.
> 
> So for example, if the `successCallback` happens to be an async one and returns a Promise, that return value is discarded – the hook doesn't `await` it [here](https://github.com/ChildMindInstitute/mindlogger-admin/blob/a55547a273aa2798fa9b264f2481b71cbb860d40/src/shared/hooks/useAsync/useAsync.ts#L65), before finishing the `then()` condition. That means the `finallyCallback` will run even if the `successCallback` is still busy doing things asynchronously.

Oh I see what you're referring to. Yes that is also true! Looks like `useAsync` isn't designed to handle async success or error callbacks.

mbanting:
Thanks for the review @LashaunnaS @farmerpaul !

PR #1953, opened 2024-10-21
feat: Mixpanel Event Tracking – item types/SSI on applet publish (M2-7519)

farmerpaul:
- [x] Tests for the changes have been added(https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/263127186/Admin+Panel+Applet+Builder+Library)

### 📝 Description

🔗 [Jira Ticket M2-7519](https://mindlogger.atlassian.net/browse/M2-7519)

This PR adds new event properties to existing Mixpanel events, which are triggered when an applet is saved. The events affected are:

- `Applet Save click`
- `Applet Created Successfully`
- `Applet edit successful`

The new properties added are:

- `Item Types`: an array of unique item types in use by the applet's activities
- `Feature` array containing `SSI`: when the applet contains at least one activity that makes use of the Phrase Builder item type

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Refer to the [ticket description](https://mindlogger.atlassian.net/browse/M2-7519) for a table listing what events to expect, how to trigger them, and what properties should be attached to them.

For each event:

2. Open Mixpanel, and navigate to the Events tab.
3. Refresh if needed.
    **Expected outcome:** The event matching the **Event Name** column should have been added to the top of the list.
    **Expected outcome:** The custom properties for that event mentioned in the **Details** column should be listed.

### DIFF
--- a/src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
+++ b/src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
@@ -13,17 +13,16 @@ import {
   useRemoveAppletData,
 } from 'shared/hooks';
 import {
-  AppletSaveEvent,
   Encryption,
   getDictionaryObject,
   getEncryptionToServer,
   getSanitizedContent,
   getUpdatedAppletUrl,
   Mixpanel,
-  MixpanelAppletSaveEventType,
   MixpanelEventType,
   MixpanelProps,
   SettingParam,
+  trackAppletSave,
 } from 'shared/utils';
 import { Activity, ActivityFlow, applet, SingleApplet } from 'shared/state';
 import { getAppletUniqueNameApi } from 'shared/api';
@@ -43,7 +42,6 @@ import { banners } from 'shared/state/Banners';
 import { ErrorResponseType, LocationState, LocationStateKeys } from 'shared/types';
 import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
 import { AppletPasswordRefType } from 'modules/Dashboard/features/Applet/Popups';
-import { ItemResponseType } from 'shared/consts';
 
 import {
   getActivityItems,
@@ -328,31 +326,6 @@ export const useSaveAndPublishSetup = (): SaveAndPublishSetup => {
   const { hasChanges: hasReportConfigChanges } = reportConfig.useReportConfigChanges() || {};
   const isDisplayNameDirty = dirtyFields?.displayName;
 
-  const trackSave = useCallback(
-    (action: MixpanelAppletSaveEventType) => {
-      const applet = getAppletData();
-
-      const itemTypes: ItemResponseType[] = [];
-      (applet?.activities ?? []).forEach((activity) => {
-        itemTypes.push(...(activity.items ?? []).map((item) => item.responseType));
-      });
-      const uniqueItemTypes = new Set(itemTypes);
-
-      const event: AppletSaveEvent = {
-        action,
-        [MixpanelProps.AppletId]: appletId,
-        [MixpanelProps.ItemTypes]: [...uniqueItemTypes],
-      };
-
-      if (uniqueItemTypes.has(ItemResponseType.PhrasalTemplate)) {
-        event[MixpanelProps.Feature] = 'SSI';
-      }
-
-      Mixpanel.track(event);
-    },
-    [appletId, getAppletData],
-  );
-
   useEffect(() => {
     const isUpdateOrCreate =
       responseTypePrefix === AppletThunkTypePrefix.Create ||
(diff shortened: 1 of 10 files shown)


src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
```diff
@@ -539,9 +561,7 @@ export const useSaveAndPublishSetup = (): SaveAndPublishSetup => {
     }
 
     if (createApplet.fulfilled.match(result)) {
-      Mixpanel.track('Applet Created Successfully', {
-        'Applet ID': appletId,
-      });
+      trackSave('Applet Created Successfully');
```
sultanofcardio: I think we should include the `createdAppletId` in this call

```typescript
trackSave('Applet Created Successfully', {
  [MixpanelProps.AppletId]: createdAppletId,
});
```

The event currently shows up with the value `new-applet`.

The `Password added successfully` event also has this issue, but fixing that may be outside the scope of this PR

sultanofcardio (approved):
Nice work. I left one suggestion for improvement

sultanofcardio:
@farmerpaul there are two other places in the codebase where the `Applet Created Successfully` event is triggered

https://github.com/ChildMindInstitute/mindlogger-admin/blob/f93cc043ae43a3d678d4c117f00c54bdf4e2a2ac/src/modules/Dashboard/features/Applet/Popups/DuplicatePopups/DuplicatePopups.tsx#L162-L164

> [!NOTE]
> This one ☝️ appears to be using the wrong applet ID. It should be using the ID of the duplicated applet

https://github.com/ChildMindInstitute/mindlogger-admin/blob/d366e4ab704f5b96e7b9bfdbd74cc2ad0d126c40/src/modules/Dashboard/features/Applets/AppletsTable/AppletItem/AppletItem.tsx#L115-L117

Should this ticket also cover those events?

src/modules/Builder/features/SaveAndPublish/SaveAndPublish.hooks.ts
farmerpaul: Fixed both instances.

farmerpaul:
Hey @sultanofcardio, thanks for finding those two other instances of `Applet Created Successfully`. I was able to fix the first one. As for the second one:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/d366e4ab704f5b96e7b9bfdbd74cc2ad0d126c40/src/modules/Dashboard/features/Applets/AppletsTable/AppletItem/AppletItem.tsx#L115-L117

I haven't been able to figure out how to get the code in this callback to actually run. I'm also a bit confounded by the logic. The `submitCallback`, which contains this code for tracking another `Applet Created Successfully` event, is the callback specified for when the `AppletPasswordPopup` (contained in this `AppletItem` component, configured specifically for _creating_ a password) is submitted. But the logic to show the `AppletPasswordPopup` – via the `checkAppletEncryption` function – is called in several different places, including most menu actions for the applet (not just applet duplication). Very confusing. :confused:

Do you think there is some unused code in place here, from perhaps an older approach for requesting the password from the user, and it just never got cleaned up?

sultanofcardio:
> I haven't been able to figure out how to get the code in this callback to actually run

I don't think it's being run. I'm not sure what led to wrapping most menu items in it, but none of them currently seem to call it. My hunch is to just update it to match the other calls and determine whether it should be removed later, because it will probably require a bit of investigation

sultanofcardio (approved):

farmerpaul:
> My hunch is to just update it to match the other calls

Problem is, the function that contains that event tracking call doesn't have access to any duplicated applet data, because that's not what this function even does. A few lines before the `Mixpanel.track` statement, that function performs a POST on `/applets/${appletId}/encryption` to update the password for the applet ID currently associated with the `AppletItem` component. So I'm going to take a stab and assume this Mixpanel statement was actually just written wrong – it should be a `Password added successfully`. I've pushed that change.

farmerpaul:
@sultanofcardio I just made a change to this PR to treat the `Feature` event property as an array instead of a scalar string, after discussions with the data team. Updated the ticket as well. Also added unit tests for `mixpanel.utils.ts`. If you could give it one last look, that'd be great! 🙏🏻

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
```diff
@@ -101,11 +102,12 @@ export const useTakeNowModal = ({ dataTestId }: UseTakeNowModalProps) => {
       newActivityOrFlow?: Activity | HydratedActivityFlow | ParticipantActivityOrFlow,
     ) => {
       const props: MixpanelPayload = {
-        [MixpanelProps.Feature]: 'Multi-informant',
         [MixpanelProps.AppletId]: appletId,
         [MixpanelProps.MultiInformantAssessmentId]: multiInformantAssessmentId,
         ...payload,
       };
+      addFeatureToAnalyticsPayload(props, 'Multi-informant');
```
sultanofcardio: This function feels like just a special case of spread appending, but I see that the inline syntax for this will be a bit awkward because the values in the payload are `unknown`. I'm doing a bit of a typed refactor that may make this function unnecessary

sultanofcardio (approved):

src/modules/Dashboard/components/TakeNowModal/TakeNowModal.tsx
farmerpaul: I'm totally down for improvements here… I wasn't necessarily too pleased with what I came up with in the web app, either, so I'm all ears!

sultanofcardio: Will have a PR up soon 🙏

sultanofcardio (approved):

PR #1958, opened 2024-10-24
feat: Add `PreviewPhrasePopup`. (M2-7164)

JustinNusca:
### 📝 Description

🔗 [M2-7164](https://mindlogger.atlassian.net/browse/M2-7164): [R2] [Phrase Builder Item Type] Preview a phrase

This PR adds a new popup — `PreviewPhrasePopup`, which can be used to view an approximation of how a specific phrase within a Phrase Builder item may appear to a user within the web app.

This new popup has been connected to "Preview Phrase" button, which was previously non-functional.

### 🪤 Peer Testing

While editing an activity within an applet, that has multiple items compatible with the phrase builder item type…

1. Create a new item, and select the type "Phrase Builder".
2. Edit the new phrase to include variations of possible fields, including each of the compatible item types.
3. Press "Preview Phrase" and observe that the popup displays the preview correctly.

[M2-7164]: https://mindlogger.atlassian.net/browse/M2-7164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
+++ b/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
@@ -5,10 +5,10 @@ import { theme, variables } from 'shared/styles';
 export const StyledImage = styled('img')({
   borderRadius: variables.borderRadius.md,
   flexShrink: 0,
-  height: 56,
+  height: '5.6rem',
   objectFit: 'cover',
   overflow: 'hidden',
-  width: 56,
+  width: '5.6rem',
 });
 
 export const StyledPreviewContainer = styled('div')({
@@ -16,7 +16,7 @@ export const StyledPreviewContainer = styled('div')({
   borderRadius: variables.borderRadius.lg2,
   display: 'flex',
   gap: theme.spacing(2.4),
-  maxHeight: 400,
+  maxHeight: '40rem',
   overflowY: 'auto',
   padding: theme.spacing(4, 3.2),
   placeItems: 'flex-start',
--- a/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
+++ b/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
@@ -1,0 +1,23 @@
+import { styled } from '@mui/material';
+
+import { theme, variables } from 'shared/styles';
+
+export const StyledImage = styled('img')({
+  borderRadius: variables.borderRadius.md,
+  flexShrink: 0,
+  height: 56,
+  objectFit: 'cover',
+  overflow: 'hidden',
+  width: 56,
+});
+
+export const StyledPreviewContainer = styled('div')({
+  border: `${variables.borderWidth.md} solid ${variables.palette.outline_variant}`,
+  borderRadius: variables.borderRadius.lg2,
+  display: 'flex',
+  gap: theme.spacing(2.4),
+  maxHeight: 400,
+  overflowY: 'auto',
+  padding: theme.spacing(4, 3.2),
+  placeItems: 'flex-start',
+});
--- a/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.tsx
+++ b/src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.tsx
@@ -1,0 +1,72 @@
+import { useCallback } from 'react';
+import { Box } from '@mui/material';
+import { FieldArrayWithId } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+
+import { useCustomFormContext } from 'modules/Builder/hooks';
+import { PhrasalTemplateField } from 'redux/modules';
+import { Modal, ModalProps } from 'shared/components';
+import { StyledModalWrapper } from 'shared/styles';
+
+import { StyledImage, StyledPreviewContainer } from './PreviewPhrasePopup.styles';
+
+export const PreviewPhrasePopup = ({
+  fields = [],
+  name = '',
+  imageUrl = '',
+  ...otherProps
+}: Omit<ModalProps, 'children'> & {
+  fields?: FieldArrayWithId<{ values: PhrasalTemplateField[] }, 'values', 'id'>[];
+  name?: string;
+  imageUrl?: string;
+}) => {
+  const { t } = useTranslation('app');
+  const { getValues } = useCustomFormContext();
+
+  const renderField = useCallback(
+    (_: PhrasalTemplateField, index = 0) => {
+      const fieldValue = getValues(`${name}.${index}` as string);
+
+      if (!fieldValue) {
+        return '';
+      }
+
+      const renderedName = ` ${t('phrasalTemplatePreviewPopup.responsePlaceholder', {
+        itemName: fieldValue?.itemName,
+      })} `;
+
+      switch (fieldValue.type) {
+        case 'sentence':
+          return <Box component="span">{fieldValue.text}</Box>;
+        case 'line_break':
+          return <Box component="hr" sx={{ m: 0, height: 32, border: 'none' }} />;
+        case 'item_response':
+          return ['sentence', 'sentence_option_row', 'sentence_row_option'].includes(
+            fieldValue.displayMode,
+          ) ? (
+            <Box component="span" fontWeight="bold">
+              {renderedName}
+            </Box>
+          ) : (
+            <ul>
+              <Box component="li" fontWeight="bold">
+                {renderedName}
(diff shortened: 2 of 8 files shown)


src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { theme, variables } from 'shared/styles';
+
+export const StyledImage = styled('img')({
+  borderRadius: variables.borderRadius.md,
+  flexShrink: 0,
+  height: 56,
+  objectFit: 'cover',
+  overflow: 'hidden',
+  width: 56,
```
farmerpaul: Just to stick with our existing dimension units convention (for what it's worth):
```suggestion
  height: '5.6rem',
  objectFit: 'cover',
  overflow: 'hidden',
  width: '5.6rem',
```

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+export const StyledPreviewContainer = styled('div')({
+  border: `${variables.borderWidth.md} solid ${variables.palette.outline_variant}`,
+  borderRadius: variables.borderRadius.lg2,
+  display: 'flex',
+  gap: theme.spacing(2.4),
+  maxHeight: 400,
```
farmerpaul:
```suggestion
  maxHeight: '40rem',
```

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplatePhrase/PhrasalTemplatePhrase.tsx
```diff
@@ -139,7 +141,7 @@ export const PhrasalTemplatePhrase = ({
               onClick={handlePreviewPhrase}
               sx={{ gap: 0.8, width: 'max-content' }}
               variant="text"
-              disabled={formState.isDirty || formState.isSubmitting || !formState.isValid}
+              disabled={formState.isSubmitting || !formState.isValid}
```
farmerpaul, line 152: This button's disabled state is not obvious. The only difference is it lacks a hover effect, and the click action bubbles up to its container. When I first tested your work, I kept thinking my app hadn't refreshed from the right git branch, because clicking it was just toggling the card's collapse state. It wasn't until I realized that the form state was invalid (which has no visual indicator until you've tried to click Save & Publish at least once) that the button was simply disabled.

CleanShot 2024-10-25 at 10 00 07

So I guess at a minimum, this button variant should have [proper disabled styling](https://www.figma.com/design/LXl9BAJSwtgWuzkxqOqvX8/ML1-Design-System?node-id=57868-142695&m=dev).

Besides the styling issue, I'm also not sure about the UX of clicking a disabled button resulting in its ancestor's click handler to execute. I think it would feel more natural if nothing happened when I clicked it when disabled, and the card expand toggle handler was limited to its surrounding area instead, but I'm not strongly attached to that.

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+      const renderedName = ` ${t('phrasalTemplatePreviewPopup.responsePlaceholder', {
+        itemName: fieldValue?.itemName,
+      })} `;
+
+      switch (fieldValue.type) {
+        case 'sentence':
+          return <Box component="span">{fieldValue.text}</Box>;
+        case 'line_break':
+          return <Box component="hr" sx={{ m: 0, height: 32, border: 'none' }} />;
```
farmerpaul, line 42: Line breaks don't actually insert a gap… this is what the web app adds when it encounters one:
```suggestion
          return <br />;
```

farmerpaul (approved):
Awesome work again @JustinNusca, so quick! I just pointed out a couple minor things, and one UX issue, but it's easy enough to fix that I'm confident to pre-approve. ✨ 🚀 ✨

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.tsx
JustinNusca: Hmm, are you sure there should be no appearance of a gap? I couldn't find explicit direction about the preview card, but I was following the [spec for the web app, which specifies that line breaks should insert a 32px gap](https://www.figma.com/design/3wcZ6bxK2QVo733ADxVjBw/SSI-2024?node-id=692-76427&t=BLw76cfZQzfaJAK6-4). 

It doesn't seem like there's a conventional/cross-browser way to style a `<br>` tag like this.

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
JustinNusca: [Updated](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1958/commits/ca4cb25aa34101f132123d46b25ef328dd70354e)!

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.styles.ts
JustinNusca: [Updated](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1958/commits/ca4cb25aa34101f132123d46b25ef328dd70354e)!

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/PhrasalTemplate/PhrasalTemplatePhrase/PhrasalTemplatePhrase.tsx
JustinNusca: I pushed up [some changes](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1958/commits/ca4cb25aa34101f132123d46b25ef328dd70354e) that should address this.

First, I fixed the appearance of the disabled state for the button's `text` variant — the theme configuration included some selector specificity conflicts that prevented the correct text color from being applied.

Since this is a global change, I clicked around other locations where we were using the `text` variant with a (conditionally) disabled state (eg, The "Feedback" & "Download Latest Report" buttons in `/dashboard/{appletId}/participants/{subjectId}/activities/{activityId}/{summary | responses}`) — there's not that many and there doesn't seem to be any adverse effect elsewhere 😀.

So that solves one problem. The second issue — the nested interactive elements behaviour — is almost identical in cause to the one I nagged you about in your [PDP pull request](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1943). 😭

My solution here is a bit of a kludge. I was hoping to do something similar to yours in that PR, but we're using slightly different components (`Accordion` vs `Collapse`) and there's a slightly different desired behaviour (we want a `disabled` button to continue to prevent clicks on the parent element). 

In the end I just added an additional `onClick` with a comment explaining the purpose on the wrapper node around both this and the "Delete Phrase" button, that simply calls `stopPropagation`. This works to stand-in as a handler for the button when it's in a disabled state, and avoids misclicks if you happen to press in the ambiguous boundary between the two buttons.

src/modules/Builder/components/PreviewPhrasePopup/PreviewPhrasePopup.tsx
farmerpaul: Ah, my bad. I didn't realize this was a [bug in the web app code](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/4b938c806d4d5dc356998dd944b845c5bb8300f5/src/entities/activity/ui/items/ActionPlan/Phrase.tsx#L46). I was using the web app as the source of truth rather than Figma. I guess I'll set a reminder to create a ticket for this next week, unless you have time to create a quick PR for the web app… 😏

JustinNusca: I was somewhat reluctant to open a new PR to address this, given all of @qiushihe's work that's going into refactoring the code in that exact area, but I left a [comment on the PR](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/541/files#r1817086469) that should hopefully help explain and allow her to take care of it 🤞.

PR #1959, opened 2024-10-24
chore: Mixpanel Refactor

sultanofcardio:
### 📝 Description

This PR modifies the way events are tracked with Mixpanel in the following ways:

- The methods in the `Mixpanel` object are no longer async. This is because they don't return anything and are never awaited
- There is now a central place to define events and their properties so that we can conveniently see which properties are included in each event, and where they are used.

Now in order to track a new event, you first define it in the event type in the `MixpanelEventType` enum

```typescript
export enum MixpanelEventType {
  UpgradeToFullAccountClicked = 'Upgrade to Full Account clicked',
}
```

Then define any extra properties it will need in the `MixpanelProps` enum

```typescript
export enum MixpanelProps {
  Via = 'Via',
}
```

Then finally define the event type and add it to the `MixpanelEvent` type union

```typescript
export type UpgradeToFullAccountClickedEvent = WithAppletId<{
  action: MixpanelEventType.UpgradeToFullAccountClicked;
  [MixpanelProps.Via]: 'Applet - Participants';
}>;

export type MixpanelEvent =
  | MixpanelInvitationSentEvent
  | LoginSuccessfulEvent
  | LoginBtnClickEvent
  // ...
  UpgradeToFullAccountClickedEvent;
```

> [!TIP]
> Wrapping the type in `WithAppletId` adds an optional `Applet ID` property to the event

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Read the code I guess 🤷

### ✏️ Notes

N/A

### DIFF
--- a/src/shared/hooks/useFeatureFlags.ts
+++ b/src/shared/hooks/useFeatureFlags.ts
@@ -24,8 +24,6 @@ export const useFeatureFlags = () => {
     const features: FeatureFlags = {};
     keys.forEach((key) => (features[key] = flags[FeatureFlagsKeys[key]]));
 
-    features.enableCahmiSubscaleScoring = true;
-
     return features;
   };
 
--- a/src/shared/hooks/useFeatureFlags.ts
+++ b/src/shared/hooks/useFeatureFlags.ts
@@ -24,6 +24,8 @@ export const useFeatureFlags = () => {
     const features: FeatureFlags = {};
     keys.forEach((key) => (features[key] = flags[FeatureFlagsKeys[key]]));
 
+    features.enableCahmiSubscaleScoring = true;
+
     return features;
   };
 
--- a/src/shared/utils/mixpanel/mixpanel.ts
+++ b/src/shared/utils/mixpanel/mixpanel.ts
@@ -1,4 +1,4 @@
-import { MixpanelPayload, MixpanelAction } from 'shared/utils/mixpanel/mixpanel.types';
+import { MixpanelEvent } from 'shared/utils/mixpanel/mixpanel.types';
 
 import { isProduction, isStaging, isUat, isDev } from '../env';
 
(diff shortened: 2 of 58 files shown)


src/shared/hooks/useFeatureFlags.ts
```diff
@@ -24,6 +24,8 @@ export const useFeatureFlags = () => {
     const features: FeatureFlags = {};
     keys.forEach((key) => (features[key] = flags[FeatureFlagsKeys[key]]));
 
+    features.enableCahmiSubscaleScoring = true;
+
```
farmerpaul: 😁
```suggestion
```

src/shared/utils/mixpanel/mixpanel.ts
```diff
@@ -10,38 +10,44 @@ const shouldEnableMixpanel =
...
     if (shouldEnableMixpanel) {
-      const { default: mixpanel } = await import('mixpanel-browser');
-      mixpanel.identify(userId);
+      import('mixpanel-browser').then(({ default: mixpanel }) => {
+        mixpanel.identify(userId);
+      });
       mixpanel.people.set({
         'User ID': userId,
         'App Build Number': process.env.REACT_APP_DEVELOP_BUILD_VERSION,
       });
```
farmerpaul: This should be in the `then`, too, right?
```suggestion
      import('mixpanel-browser').then(({ default: mixpanel }) => {
        mixpanel.identify(userId);
        mixpanel.people.set({
          'User ID': userId,
          'App Build Number': process.env.REACT_APP_DEVELOP_BUILD_VERSION,
        });
      });
```

src/shared/utils/mixpanel/mixpanel.types.ts
farmerpaul, line 1: Love the strong typing!

farmerpaul (approved):
This looks great, thanks for taking this on, @sultanofcardio! Aside from the minor things I noted, I think this should be merged sooner than later and I'll adjust my current unmerged admin PR to use the new pattern, rather than the other way around.

I think a corresponding PR for the web app would be in order – do you want me to take this on, or how's your workload? 😏 Seems you and I both just have a bunch of unit tests to write till next sprint…

src/shared/hooks/useFeatureFlags.ts
sultanofcardio: 😅 Thanks

src/shared/utils/mixpanel/mixpanel.ts
sultanofcardio: Indeed it should. Good eye 🙏

sultanofcardio:
> I think a corresponding PR for the web app would be in order – do you want me to take this on, or how's your workload? 😏

Agreed. I can do that one as well

PR #1961, opened 2024-10-26
feat: Assign Event Tracking (M2-7607)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-7607](https://mindlogger.atlassian.net/browse/M2-7607)

This PR adds the following events to Mixpanel tracking:
- `Start Assign Activity/Flow`
- `Confirm Assign Activity/Flow`
- `Start Unassign Activity/Flow`
- `Confirm Unassign Activity/Flow`

It also updates the `Applet Save click` event to add some extra properties related to activity and flow assignments

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Perform the following actions and confirm in the [Mindlogger Analytics Dev](https://mixpanel.com/project/3333580/view/3841316/app/events) project in Mixpanel that the proper events show up with the properties specified in the ticket:

- Perform an activity/flow assignment and un-assignment from the Applet Activities tab using the button in the toolbar
- Perform an activity/flow assignment and un-assignment from the Applet Activities tab using the more menu on the activity card
- Perform an activity/flow assignment and un-assignment from the Participant Assignment tab
- Perform an activity/flow assignment and un-assignment from the Data Viz page
- Make a minor change to the applet (like changing the name) and confirm that the `Applet Save click` event has the updated properties as specified in the ticket

### ✏️ Notes

- I’ve still implemented tracking assignment events on the Participant - Activities page for completeness, even though the two features are mutually exclusive
- This PR is a replacement for #1960 in order to get an Amplify environment running


src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
```diff
@@ -325,6 +326,37 @@ export const ActivityAssignDrawer = ({
...
+          let flowCount = 0;
+
+          for (const assignment of reviewedAssignments) {
+            if (assignment.activityId) {
+              activityCount++;
+            }
+
+            if (assignment.activityFlowId) {
+              flowCount++;
+            }
```
farmerpaul: The idea with tracking activity & flow counts is to get a sense of how many _unique_ activities/flows a user is assigning at one time – essentially, how many checkboxes were checked in the `ActivitiesList`. But this code will return a value of activity/flow count multiplied by the number of rows in the Assignments table. I saw how the wording in the Jira task wasn't 100% clear about that, so I went ahead and tweaked it to include the word "unique" for these properties.

Maybe all you need to do is just pass `activityIds.length` and `flowIds.length` below?

src/modules/Dashboard/features/RespondentData/RespondentDataHeader/RespondentDataHeader.tsx
```diff
@@ -86,6 +92,13 @@ export const RespondentDataHeader = ({
 
   const handleAssignActivity = () => {
     setShowActivityAssign(true);
+    Mixpanel.track({
+      action: MixpanelEventType.StartAssignActivityOrFlow,
+      [MixpanelProps.AppletId]: appletId,
+      [MixpanelProps.ActivityId]: activityId,
+      [MixpanelProps.EntityType]: 'activity',
```
farmerpaul: This header component could be showing in the context of an activity or a flow. This would be one way to handle it:
```suggestion
      [activityFlowId ? MixpanelProps.ActivityFlowId : MixpanelProps.ActivityId]: activityFlowId || activityId,
      [MixpanelProps.EntityType]: activityFlowId ? 'flow' : 'activity',
```

farmerpaul (approved):
Great work. Code looks good, just noticed a couple issues with simple fixes, so preapproving!

src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
sultanofcardio: That's a good call. Thanks for pointing this out, I overlooked that

farmerpaul (approved):

PR #1964, opened 2024-10-29
Feature: Add respondent and subject labels to DataViz responses page (M2-7855)

LashaunnaS:

### 📝 Description

🔗 [Jira Ticket M2-7855](https://mindlogger.atlassian.net/browse/M2-7855)

Changes include:

- Update to the `useRespondenLabel` hook to add appropriate labels for `subject`, `respondent`, or the users credentials without the label. 
- This also adds additional formatted translations for "subject" and "respondent" both with colons. 
- I've also added a column for displaying the Respondents credential on the DataViz response page. (tbd)

### 📸 Screenshots

|  Screenshot 2024-11-04 at 13 35 28 | Screenshot 2024-11-04 at 13 36 17 |

### 🪤 Peer Testing

1. navigate to a users DataViz screen and select the `Responses` tab.
2. confirm that the left Menu displays the Subject with the name and id.
3. confirm that in the activity summary header, there's a "Respondent" column with the respondent for the activity listed below it.
4. do the same for a flow summary.

### DIFF
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -261,6 +261,8 @@ export const RespondentDataReview = () => {
     ? activityAnswers?.[0]?.sourceSubject
     : flowAnswers?.[0]?.answers[0].sourceSubject;
 
+  console.log('sourceSubject', activityAnswers);
+
   return (
     <StyledContainer>
       <ReviewMenu
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -261,8 +261,6 @@ export const RespondentDataReview = () => {
     ? activityAnswers?.[0]?.sourceSubject
     : flowAnswers?.[0]?.answers[0].sourceSubject;
 
-  console.log('sourceSubject', activityAnswers);
-
   return (
     <StyledContainer>
       <ReviewMenu
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -251,6 +251,9 @@ export const RespondentDataReview = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [lastActivityCompleted]);
 
+  const sourceSubject =
+    activityAnswers?.[0]?.sourceSubject || flowAnswers?.[0]?.answers[0].sourceSubject;
+
   return (
     <StyledContainer>
       <ReviewMenu
@@ -296,8 +299,12 @@ export const RespondentDataReview = () => {
           />
           {!isLoading && (
             <>
-              {selectedAnswer && responsesSummary && !error && (
-                <ResponsesSummary {...responsesSummary} data-testid={dataTestid} />
+              {selectedAnswer && responsesSummary && sourceSubject && !error && (
+                <ResponsesSummary
+                  {...responsesSummary}
+                  sourceSubject={sourceSubject}
+                  data-testid={dataTestid}
+                />
               )}
               <EmptyResponses
                 hasAnswers={hasAnswers}
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -251,8 +251,9 @@ export const RespondentDataReview = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [lastActivityCompleted]);
 
-  const sourceSubject =
-    activityAnswers?.[0]?.sourceSubject || flowAnswers?.[0]?.answers[0].sourceSubject;
+  const sourceSubject = hasActivityResponses
+    ? activityAnswers?.[0]?.sourceSubject
+    : flowAnswers?.[0]?.answers[0].sourceSubject;
 
   return (
     <StyledContainer>
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -23,6 +23,7 @@ import { useAsync } from 'shared/hooks';
 import { StyledContainer, theme } from 'shared/styles';
 import { users } from 'redux/modules';
 import { page } from 'resources';
+import { RespondentDetails } from 'modules/Dashboard/types';
 
 import { Feedback } from './Feedback';
 import { ActivityResponses } from './ActivityResponses';
@@ -305,10 +306,10 @@ export const RespondentDataReview = () => {
           />
           {!isLoading && (
             <>
-              {selectedAnswer && responsesSummary && !!sourceSubject && !error && (
+              {selectedAnswer && responsesSummary && !error && (
                 <ResponsesSummary
                   {...responsesSummary}
-                  sourceSubject={sourceSubject}
+                  sourceSubject={sourceSubject as RespondentDetails}
                   data-testid={dataTestid}
                 />
               )}
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
@@ -251,6 +251,11 @@ export const RespondentDataReview = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [lastActivityCompleted]);
 
+  /**
+   * Determines the source subject based on the presence of activity responses.
+   * If activity responses are available, it retrieves the source subject from the first activity answer.
+   * Otherwise, it retrieves the source subject from the first flow answer.
+   */
   const sourceSubject = hasActivityResponses
     ? activityAnswers?.[0]?.sourceSubject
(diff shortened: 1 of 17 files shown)


src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -251,6 +251,10 @@ export const RespondentDataReview = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [lastActivityCompleted]);
 
+  const sourceSubject = hasActivityResponses
+    ? activityAnswers?.[0]?.sourceSubject
+    : flowAnswers?.[0]?.answers[0].sourceSubject;
```
sultanofcardio, line 261: I think we could add a comment here explaining that we just pick the first source subject because they're currently all the same

src/modules/Dashboard/features/RespondentData/RespondentDataReview/ResponsesSummary/ResponsesSummary.hooks.ts
```diff
@@ -18,6 +20,8 @@ export const useResponsesSummary = ({
     DateFormats.MonthDayYearTimeSeconds,
   );
 
+  const respondent = useRespondentLabel({ hideLabel: true, sourceSubject });
```
sultanofcardio: I think the `useRespondentLabel` function is doing too many things and we could probably get by with just calling `getRespondentName` directly here
```suggestion
  const respondent = sourceSubject?.secretUserId
    ? getRespondentName(sourceSubject.secretUserId, sourceSubject.nickname)
    : '';
```

I talk more about this in a subsequent comment

src/shared/hooks/useRespondentLabel.ts
```diff
@@ -1,29 +1,41 @@
...
   hideNickname?: boolean;
+  hideLabel?: boolean;
+  sourceSubject?: RespondentDetails | null;
 };
 
 export const useRespondentLabel = (
   parameters: Parameters = {
     isSubject: false,
     hideNickname: false,
+    hideLabel: false,
```
sultanofcardio: The existing behaviour of this hook is kinda confusing to me, and it seems to stem from it trying to do too many things.

- If there is no `secretUserId`, it returns an empty string
- If `hideNickname` is true, it just returns the subject's `secretUserId`
- It always says Respondent, even if `isSubject` is set to true

This is further compounded by the fact that the `useRespondent` hook now always returns a null result, because that use case is no longer relevant so the api call that populates it never gets called. Instead of further augmenting it to fit more use cases, I think this hook and its usage could be greatly simplified by requiring the caller to pass the subject in as an argument. It doesn't even really need to be a hook tbh:

```typescript
import { RespondentDetails } from 'modules/Dashboard/types';
import { getRespondentName } from 'shared/utils';
import i18next from 'i18n';

type Parameters = {
  subject: RespondentDetails;
  type: 'subject' | 'respondent';
};

/**
 * Label a subject using their secretUserId and nickname (if available). The label used depends on the type of subject
 * @param parameters
 */
export const getSubjectLabel = ({ subject, type }: Parameters) => {
  const { t } = i18next;
  const { secretUserId, nickname } = subject;

  const label = `${t(type ? 'subjectIdentifierWithColon' : 'respondentIdentifierWithColon')}${' '}`;

  return `${label}${getRespondentName(secretUserId, nickname)}`;
};

```

Then its various usages can be replaced like this

[StickyHeader.tsx:17](https://github.com/ChildMindInstitute/mindlogger-admin/blob/5910bfd476aebc589825341c579baa7aaf76b5c2/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/ReportMenu/StickyHeader/StickyHeader.tsx#L17)

```typescript
const subject = users.useSubject()?.result;
const respondentLabel = subject ? getSubjectLabel({ subject, type: 'subject' }) : '';
```

[ReviewMenu.tsx:33](https://github.com/ChildMindInstitute/mindlogger-admin/blob/8c55fc0bf22965d4d0362e5f2605a7e56b75be74/src/modules/Dashboard/features/RespondentData/RespondentDataReview/ReviewMenu/ReviewMenu.tsx#L33)

```typescript
const subject = users.useSubject()?.result;
const respondentLabel = subject?.secretUserId
  ? getRespondentName(subject.secretUserId, subject?.nickname)
  : '';
```

[ResponsesSummary.hooks.ts:23](https://github.com/ChildMindInstitute/mindlogger-admin/blob/22a8c1c5dae12444440feef7708fa07494f647ea/src/modules/Dashboard/features/RespondentData/RespondentDataReview/ResponsesSummary/ResponsesSummary.hooks.ts#L23)

```typescript
const respondent = sourceSubject?.secretUserId
  ? getRespondentName(sourceSubject.secretUserId, sourceSubject.nickname)
  : '';
```

sultanofcardio (approved):
Nice work @LashaunnaS. This PR nails the functionality, so I'm happy to approve it. I've left you some suggestions for how I think things could be improved

src/shared/hooks/useRespondentLabel.ts
LashaunnaS: @sultanofcardio thanks for the review! Good callouts, I updated the `useRespondentLabel` hook to no longer handle `hideLabel` and `sourceSubject`. Where I was calling the hook to display the respondent I pulle din your suggestion to use `getRespondentName`.

src/resources/app-fr.json
```diff
@@ -1021,6 +1020,8 @@
   "noResponseIdentifier": "Aucun identifiant de réponse pour le moment.",
   "viewingResponsesSubmitted": "Consultation des réponses soumises le :",
   "responseIdentifierWithColon": "Identifiant de réponse :",
+  "respondentIdentifierWithColon": "Intimé :",
```
farmerpaul: I believe this might have gotten mistranslated by Google Translate – we have been using the word `Répondant` for respondent in the French language file.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -296,8 +305,12 @@ export const RespondentDataReview = () => {
           />
           {!isLoading && (
             <>
-              {selectedAnswer && responsesSummary && !error && (
-                <ResponsesSummary {...responsesSummary} data-testid={dataTestid} />
+              {selectedAnswer && responsesSummary && !!sourceSubject && !error && (
```
farmerpaul: Looks like a new test is failing in `RespondentDataReview.test.tsx`, maybe you're already on top of it. I'm guessing it's because the data for `sourceSubject` needs to get mocked into the response?

src/modules/Dashboard/features/RespondentData/RespondentDataReview/ResponsesSummary/ResponsesSummary.hooks.ts
```diff
@@ -26,11 +30,16 @@ export const useResponsesSummary = ({
     },
     {
       id: 'review-desc-2',
+      title: t('respondentIdentifierWithColon'),
+      content: respondent,
+    },
+    {
```
farmerpaul, line 37: Adding another column here is causing the last column to wrap to the next row, but in Figma all columns should be on the same row:

| Figma | This PR | 
| - | - |
| CleanShot 2024-11-06 at 10 37 01@2x | CleanShot 2024-11-06 at 10 36 54@2x |

Might just need to update the `columns` prop in the `ResponsesSummary` component!

Also I noticed that the labels and the values have new styling in Figma – the labels have `title-bold/small` (whatever that equates to in our Typography components) while the values have `body/large`.

farmerpaul (approved):
Great work @LashaunnaS! I just identified a couple nits, and a failing unit test that you might already be working on.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
LashaunnaS: @farmerpaul I removed that additional check because I assume the `sourceSubject` will always be returned from the answer object. I don't believe the whole summary should not return if that fails . Is that a correct assumption?

farmerpaul: Ah correct. Yes if for whatever reason the `sourceSubject` is missing we should still render the component. Does this fix that broken test?

src/modules/Dashboard/features/RespondentData/RespondentDataReview/ResponsesSummary/ResponsesSummary.tsx
```diff
@@ -14,13 +14,25 @@ export const ResponsesSummary = ({
...
               {title}
             </StyledTitleMedium>
-            <StyledTitleBoldMedium sx={{ color: variables.palette.on_surface_variant }}>
+            <StyledTitleBoldMedium
+              sx={{
+                color: variables.palette.on_surface_variant,
+                fontWeight: variables.font.weight.regular,
+                fontSize: variables.font.size.md,
+              }}
+            >
```
farmerpaul: Nit: just wanted to confirm whether you checked that these styles aren't available as reusable styled components in [Typography.ts](https://github.com/ChildMindInstitute/mindlogger-admin/blob/develop/src/shared/styles/styledComponents/Typography.ts#L14)? Often predefined font styles from Figma have been defined there, or if they haven't, that's a good place to add it. Either way, not blocking. 🙂

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -261,6 +261,8 @@ export const RespondentDataReview = () => {
     ? activityAnswers?.[0]?.sourceSubject
     : flowAnswers?.[0]?.answers[0].sourceSubject;
 
+  console.log('sourceSubject', activityAnswers);
+
```
farmerpaul:
```suggestion
```

LashaunnaS: 😮‍💨

PR #1965, opened 2024-10-29
fix: Cascade system item changes to report scores (M2-8146)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-8146](https://mindlogger.atlassian.net/browse/M2-8146)

This PR creates logic that cascades changes to activity system items between the subscales configuration and the report configuration.

Whenever a subscale with a lookup table is added to an activity, two non-editable system items are added to the activity (`gender_screen` and `age_screen`). These items then become available for printing in reports. Whenever the subscale lookup table is removed/replaced, these items also get removed and re-added to the activity. Previously if these items were selected for printing, and then subsequently removed/replaced, report items would retain a reference to the older versions of these system items and subsequently throw an error when it fails to find them.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Create an activity in an empty applet with:
- A single selection item with scores
- A subscale with a lookup table (make sure the age field type is set to Text)
- A report score with score type set to Raw Score, and check all the items to be printed
- Save the applet
- Set the subscale age field type to Dropdown
- Save the applet
- Confirm there are no console errors

### ✏️ Notes

N/A

### DIFF
--- a/src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.utils.ts
+++ b/src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoresAndReports.utils.ts
@@ -59,3 +59,9 @@ export const getTableScoreItems = (items?: ItemFormValues[]) =>
       },
     ];
   }, []);
+
+export const reportIsScore = (report: ScoreOrSection): report is ScoreReport =>
+  report.type === ScoreReportType.Score;
+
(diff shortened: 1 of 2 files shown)


farmerpaul (approved):
AC appears to be satisfied, nice work :+1:

mbanting (approved):
LGTM. Tested and works as expected

PR #1968, opened 2024-10-29
fix: Exclude pending accounts from Take Now & Assign (M2-8035)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8035](https://mindlogger.atlassian.net/browse/M2-8035)

Pending accounts shouldn't be selectable for either Take Now or Assign Activity. Updated the `useParticipantDropdown` hook to filter out pending accounts (by default, but it can still be overridden via a prop if ever desired).

Also disabled clicking pending participants from the Applet > Participants table (after confirming with product), to further prevent Take Now modals from being pre-populated with pending participants.

> [!NOTE]
> ✨ In making the change to Participants table row clickability, I felt it important to distinguish which rows were clickable vs. not by finally adding the table row hover effect, which was in fact always part of the original designs. To do so, I did some minor refactoring of table row hover styling, which was already present in the app on the Applets table. See [this commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1968/commits/a01f0001524b7f4878a2e12e7d4f1c16a30c7e8d) for details.

#### Participants table row clickability

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/20f1bf41-0533-4388-98af-4fd795756691"></video>|<video src="https://github.com/user-attachments/assets/781eb0e7-b9d2-43d1-b249-defefd920ced"></video>|

#### Take Now & Assign Activity dropdown lists

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/616b09c5-119b-4beb-960b-2fa55cde6e54"></video>|<video src="https://github.com/user-attachments/assets/f3948942-0ad4-4364-8049-85b4c39643fd"></video>|

#### Improvements to Applets table & Applet Overview table row styling

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/c9aa9148-ec18-489b-a45c-13891466e70b"></video>|<video src="https://github.com/user-attachments/assets/8797617f-db4c-4421-9652-3565603c0c40"></video>|

### 🪤 Peer Testing

1. Create a new full account and leave it in pending state.
2. Perform Take Now (from anywhere in the app) and try to select that account as the subject.
    **Expected outcome:** The pending account is not selectable from the dropdown list.
2. Perform Assign Activity (again, from anywhere in the app) and try to select that account as the subject of an assignment.
    **Expected outcome:** The pending account is not selectable from the dropdown list.
3. Navigate to **Applet > Participants** and try to access the pending account's PDP by clicking its row.
    **Expected outcome:** The row does not respond to being clicked.

### DIFF
--- a/src/modules/Dashboard/components/ParticipantDropdown/ParticipantDropdown.hooks.ts
+++ b/src/modules/Dashboard/components/ParticipantDropdown/ParticipantDropdown.hooks.ts
@@ -4,7 +4,7 @@ import { DEFAULT_ROWS_PER_PAGE, Roles } from 'shared/consts';
 import { ParticipantsData } from 'modules/Dashboard/features/Participants';
 import { getWorkspaceManagersApi, getWorkspaceRespondentsApi } from 'api';
 import { useAsync } from 'shared/hooks';
-import { Manager, Respondent } from 'modules/Dashboard/types';
+import { Manager, Respondent, RespondentStatus } from 'modules/Dashboard/types';
 import { auth, workspaces } from 'redux/modules';
 
 import {
@@ -23,6 +23,7 @@ const ALLOWED_TEAM_MEMBER_ROLES: readonly Roles[] = [
 
 export const useParticipantDropdown = ({
   appletId,
+  includePendingAccounts = false,
   skip = false,
   successCallback,
   errorCallback,
@@ -41,7 +42,11 @@ export const useParticipantDropdown = ({
     (response) => {
       if (response?.data) {
         const options = (response.data as ParticipantsData).result
-          .filter((r) => !r.isAnonymousRespondent)
+          .filter(
+            (r) =>
+              !r.isAnonymousRespondent &&
+              (includePendingAccounts || r.status !== RespondentStatus.Pending),
+          )
           .map(participantToOption);
 
         setAllParticipants(options);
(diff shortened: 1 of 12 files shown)


sultanofcardio (commented):
Everything looks good to me, so I'm happy to approve. There are a few edge cases we haven't considered though, but I'm not sure if they're all worth handling in this PR so you can let me know how you want to proceed.

It is currently possible (before this change) to initiate take now with a pending account as the target subject. When the new PDP design launches, these take now submissions will show up on the by participant page of the full account that performed them. If the target subject is still a pending account, it will still be possible to initiate a take now with them pre-filled as the target subject

image image

Similarly, if we click the view data button, there is a take now button on the data viz page

image

Finally, the back button on that data viz page will take you to the PDP of the pending account.

https://github.com/user-attachments/assets/31f81497-638a-402e-bb6e-6cd464f76ce9

Do you think these should be addressed here or subsequently?

farmerpaul:
@sultanofcardio good point. This would have to entail that actual users have previously performed Take Now or assigned activities with a pending account as the target subject to be able to access them from the By Participant tab, right? So it feels like somewhat more of an edge-case and doesn't need to be covered by this PR.

I can already tell that it's going to either entail an awkward second request, or we need to make changes to the BE endpoint that returns the results for the expanded view, in order to discover whether an account listed there has a "pending" account status, because currently that info is not yet available in that context. So it's going to be a bigger task. I'll put it out to Cari to see if she thinks it's even worth addressing.

sultanofcardio (approved):
Agreed

PR #1970, opened 2024-10-31
fix: Remove MindLogger update version banner (M2-8082)

AlejandroCoronadoN:
### 📝 Description
As an Admin I want the MindLogger version Info banner to be removed to avoid confusion for new Admin users.

🔗 [Jira Ticket M2-8082](https://mindlogger.atlassian.net/jira/software/c/projects/M2/boards/4?assignee=712020%3A6322a990-2eaf-42f4-9fed-30af94baf20e&selectedIssue=M2-8082)

Changes include:

- [Removed addBanner action when session status changes avoiding rendering VersionWarningBanner]
- [Updated tests for VersionWarningBanner and only testing for removing all banners action when sesion status changes]

### 📸 Screenshots

| image | image |

### 🪤 Peer Testing
1. Start your admin app
2. Login with your user
3. "Update version Banner" () shouldn't be displayed at the top of the screen

### ✏️ Notes
* After tlaking with Cari, Marty and Paul, we decided that the best approach was to avoid rendering the Banner component instead of entirely delete it because we might rehuse it in the future.

### DIFF
--- a/src/shared/hooks/useSessionBanners.test.ts
+++ b/src/shared/hooks/useSessionBanners.test.ts
@@ -18,7 +18,7 @@ const emptyState: PreloadedState<RootState> = {
 const populatedState: PreloadedState<RootState> = {
   banners: {
     data: {
-      banners: [{ key: 'VersionWarningBanner' }],
+      banners: [{ key: 'SaveSuccessBanner' }],
     },
   },
 };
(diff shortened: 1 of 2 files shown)


src/shared/hooks/useSessionBanners.test.ts
AlejandroCoronadoN, line 1: I just selected any Banner component for this test since we are testing that all banners are removed. I choosed SaveSuccessBanner arbitrarily.

src/shared/hooks/useSessionBanners.ts
AlejandroCoronadoN, line 1: I guess we still want that other action that removes all banners. I'm not sure when it get's executed but I think is better to keep it and just remove the VersionWarningBanner related action.

farmerpaul (approved):
Tests pass, and code and assumptions are all good @AlejandroCoronadoN, nice work! 🚀

PR #1974, opened 2024-11-07
fix: By Participant list for limited respondents (M2-8092)

farmerpaul:
- [ ] ~~Tests for the changes have been added~~ Will to be done as part of M2&zwnj;-&zwnj;8024 (specific tests for this PR can't be added until general PDP tests have been added)

### 📝 Description

🔗 [Jira Ticket M2-8092](https://mindlogger.atlassian.net/browse/M2-8092)

This PR adds support for viewing the PDP **By Participants** tab for a limited account, since they can be possible valid respondents if they had submissions done on their behalf via Take Now.

Also updated the copy for the limited account empty state on this tab to reflect this nuance.

> [!NOTE]
> Testing this PR requires being connected to the backend containing the changes in https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1654.

#### With submissions:
https://github.com/user-attachments/assets/c68958a9-afbc-4bce-a794-1cbe3ecdb904

#### Without submissions:
CleanShot 2024-11-07 at 14 46 50@2x

#### Preconditions:

An applet with at least one activity or flow, and a new limited account.

#### Steps:

1. Navigate to the PDP for the limited account, then the **By Participant** tab.
    **Expected outcome:** The empty state displays the [updated copy](https://www.figma.com/design/6BNx63peNVzrHAF1FoZ78x/Multi-Informant-2024?node-id=13935-119943&t=ubTxBcV95pvhJbap-4).
1. Navigate to the **Applet > Activities** tab.
2. For any activity or flow, click **Take Now** from the ⋯ menu, providing the limited account for the **Who will be providing the responses?** dropdown, and any participants for the **Who will be inputting the responses?** and **Who are the responses about?** dropdowns.
3. Complete the assessment in the Web App so that there is at least one submission by the limited account as the respondent.
4. Navigate to the PDP for the limited account, then the **By Participant** tab.
    **Expected outcome:** The activity or flow just completed should be listed in `Inactive` state.
6. Expand the card for the activity or flow.
    **Expected outcome:** One subject entry should be listed for the assessment just completed via Take Now.
7. Open the ⋯ menu for that activity or flow, and hover over the **Assign** menu item.
    **Expected outcome:** A tooltip explaining why it's disabled should be shown.
6. Click the **Take Now** menu item.
    **Expected outcome:**
    The Take Now dialogue should be pre-populated with the limited account in the **Who will be providing the responses?** dropdown, and the participant shown in the row that you clicked the ⋯ menu item from in the **Who are the responses about?** dropdown. The checkbox should be disabled and unchecked, and the **Who will be inputting the responses?** dropdown should be empty.

### DIFF
--- a/src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
+++ b/src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
@@ -199,6 +199,7 @@ export const useAssignmentsTab = ({
           EditablePerformanceTasks.includes(performanceTaskType ?? ''));
       const isAssignable =
         status === ActivityAssignmentStatus.Active || status === ActivityAssignmentStatus.Inactive;
+      const isLimitedRespondent = !respondentSubject?.userId;
       const isTargetTeamMember = targetSubjectArg?.tag === 'Team';
       const isAssigned = !!assignments?.some(
         (a) =>
@@ -207,7 +208,7 @@ export const useAssignmentsTab = ({
       );
       const isAssignDisplayed =
         canAssign && (!autoAssign || status === ActivityAssignmentStatus.Hidden);
-      const isAssignDisabled = !isAssignable || isTargetTeamMember;
+      const isAssignDisabled = !isAssignable || isTargetTeamMember || isLimitedRespondent;
       const isUnassignDisplayed = canAssign && isAssignable && (autoAssign || isAssigned);
 
       let assignTooltip: string | undefined;
@@ -217,6 +218,8 @@ export const useAssignmentsTab = ({
           : t('assignActivityDisabledTooltip');
       } else if (isTargetTeamMember) {
         assignTooltip = t('assignToTeamMemberTooltip');
+      } else if (isLimitedRespondent) {
+        assignTooltip = t('assignToLimitedRespondentTooltip');
       }
 
       const showDivider =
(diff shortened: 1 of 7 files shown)


src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
```diff
@@ -207,7 +208,7 @@ export const useAssignmentsTab = ({
       );
       const isAssignDisplayed =
         canAssign && (!autoAssign || status === ActivityAssignmentStatus.Hidden);
-      const isAssignDisabled = !isAssignable || isTargetTeamMember;
+      const isAssignDisabled = !isAssignable || isTargetTeamMember || isLimitedRespondent;
```
sultanofcardio, line 212: It's currently possible to access the assign menu item from the participants table more menu on a limited account. When this happens, the participant is pre-filled as the target of the assignment.

image image


I think it would make sense to take a similar approach on at least the "about participant" view of the assignments tab. Though if we just did the same thing for both views, that'd also be fine.

src/modules/Dashboard/features/Participant/Assignments/EmptyState/EmptyState.types.ts
```diff
@@ -4,4 +4,6 @@ export type EmptyStateProps = {
   icon: Icons;
   title: string;
   onClickAssign?: () => void;
+  /** @default '50.7rem' */
+  width?: string;
```
sultanofcardio: `maxWidth`?

```suggestion
  maxWidth?: string;
```

sultanofcardio (approved):
Works as expected. Approving with a couple suggestions

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
farmerpaul: Good catch. That's actually a regression caused by incorrect logic for determining if the respondent is a limited account. What you're describing is actually part of the original AC for [M2-7828](https://mindlogger.atlassian.net/browse/M2-7828).

In fixing that bug I noticed another bug that would mislead users into thinking activities/flows could hypothetically be assigned to a limited account as a respondent; see explanation in [this commit](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1974/commits/1221808051864316a18243ea245b401fa6b5a378).

PR #1979, opened 2024-11-21
chore: remove enable admin app soft lock feature flag (M2-7271)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-7271](https://mindlogger.atlassian.net/browse/M2-7271)

Deleted references to the `enableAdminSoftLock` feature flag.

Also deleted never-used type definitions for LORIS in our `featureFlags.ts` file that diverged from the standard way we define feature flags, and I guess never got used because `enableLorisIntegration` was eventually added using the standard pattern.

### 🪤 Peer Testing

1. Perform **Take Now** on any activity or flow
2. Complete the assessment in the web app and return to the admin app.
    **Expected outcome:** You should be soft-locked of the Admin App, same as Take Now normally works.


ChaconC (approved):
LGTM

We'll also need to archive the key once this is released

PR #1980, opened 2024-11-22
chore: Include extra data in call to JIRA webhook

sultanofcardio:
### 📝 Description

Whenever a release candidate deployment happens from this repository, a GitHub Actions workflow attempts to process the tickets it contains, and consumes a Jira webhook that triggers an automation to move those tickets into the "To Be Tested" column. I recently updated that automation to also leave a comment on the ticket when the deployment happens

image

This PR adds more details to the webhook call so that the comment may include more details about the deployment. These details are:
- The release candidate tag
- The repository URL

Once merged here and in the other repositories, I'll update the automation to include these details.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

Due to the external dependency on Jenkins, we won't be able to properly test this before merging. Please review the code change carefully

### ✏️ Notes

N/A

### DIFF
--- a/.github/workflows/update-jira-tickets.yaml
+++ b/.github/workflows/update-jira-tickets.yaml
@@ -103,6 +103,8 @@ jobs:
           echo "tickets=${jiraTickets}" >> $GITHUB_OUTPUT
 
       - name: Periodically ping Jenkins for current tag build status
+        env:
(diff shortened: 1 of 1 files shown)


andrevitalb (approved):
Nice and simple! Good addition

farmerpaul (approved):

PR #1987, opened 2024-12-05
fix: fixes datepicker issue within dataviz (M2-7348)

ramirlm:
### 📝 Description

🔗 [Jira Ticket M2-7348](https://mindlogger.atlassian.net/browse/M2-7348)

Dates without responses were being auto-selected, and the actual response day could not be selected. This behavior is being fixed and now the view loads the days correctly.

Changes include:

- Fixes for the calendar picker on Dataviz, interpolating a UTC 00:00 time into the date in order to avoid gaps

### 🪤 Peer Testing

- Create an applet, and add responses to it
- Open the participant view, click on **responses**
- Check if the responses are showing on the correct day
- Click on any response, and check if it opens correctly 

    **Expected outcome:** The responses are loading on the correct days, and the days without response are not being able to be selected
    
    The updated code only shows selectable options where the limited user has inputted answers for that activity, for that subject:
    
CleanShot 2024-12-10 at 08 58 57@2x

### ✏️ Notes

- Backend changes for this task were not necessary


farmerpaul (approved):

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -233,7 +233,9 @@ export const RespondentDataReview = () => {
 
     const lastActivityCompletedDate = lastActivityCompleted && new Date(lastActivityCompleted);
     const selectedDateInParam =
-      selectedDateParam && isValid(new Date(selectedDateParam)) && new Date(selectedDateParam);
+      selectedDateParam &&
```
ChaconC: This is addressing the date parsing issue while inconvenient and nice to finally have a fix for, is not the root cause of the issue in this ticket.

The primary issue is the endpoint is returning dates for other participants which when selected, don't contain any responses making it hard to determine which dates need to be reviewed.

https://github.com/user-attachments/assets/7e90f097-272c-4fba-8df2-6dfcdb701f9c

ChaconC (changes requested):

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
ramirlm: Will look into it @ChaconC – thanks!

farmerpaul: @ChaconC I saw that this was only a partial fix, too, and it's an important part of it, because the date offset was actually preventing you from accessing certain dates that you should be able to access for the current subject. I guess that to complete the fix (and before this can be be sent over to QA), the [separate BE ticket](https://mindlogger.atlassian.net/browse/M2-8215) also needs to be fixed. Is that right? Or are there more FE changes needed?

ChaconC: That's right @farmerpaul! Both parts need to be completed to close this one out

ramirlm: Correct @farmerpaul @ChaconC :)

ramirlm: Fixed, keeping as draft to wait for https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1680 to be merged.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -143,12 +143,14 @@ export const RespondentDataReview = () => {
...
 
+    // Map each date string to a Date object, appending 'T00:00:00' to remove timezone offset
     const submitDates = datesResult.dates.map((date: string) => new Date(`${date}T00:00:00`));
+
     setResponseDates(submitDates);
   });
 
   const handleGetSubmitDates = (date: Date) => {
-    if (!appletId || !subjectId) return;
+    if (!appletId || !subjectId || !selectedActivity?.id) return;
```
farmerpaul: We don't actually need activity ID to be defined as it's an optional parameter to `getAppletSubmitDateList` (but open to hearing your thoughts).
```suggestion
    if (!appletId || !subjectId) return;
```

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -251,7 +254,7 @@ export const RespondentDataReview = () => {
 
     handleSetInitialDate(new Date());
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [lastActivityCompleted]);
+  }, [lastActivityCompleted, selectedActivity?.id]);
```
farmerpaul: I'm assuming you discovered you needed to trigger the effect when the activity ID becomes available, but shouldn't that also be true for the flow ID?

```suggestion
  }, [lastActivityCompleted, selectedActivity?.id, selectedFlow?.id]);
```

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
ramirlm: This was necessary to make sure the method in the api is being called with a non-empty value for `activityOrFlowId` @farmerpaul

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
ramirlm: 🎯 @farmerpaul

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
farmerpaul: I'm not sure we should be restricting the `handleGetSubmitDates` to only instances of this screen where an activity has been selected. The `RespondentDataReview` screen is accessible in the app from several routes, including a route that doesn't include either an activity ID or flow ID as a parameter:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/c1ba7658fb4e4cb934bd84dfe0041d2be4e121a8/src/modules/Dashboard/routes/routes.tsx#L136-L147

https://github.com/ChildMindInstitute/mindlogger-admin/blob/a55547a273aa2798fa9b264f2481b71cbb860d40/src/resources/pages.json#L27

So it's still possible to invoke this component where activity ID or activity flow ID have not been defined. I was able to navigate to this URL after checking out your branch:

`https://localhost:3000/dashboard/8e9bd4f9-7ea3-4cfd-b6b1-040108774370/participants/83b7ff36-2175-4fd0-80c7-bc9f1d08ddee/dataviz/responses`

And because no default activity was selected, the `handleGetSubmitDates` callback was bypassed due to this condition, and resulted in a datepicker showing all dates as being available:



Now, whether the app is even supposed to still be calling this route is a separate question (I think it's deprecated), one which I think I need to bring up with product. But I think it's still accessible from a few spots, so we should probably make sure not to affect the datepicker when those routes are still being called.

So for now, I think I stand by my original suggestion of not changing this line:

```suggestion
    if (!appletId || !subjectId) return;
```

ramirlm:
@farmerpaul FYI still ongoing fixing this – I appreciate the insights though!

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -228,28 +236,28 @@ export const RespondentDataReview = () => {
...
-      return;
-    }
+    const responseDate = getValues('responseDate');
+    if (!responseDate) return;
 
-    handleSetInitialDate(new Date());
+    handleGetSubmitDates(responseDate);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [lastActivityCompleted]);
+  }, [activityOrFlowId]);
```
farmerpaul, line 260: I'm not sure I completely follow all the changes you're making in this effect, and introducing a second effect. It appears that with these changes, you're no longer referencing the `selectedDateParam` parameter, but it's an important part of this page's functionality. When this parameter is provided, its value must take priority over whatever the most recent response date is for the selected activity/flow.

I'm currently working on a related task, [M2-8362](https://mindlogger.atlassian.net/browse/M2-8362), and I discovered that `selectedDateParam` is being incorrectly converted to a date in the existing code, much like how you identified that the values of `datesResult.dates` were being incorrectly converted. I fixed it by employing a similar technique as you did [here](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1987/files#diff-9c32f7ebd71024c91037fdb69acfb323fd532c988821129aa024cf6b7a6ec64aR151):

```ts
    if (selectedDateParam) {
      const selectedDate = new Date(`${selectedDateParam}T00:00:00`);
      if (isValid(selectedDate)) {
        handleSetInitialDate(selectedDate);

        return;
      }
    }
```

Could you please make sure to handle the possible value of `selectedDateParam` using a snippet similar to what I suggested above? Let me know if you have any questions.

farmerpaul (changes requested):
Hi @ramirlm, I know that this ticket is still a WIP and I'm sure you'll resume work when you're ready to, but I just wanted to draw your attention to one detail I noticed about its current state.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
farmerpaul: @ramirlm – just wanted to give you the heads-up that I've included this change in [my own related PR](https://github.com/ChildMindInstitute/mindlogger-admin/pull/2015), which has now been merged. So when you rebase on `develop`, you'll likely need to carefully resolve a few conflicts here.

farmerpaul:
Closing in favour of #2053

PR #1989, opened 2024-12-05
feat: Add properties to SSI Phrase Builder analytics events (M2-8295)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8295](https://mindlogger.atlassian.net/browse/M2-8295)

This PR adds new SSI-related properties to these existing Mixpanel analytics events:
- `Applet Save click`
- `Applet edit successful`
- `Applet Created Successfully`

The new properties are:

> [!NOTE]
> For all item counts below, only items that are **not hidden** should be captured in the counts. Hidden items are not of interest to the data team and should be ignored.

- `Item Count`: Total count of all items in the applet
- `Phrase Builder Item Count`: Total count of Phrase Builder items in the applet (𝑥)
- `Count of Unique Items in Phrase Builders`: Sum of the counts of the unique items referenced within each Phrase
Builder item in the applet (𝑦)
- `Average Unique Items Per Phrase Builder`: The result of property (𝑦) divided by property (𝑥). *
- `Average Items Per Phrase Builder`: The number of "Item (Response)" fields contained in all Phrase Builders divided by property (𝑥). *
- `Average Text Boxes per Phrase Builder`: The number of "Text" fields contained in all Phrase Builders divided by property (𝑥). *
- `Average Line Breaks per Phrase Builder`: The number of "Line Break" fields contained in all Phrase Builders divided by property (𝑥). *

> [!NOTE]
> \* For any of these properties, where the result is a division by the # of Phrase Builders (𝑥): if (𝑥) is zero, then the value of the property should be null (not set).

### 🪤 Peer Testing

Refer to the [ticket description](https://mindlogger.atlassian.net/browse/M2-8295) for a detailed listing of how to trigger the relevant events and what properties should be attached to them.

**For each event:**

2. Open Mixpanel, and navigate to the Events tab.
3. Refresh if needed.
    **Expected outcome:** The event matching the **Event Name** column should have been added to the top of the list.
    **Expected outcome:** The custom properties for that event mentioned in the **Details** column should be listed.


farmerpaul:
@sultanofcardio of the PRs I've requested a review from you for, this one should be the priority. Thx! 🙏🙏

src/shared/utils/mixpanel/mixpanel.utils.ts
```diff
@@ -26,19 +26,49 @@ export const trackAppletSave = ({
...
 
   const event: AppletSaveEvent = {
     action,
     [MixpanelProps.AppletId]: appletId,
     [MixpanelProps.ItemTypes]: [...uniqueItemTypes],
+    [MixpanelProps.ItemCount]: itemCount,
+    [MixpanelProps.PhraseBuilderItemCount]: phraseBuilderItemCount,
+    [MixpanelProps.ItemsIncludedInPhraseBuilders]: itemsIncludedInPhraseBuilders,
+    [MixpanelProps.AverageItemsPerPhraseBuilder]: phraseBuilderItemCount
+      ? itemsIncludedInPhraseBuilders / phraseBuilderItemCount
```
sultanofcardio: This calculation can result in long decimals, which may not be desirable. We could do something like this instead

```suggestion
      ? parseFloat((itemsIncludedInPhraseBuilders / phraseBuilderItemCount).toFixed(2))
```

sultanofcardio (approved):

farmerpaul:
@sultanofcardio could I ask you to review this once again? The data team has requested adding some new properties, and to skip hidden items from item counts – please see description above. Thank you! 🙏🏻

src/shared/utils/mixpanel/mixpanel.utils.ts
```diff
@@ -27,40 +27,98 @@ export const trackAppletSave = ({
...
+
+  for (const activity of applet.activities) {
     if (activity.autoAssign) {
       autoAssignedActivityCount++;
     } else {
       manualAssignedActivityCount++;
     }
+
+    const visibleItemNames = new Set<string>();
+    const referencedItemNames = new Set<string>();
```
sultanofcardio: I think this `referencedItemNames` is a remnant of the previous implementation. It's shadowed in the inner loop and doesn't end up doing anything, so we can remove it

src/shared/utils/mixpanel/mixpanel.utils.ts
```diff
@@ -27,40 +27,98 @@ export const trackAppletSave = ({
...
+                phraseBuilderFieldCounts.itemResponse++;
+              }
+            } else if (field.type === 'sentence') {
+              phraseBuilderFieldCounts.sentence++;
+            } else if (field.type === 'line_break') {
+              phraseBuilderFieldCounts.lineBreak++;
+            }
+          }
+        }
+        itemsIncludedInPhraseBuilders += referencedItemNames.size;
```
sultanofcardio: What's being counted here is the _sum of the counts of the unique items referenced within each Phrase
Builder_, like you mention in f21c30e0. I trust that this is what it should actually be, but it's a little different from what the ticket description says

> Total count of unique items that were included as Item Response fields in Phrase Builder items in the applet

I can see that being misinterpreted by QA as what you had implemented before: _count of the unique items referenced across all Phrase Builders_

So we should probably update the ticket description.

Moreover, I think calling this `Items Included in Phrase Builders` is misleading. Given what it actually is, something like `Phrase Builder Unique Item Response Count` (or something like that) is more appropriate

src/shared/utils/mixpanel/mixpanel.utils.ts
```diff
@@ -27,40 +27,98 @@ export const trackAppletSave = ({
...
+  const phraseBuilderFieldCounts = {
+    itemResponse: 0,
+    sentence: 0,
+    lineBreak: 0,
+  };
   let autoAssignedActivityCount = 0;
   let manualAssignedActivityCount = 0;
-  for (const activity of applet.activities ?? []) {
-    itemTypes.push(...(activity.items ?? []).map((item) => item.responseType));
+  const itemTypes: ItemResponseType[] = [];
```
sultanofcardio: This array seems to only be used as the argument to a Set constructor. Maybe we could just make this a Set

sultanofcardio (approved):
A few non blocking comments, but looks good ✅

src/shared/utils/mixpanel/mixpanel.utils.ts
farmerpaul: Ah yes, the result of my not doing a careful enough rebase!

src/shared/utils/mixpanel/mixpanel.utils.ts
farmerpaul: I've updated the ticket description to redefine that parameter, and renamed it as suggested!

PR #1994, opened 2024-12-18
feat: Separate Extended IFCL Items into 3 groups (M2-8402)

ChaconC:

### 📝 Description

Add 2 additional feature flags that adjust the behavior from the `enableItemFlowExtendedItems` feature flag to limit which items are visible when creating Item Flow conditional logic, in order to allow for an increased rollout of the extended items.

🔗 [Jira Ticket M2-8402](https://mindlogger.atlassian.net/browse/M2-8402)

### 📸 Screenshots

| All Items Visible                      | G2 Enabled, G3 Disabled                                | G2 Disabled, G3 Enabled                                  |
| -------------------------------------- | ------------------------------------- | ------------------------------------- |
| CleanShot 2024-12-18 at 15 04 53 | CleanShot 2024-12-18 at 15 05 27 | CleanShot 2024-12-18 at 15 05 10 | 

### 🪤 Peer Testing

The `enableItemFlowExtendedItems` feature flag should be enabled to display the updated behavior.

* Create an applet with multiple item types, including "extended" support items. For example: Slider Select, Number Select, Date, Matrix Multi-Select. 
* Add Item Flow Conditional logic
* Items in the top row should be controlled by the 3 related feature flags.

### ✏️ Notes

In order to facilitate testing, you may adjust the return value in the `useFeatureFlags` hook to control which groups are visible.

### DIFF
--- a/src/modules/Builder/components/ConditionRow/ConditionRow.utils.tsx
+++ b/src/modules/Builder/components/ConditionRow/ConditionRow.utils.tsx
@@ -58,26 +58,52 @@ const scoreItemTypes = [
   ItemResponseType.SingleSelection,
   ItemResponseType.MultipleSelection,
 ];
-const itemFlowItemTypes = [
-  ...scoreItemTypes,
-  ItemResponseType.Date,
-  ItemResponseType.NumberSelection,
-  ItemResponseType.Time,
-  ItemResponseType.TimeRange,
-  ItemResponseType.SingleSelectionPerRow,
-  ItemResponseType.MultipleSelectionPerRow,
-  ItemResponseType.SliderRows,
+
+const itemFlowItemGroups = [
+  [...scoreItemTypes, ItemResponseType.NumberSelection], // Group 1
+  [ItemResponseType.Date, ItemResponseType.Time, ItemResponseType.TimeRange], // Group 2
+  [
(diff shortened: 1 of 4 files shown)


src/modules/Builder/components/ConditionRow/ConditionRow.utils.tsx
```diff
@@ -58,26 +58,52 @@ const scoreItemTypes = [
...
-  ItemResponseType.SliderRows,
+
+const itemFlowItemGroups = [
+  [...scoreItemTypes, ItemResponseType.NumberSelection], // Group 1
+  [ItemResponseType.Date, ItemResponseType.Time, ItemResponseType.TimeRange], // Group 2
+  [
+    ItemResponseType.SingleSelectionPerRow,
+    ItemResponseType.MultipleSelectionPerRow,
+    ItemResponseType.SliderRows,
+  ], // Group 3
```
felipeMetaLab: I believe the groups are understandable, so is the comments necessary ?
if you think the comments is helping to understand, I strongly believe that create other well named variable to represent those groups could be good.

felipeMetaLab (approved):
Every thing is looking good to me but I added a minor suggestion

PR #1996, opened 2024-12-19
fix: Version bumps to safely remove dependabot warnings (M2-7989)

ramirlm:
### 📝 Description

🔗 [Jira Ticket M2-7989](https://mindlogger.atlassian.net/browse/M2-7989)

- Updates the libraries `crypto-browserify`, `dompurify` and `axios` to the desired versions to avoid critical security warnings from dependabot
The remaining dependencies mentioned in the ticket are not able to be updated 

### 📸 Screenshots

Google Chrome 2024-12-18 21 14 56

### 🪤 Peer Testing

- Running locally requires executing `rm -rf node_modules && npm install` in the app directory
- The whole application should be tested, encrypt/decrypt operations must work properly and at the usual execution duration


package.json
```diff
@@ -21,15 +21,15 @@
     "@types/node": "^20.7.1",
     "@types/react": "^18.2.23",
     "@types/react-dom": "^18.2.8",
-    "axios": "^1.5.1",
```
ramirlm, line 24: Versions had to be kept as strict versions due to the errors reported on https://mindlogger.atlassian.net/browse/M2-7989?focusedCommentId=26970

felipeMetaLab (approved):
It looks good for me, I like to keep versions strict.

PR #1998, opened 2024-12-19
chore: Rename symbols – "respondent" ⇒ "subject"/"participant"

farmerpaul:
### 📝 Description

I've been finding it annoying dealing with mock variables named as if they are describing "respondents" when they're actually describing participants more generally. Similar confusion arises with some of the type names; for example, `RespondentDetails` representing the details of a subject entity on the BE (and not necessarily haven't anything to do with the subject having a "respondent" role).

So I thought that during a period when there isn't much parallel activity happening in this repo and low risk of conflicts, I opted to do some renaming of these symbols:

[`dashboard.types.ts`](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1998/files#diff-2958a24fb2a8c137ee0b146c134c80bc32eb48af38b89fff44b397ca348ce424):

- rename type `Respondent` ⇒ `Participant`
- rename type `RespondentDetail` ⇒ `ParticipantDetail`
- rename enum `RespondentStatus` ⇒ `ParticipantStatus`
- rename type `RespondentDetails` ⇒ `SubjectDetails`
  - This one I thought it made more sense to use "Subject" since it's a type that's so commonly used to represent a "subject" entity structure returned from the BE; open to feedback though

[`mock.ts`](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1998/files#diff-68be31d3aa680bf1e385d1094572c72dca1fe4a995fbc22a2b042bf9d2e6b5d5):

- rename mock variables `*Respondent*` ⇒ `*Participant*`
- make mocked participant data more descriptive where possible
- align naming/structure of `mockedFull*1` and `mockedFull*2` mock accounts

Several Dataviz-related tests:
- use `mockedFullSubject1` rather than `mockedFullParticipantId` where used incorrectly (didn't actually impact actual tests passing, but good practice and makes understanding API calls and tests easier for new devs; thanks @sultanofcardio!)

There is no Jira ticket explicitly associated with this change, I opted to do it as a side effect of writing PDP unit tests and trying to make the code make more sense.

### 🪤 Peer Testing

TSC should compile correctly and unit tests should pass, so no manual testing should be needed.

### DIFF
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackAssessment/FeedbackAssessment.test.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackAssessment/FeedbackAssessment.test.tsx
@@ -9,7 +9,7 @@ import {
   mockedApplet,
   mockedAppletId,
   mockedCurrentWorkspace,
-  mockedFullParticipantId,
+  mockedFullSubjectId1,
 } from 'shared/mock';
 import { Roles } from 'shared/consts';
 import { Item, initialStateData } from 'shared/state';
@@ -24,7 +24,7 @@ import { FeedbackForm } from '../Feedback.types';
 import { FeedbackAssessment } from './FeedbackAssessment';
 import { FeedbackAssessmentProps } from './FeedbackAssessment.types';
 
-const route = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
+const route = `/dashboard/${mockedAppletId}/participants/${mockedFullSubjectId1}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
 const routePath = page.appletParticipantDataReview;
 const preloadedState = {
   workspaces: {
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackAssessment/FeedbackAssessment.test.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackAssessment/FeedbackAssessment.test.tsx
@@ -9,7 +9,7 @@ import {
   mockedApplet,
   mockedAppletId,
   mockedCurrentWorkspace,
-  mockedRespondentId,
+  mockedFullParticipantId,
 } from 'shared/mock';
 import { Roles } from 'shared/consts';
 import { Item, initialStateData } from 'shared/state';
@@ -24,7 +24,7 @@ import { FeedbackForm } from '../Feedback.types';
 import { FeedbackAssessment } from './FeedbackAssessment';
 import { FeedbackAssessmentProps } from './FeedbackAssessment.types';
 
-const route = `/dashboard/${mockedAppletId}/participants/${mockedRespondentId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
+const route = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
 const routePath = page.appletParticipantDataReview;
 const preloadedState = {
   workspaces: {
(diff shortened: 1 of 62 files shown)


src/modules/Dashboard/state/Users/Users.schema.ts
```diff
@@ -1,8 +1,8 @@
...
-import { Respondent, RespondentDetails } from 'modules/Dashboard/types';
+import { Participant, SubjectDetails } from 'modules/Dashboard/types';
 
 export type UsersSchema = {
-  allRespondents: BaseSchema<{ result: Respondent[]; count: number } | null>;
-  respondentDetails: BaseSchema<{ result: RespondentDetails } | null>;
-  subjectDetails: BaseSchema<{ result: RespondentDetails } | null>;
+  allRespondents: BaseSchema<{ result: Participant[]; count: number } | null>;
+  respondentDetails: BaseSchema<{ result: SubjectDetails } | null>;
+  subjectDetails: BaseSchema<{ result: SubjectDetails } | null>;
```
sultanofcardio, line 7: Good call not renaming these. I honestly can't wait for these to be removed as `respondentDetails` never even gets populated anymore 🤞

src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackNotes/FeedbackNotes.test.tsx
```diff
@@ -26,8 +26,8 @@ const noteTestId = 'respondents-summary-feedback-notes-note';
 const newNoteValue = 'New note has been added';
 const mockedAnswerId = 'some-answer-id';
 const mockedSubmitId = 'some-submit-id';
-const routeWithAnswerId = `/dashboard/${mockedAppletId}/participants/${mockedRespondent}/dataviz/responses?selectedDate=2023-11-27&answerId=${mockedAnswerId}`;
-const routeWithSubmitId = `/dashboard/${mockedAppletId}/participants/${mockedRespondent}/dataviz/responses?selectedDate=2023-11-27&submitId=${mockedSubmitId}`;
+const routeWithAnswerId = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipant}/dataviz/responses?selectedDate=2023-11-27&answerId=${mockedAnswerId}`;
+const routeWithSubmitId = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipant}/dataviz/responses?selectedDate=2023-11-27&submitId=${mockedSubmitId}`;
```
sultanofcardio: I think these should be `mockedFullParticipant.details[0].subjectId` or `mockedFullParticipantDetail.subjectId`

```suggestion
const routeWithAnswerId = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantDetail.subjectId}/dataviz/responses?selectedDate=2023-11-27&answerId=${mockedAnswerId}`;
const routeWithSubmitId = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantDetail.subjectId}/dataviz/responses?selectedDate=2023-11-27&submitId=${mockedSubmitId}`;
```

Though the value apparently doesn't really matter to the test, we should probably fix it

src/modules/Dashboard/features/RespondentData/RespondentDataReview/Feedback/FeedbackAssessment/FeedbackAssessment.test.tsx
```diff
@@ -24,7 +24,7 @@ import { FeedbackForm } from '../Feedback.types';
 import { FeedbackAssessment } from './FeedbackAssessment';
 import { FeedbackAssessmentProps } from './FeedbackAssessment.types';
 
-const route = `/dashboard/${mockedAppletId}/participants/${mockedRespondentId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
+const route = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
```
sultanofcardio: Again, this seems to not matter to the test, but it should be a subject ID. Sigh, so much to clean up

```suggestion
const route = `/dashboard/${mockedAppletId}/participants/${mockedFullParticipantDetail.subjectId}/dataviz/responses?selectedDate=2023-11-27&answerId=0a7bcd14-24a3-48ed-8d6b-b059a6541ae4`;
```

sultanofcardio (approved):
Thanks for keeping the change set size to something relatively manageable. There are still some oddities with the names of some object properties and function parameter being out of sync with their types, but I think this is a step in the right direction.

Thanks for making this change 🙏

src/modules/Dashboard/state/Users/Users.schema.ts
farmerpaul: I know, I started digging into that again to wrap my head around what these actually represented, and how exactly they are used… and then gave up and just left them for now. But that can be another PR soon, maybe!

ChaconC (approved):

PR #1999, opened 2024-12-20
test: Add PDP Unit Tests (Part 1) (M2-8024)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8024](https://mindlogger.atlassian.net/browse/M2-8024)

This PR adds unit tests for:
- `useAssignmentsTab` hook, heavily used by both tabs on the PDP
- `ActivityFlowThumbnail` component

More tests specific to each page component (`AboutParticipant`, `ByParticipant`) will be coming later.

### 📸 Screenshots

CleanShot 2024-12-20 at 13 23 10@2x

CleanShot 2024-12-20 at 13 17 48@2x

### 🪤 Peer Testing

Unit tests should pass, so no manual testing should be needed.


src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+  describe('Edit', () => {
+    describe('should show or hide edit ability depending on role', () => {
+      test.each`
+        canEdit  | role                 | description
+        ${true}  | ${Roles.Manager}     | ${'editing for Manager'}
+        ${true}  | ${Roles.SuperAdmin}  | ${'editing for SuperAdmin'}
+        ${true}  | ${Roles.Owner}       | ${'editing for Owner'}
+        ${false} | ${Roles.Coordinator} | ${'editing for Coordinator'}
+        ${true}  | ${Roles.Editor}      | ${'editing for Editor'}
+        ${false} | ${Roles.Respondent}  | ${'editing for Respondent'}
```
sultanofcardio: Is the respondent role a valid test case for the admin portal?
```suggestion
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+            targetSubject={mockedOwnerSubject}
+          />,
+          {
+            ...renderOptions,
+            preloadedState: { ...renderOptions.preloadedState, ...getPreloadedState(role) },
+          },
+        );
+
+        const actionDots = screen.queryByTestId(`${testId}-activity-actions-dots`);
+        if (actionDots) fireEvent.click(actionDots);
```
sultanofcardio: If for whatever reason the test doesn't find the action dots, then they never get clicked and the action menu won't show up. Then the `canEdit=false` scenario becomes void.

I also suggest using `userEvent.click` over `fireEvent.click` in new tests where possible, as the testing library docs suggest that

```suggestion
        const actionDots = screen.getByTestId(`${testId}-activity-actions-dots`);
        await userEvent.click(actionDots);
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+
+    describe('should show or hide Export Data button depending on role', () => {
+      test.each`
+        canExportData | role                 | description
+        ${true}       | ${Roles.Manager}     | ${'Export Data for Manager'}
+        ${true}       | ${Roles.SuperAdmin}  | ${'Export Data for SuperAdmin'}
+        ${true}       | ${Roles.Owner}       | ${'Export Data for Owner'}
+        ${false}      | ${Roles.Coordinator} | ${'Export Data for Coordinator'}
+        ${false}      | ${Roles.Editor}      | ${'Export Data for Editor'}
+        ${false}      | ${Roles.Respondent}  | ${'Export Data for Respondent'}
```
sultanofcardio:
```suggestion
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+              targetSubject={mockedOwnerSubject}
+            />,
+            {
+              ...renderOptions,
+              preloadedState: { ...renderOptions.preloadedState, ...getPreloadedState(role) },
+            },
+          );
+
+          const actionDots = screen.queryByTestId(`${testId}-activity-actions-dots`);
+          if (actionDots) fireEvent.click(actionDots);
```
sultanofcardio:
```suggestion
          const actionDots = screen.getByTestId(`${testId}-activity-actions-dots`);
          await userEvent.click(actionDots);
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+    test('should show Export modal', async () => {
+      const { rerender } = renderWithProviders(
+        <UseAssignmentsHookTest
+          activityOrFlow={mockParticipantActivities.autoAssignActivity}
+          targetSubject={mockedOwnerSubject}
+        />,
+        renderOptions,
+      );
+
+      userEvent.click(screen.getByTestId(`${testId}-activity-actions-dots`));
```
sultanofcardio:
```suggestion
      await userEvent.click(screen.getByTestId(`${testId}-activity-actions-dots`));
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+  describe('Take Now', () => {
+    describe('should show or hide Take Now button depending on role', () => {
+      test.each`
+        canDoTakeNow | role                 | description
+        ${true}      | ${Roles.Manager}     | ${'Take Now for Manager'}
+        ${true}      | ${Roles.SuperAdmin}  | ${'Take Now for SuperAdmin'}
+        ${true}      | ${Roles.Owner}       | ${'Take Now for Owner'}
+        ${false}     | ${Roles.Coordinator} | ${'Take Now for Coordinator'}
+        ${false}     | ${Roles.Editor}      | ${'Take Now for Editor'}
+        ${false}     | ${Roles.Respondent}  | ${'Take Now for Respondent'}
```
sultanofcardio:
```suggestion
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+        const actionDots = screen.queryByTestId(`${testId}-activity-actions-dots`);
+        if (actionDots) userEvent.click(actionDots);
+
+        await waitFor(() => {
+          if (canDoTakeNow) {
+            expect(screen.getByTestId(`${testId}-activity-take-now`)).toBeVisible();
+          } else {
+            expect(screen.queryByTestId(`${testId}-activity-take-now`)).toBeNull();
+          }
+        });
```
sultanofcardio: If you await this `userEvent.click`, you won't need the `waitFor` below it

```suggestion
        const actionDots = screen.getByTestId(`${testId}-activity-actions-dots`);
        await userEvent.click(actionDots);

        if (canDoTakeNow) {
          expect(screen.getByTestId(`${testId}-activity-take-now`)).toBeVisible();
        } else {
          expect(screen.queryByTestId(`${testId}-activity-take-now`)).toBeNull();
        }
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+        renderOptions,
+      );
+
+      userEvent.click(screen.getByTestId(`${testId}-activity-actions-dots`));
+
+      await waitFor(() => {
+        const takeNowButton = screen.getByTestId(`${testId}-activity-take-now`);
+        expect(takeNowButton).toBeVisible();
+        expect(takeNowButton).toHaveAttribute('aria-disabled', 'true');
+      });
```
sultanofcardio:
```suggestion
      await userEvent.click(screen.getByTestId(`${testId}-activity-actions-dots`));

      const takeNowButton = screen.getByTestId(`${testId}-activity-take-now`);
      expect(takeNowButton).toBeVisible();
      expect(takeNowButton).toHaveAttribute('aria-disabled', 'true');
```

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
```diff
@@ -0,0 +1,916 @@
...
+            .querySelector('input');
+
+          const text = `${mockedFullSubject1.secretUserId} (${mockedFullSubject1.nickname}) (${mockedFullSubject1.tag})`;
+          expect(sourceSubjectInputElement).toHaveValue(populated[0] ? text : '');
+          expect(targetSubjectInputElement).toHaveValue(populated[1] ? text : '');
+        },
+      );
+    });
+
+    test('should pre-populate both respondent and subject in Take Now modal', async () => {
```
sultanofcardio: Is this a duplicate of the previous test?

sultanofcardio (approved):
Nice work. I think all of my comments are style related

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
farmerpaul: Hmm… I'm not too sure about that. For some background, I templated these role-based tests from another test suite, and after doing a search through the code I discovered there are many other test cases that check against this role.

I just re-reviewed the permissions chart from [Confluence](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/89423875/Roles+and+permissions#reference%3A):

image

I suspect `Roles.Respondent` represents the "Respondent Manager" column in this chart – not the "Respondent" column at the end. So I guess the idea is they are permitted to _manage_ respondents. However, it's no longer even possible to assign this role to Team Members since we added multi-informant support – we restricted the roles to just these ones:

CleanShot 2025-01-06 at 12 41 30

Since there's inherited legacy support for the "Respondent Manager" role, maybe we should just leave these test cases in? I suppose we could also do a search in the production database to see if anyone is actually even using it… but for now I think I'll just leave it in place, but open to hearing your thoughts!

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
farmerpaul: > If for whatever reason the test doesn't find the action dots, then they never get clicked and the action menu won't show up. Then the `canEdit=false` scenario becomes void.

If the action dots aren't found, that means all menu items were hidden (which may be the case for certain roles). I guess it doesn't make a lot of sense to then go and check if the desired menu item is then not found, but effectively the end result is the same: the menu item also doesn't exist if the action dots are missing altogether, so the test passes. I'll massage the logic slightly to skip the menu item check if the action dots are missing.

> I also suggest using `userEvent.click` over `fireEvent.click` in new tests where possible, as the testing library docs suggest that

I was on the fence about that. At first I did switch them all over to `userEvent` and saw how it was causing tests to then fail, even after I thought I'd introduced the required `await`s. Might have just not done it quite correctly. I figured it wasn't harmful to stick to `fireEvent` where possible, using `userEvent` only as needed, for the sake of simplicity/performance, but perhaps the performance impact is negligible and I should just suck it up and figure out why it caused tests to start failing. I'll take another look.

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
farmerpaul: Yup! 🙈

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
farmerpaul: Hmm… I think I take back some of what I said. I think that permissions chart is just mis-labelled/outdated – the "Respondent Manager" column probably represents the `Coordinator` role. So I'd then agree with you that the `Respondent` role is just invalid for the admin app, and is probably an inherently superfluous role given that any subject can be a respondent.

So… I'm going to take you up on your suggestion and remove these test cases, at least for these new tests. I'm not planning on touching other unit tests for now.

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.test.tsx
farmerpaul: Okay, I took your advice! Used `userEvent` in all places. Clearly I wasn't using it properly everywhere, so no wonder tests weren't always passing. I like that it allowed me to get away from using `waitFor` in several places. Thanks for the nudge @sultanofcardio!

Also improved the logic with the action dots.

PR #2002, opened 2025-01-03
feature: Reports fields rename (M2-8268)

andrevitalb:
- [X] Tests for the changes have been 

### 📝 Description

🔗 [Jira Ticket M2-8268](https://mindlogger.atlassian.net/browse/M2-8268)

Renaming & reordering for both `report.csv` and `activity_user_journey.csv` fields. Also, renamed `report.csv` to `responses.csv`.

All these changes were added behind a workspace-wide feature flag, so won't be directly enabled by default.

#### Current expected columns & order (for `report.csv`)
```plaintext
id
activity_flow_submission_id
activity_scheduled_time
activity_start_time
activity_end_time
flag
secret_user_id
userId
source_user_subject_id
source_user_secret_id
source_user_nickname
source_user_relation
source_user_tag
target_user_subject_id
target_user_secret_id
target_user_nickname
target_user_tag
input_user_subject_id
input_user_secret_id
input_user_nickname
activity_id
activity_name
activity_flow_id
activity_flow_name
item
response
prompt
options
version
rawScore
reviewing_id
schedule_id
timezone_offset
legacy_user_id
```

#### New expected columns & order
```plaintext
target_id
target_secret_id
target_nickname
target_tag
source_id
source_secret_id
source_nickname
source_tag
source_relation
input_id
input_secret_id
input_nickname
userId
secret_user_id
legacy_user_id
applet_version
activity_flow_id
activity_flow_name
activity_flow_submission_id
activity_id
activity_name
activity_submission_id
activity_start_time
activity_end_time
activity_schedule_id
activity_schedule_start_time
utc_timezone_offset
activity_submission_review_id
item_id
item_name
item_prompt
item_response_options
item_response
item_response_status
rawScore
```

The changes get propagated to `activity_user_journey.csv`. Additional fields for said report were left unchanged.

### 🪤 Peer Testing

In order to test functionality, set `enableDataExportRenaming` feature flag to `true`.

### ✏️ Notes

Something to keep in mind is that changes to subscales will come in a separate ticket, under the same feature flag that was used for all the changes added here


src/shared/utils/exportData/getReportCSVObject.test.ts
```diff
@@ -160,8 +160,63 @@ const result = {
...
 describe('getReportCSVObject', () => {
-  test('returns correct result', () => {
+  test('returns correct result with legacy naming', () => {
+    expect(
+      getReportCSVObject({
+        ...getPreparedProperties({
+          //eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          //@ts-ignore
+          activityItem: singleSelectionItem,
+          //eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
felipeMetaLab, line 209: Ditto, 
Is there any reason for disabling the eslint and TS here ?
Why is it necessary?
Is there a way to make it work without forcing the bypass?

src/shared/utils/exportData/getReportAndMediaData.test.ts
```diff
@@ -624,6 +605,95 @@ describe('getReportAndMediaData', () => {
         },
       ]);
     });
+    test('should return an array with items and subscale calculation in first item (new naming)', () => {
+      const rawAnswersObject = getObjectFromList(
+        mockedDecryptedAnswersWithSubscales,
+        (item) => item.activityItem.name,
+      );
+      //eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
felipeMetaLab, line 613: Ditto, 
Is there any reason for disabling the eslint and TS here ?
Why is it necessary?
Is there a way to make it work without forcing the bypass?

src/shared/utils/exportData/getJourneyCSVObject.test.ts
```diff
@@ -169,12 +169,72 @@ const result = {
...
+  response_option_selection_time: '1689755869391',
 };
 
 describe('getJourneyCSVObject', () => {
-  test('returns correct result', () => {
+  test('returns correct result with legacy naming', () => {
+    expect(
+      getJourneyCSVObject({
+        ...getPreparedProperties({
+          //eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
felipeMetaLab, line 224: Is there any reason for disabling the eslint and TS here ?
Why is it necessary?
Is there a way to make it work without forcing the bypass?

felipeMetaLab (commented):
Disabling eslint and TS

src/shared/utils/exportData/getJourneyCSVObject.test.ts
andrevitalb: This replicates pre-existing behavior for tests. The effort required to make the corresponding updates to remove these comments exceeds this ticket's scope. We could explore the option of making the corresponding updates to these tests at a later stage

felipeMetaLab (approved):
Beside the comment to disable Ts and eslint the code looks good for me.

src/shared/consts.tsx
```diff
@@ -312,11 +312,15 @@ export const CONDITION_TYPES_TO_HAVE_RANGE_VALUE = [
   ConditionType.OutsideOfTimes,
 ];
 
-export const GENERAL_REPORT_NAME = 'report';
+const LEGACY_GENERAL_REPORT_NAME = 'report';
+const NEW_GENERAL_REPORT_NAME = 'responses';
+
+export const getGeneralReportName = (enableDataExportRenaming: boolean) =>
```
mbanting: Not sure if capturing this condition in a function is needed. It's only called in one place (exportDataSucceed) and the use of a feature flag is temporary. Once the enableDataExportRenaming flag is archived, we'll need to remove all references (or params based on it) to it, so it's good keep this to only where this is needed. Suggest removing this function and simply determine this in exportDataSucceed

src/shared/utils/exportData/exportDataSucceed.ts
```diff
@@ -27,16 +27,18 @@ const exportProcessedData = async ({
...
   flankerItemsData,
   suffix,
-}: AppletExportData & { suffix: string }) => {
+  enableDataExportRenaming,
+}: AppletExportData & { suffix: string; enableDataExportRenaming: boolean }) => {
   await exportTemplate({
     data: reportData,
-    fileName: GENERAL_REPORT_NAME + suffix,
-    defaultData: reportData.length > 0 ? null : reportHeader,
+    fileName: getGeneralReportName(enableDataExportRenaming) + suffix,
```
mbanting: I suggest simply determine the report name here
```
fileName: (enableDataExportRenaming ? NEW_GENERAL_REPORT_NAME : LEGACY_GENERAL_REPORT_NAME) + suffix,
```

src/shared/consts.tsx
```diff
@@ -348,9 +352,53 @@ export const reportHeader = [
...
+  'item_name',
+  'item_prompt',
+  'item_response_options',
+  'item_response',
+  'item_response_status',
+  'rawScore',
 ];
 
-export const activityJourneyHeader = [
+export const getReportHeader = (enableDataExportRenaming?: boolean) =>
```
mbanting: Same comment as above. I realize this function is also used in the tests (`exportDataSucceed.test.ts`), but this function itself will likely be deleted once the flag is archived, so minimizing the use of this function in other places like the tests also make sense. Instead can you import `legacyReportHeader` and `newReportHeader` in the tests and compare against those values instead of relying on this function.

src/shared/consts.tsx
```diff
@@ -387,6 +435,52 @@ export const activityJourneyHeader = [
...
+  'press_popup_skip_time',
+  'press_popup_keep_time',
+  'press_back_time',
+  'press_undo_time',
+  'press_skip_time',
+  'press_done_time',
+  'response_option_selection_time',
+];
+
+export const getActivityJourneyHeader = (enableDataExportRenaming?: boolean) =>
```
mbanting: Same comment as above

src/shared/utils/exportData/exportDataSucceed.test.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 // @ts-nocheck
+import { getActivityJourneyHeader, getReportHeader } from 'shared/consts';
```
mbanting: Can this import  legacyReportHeader, reportHeader, legacyActivityJourneyHeader, and activityJourneyHeader instead

src/shared/consts.tsx
```diff
@@ -348,9 +352,53 @@ export const reportHeader = [
   'version',
   'rawScore',
   'reviewing_id',
+  'schedule_id',
+  'timezone_offset',
+  'legacy_user_id',
+];
+
+const newReportHeader = [
```
mbanting: Let's use `reportHeader` instead. One less change during flag cleanup.

src/shared/consts.tsx
```diff
@@ -387,6 +435,52 @@ export const activityJourneyHeader = [
   'version',
 ];
 
+export const newActivityJourneyHeader = [
```
mbanting: Let's use `activityJourneyHeader` instead. One less change during flag cleanup.

src/shared/utils/exportData/exportDataSucceed.ts
```diff
@@ -27,16 +27,18 @@ const exportProcessedData = async ({
...
+    defaultData: reportData.length > 0 ? null : getReportHeader(enableDataExportRenaming),
```
mbanting: Similar comment as above

src/shared/utils/exportData/exportDataSucceed.ts
```diff
@@ -27,16 +27,18 @@ const exportProcessedData = async ({
...
   });
   await exportTemplate({
     data: activityJourneyData,
     fileName: JOURNEY_REPORT_NAME + suffix,
-    defaultData: activityJourneyData.length > 0 ? null : activityJourneyHeader,
+    defaultData:
+      activityJourneyData.length > 0 ? null : getActivityJourneyHeader(enableDataExportRenaming),
```
mbanting: Same comment as above. Determine the headers here instead of relying on an external function that will be removed and unavailable when the flag is archived.

src/shared/utils/exportData/getJourneyCSVObject.test.ts
```diff
@@ -169,12 +169,72 @@ const result = {
   input_user_nickname: 'Mock input user',
   version: '2.0.0',
   event_id: null,
-  timezone_offset: '',
-  legacy_user_id: '',
+  timezone_offset: null,
+  legacy_user_id: null,
+};
+
+const newResult = {
```
mbanting: A bit of a nit, but let's just call this `result` as it is the result when the flag is on and also when it is ultimately archived. Minimizes cleanup when flag is removed.

src/shared/utils/exportData/getReportCSVObject.test.ts
```diff
@@ -160,8 +160,63 @@ const result = {
   legacy_user_id: '',
 };
 
+const newResult = {
```
mbanting: A bit of a nit, but let's just call this `result` as it is the result when the flag is on and also when it is ultimately archived. Minimizes cleanup when flag is removed.

mbanting (changes requested):
Nice work. Main comment is the flag is temporary. So try and minimize it's proliferation in the code to only where it's necessary. Reduces cleanup and risk later on when flag is archived and code needs to be updated to remove references to it, and where QA might not test to the same level as when the feature is first implemented. I pointed out a couple of places where this can be done.

andrevitalb:
On it 🤓

src/shared/consts.tsx
```diff
@@ -312,11 +312,12 @@ export const CONDITION_TYPES_TO_HAVE_RANGE_VALUE = [
   ConditionType.OutsideOfTimes,
 ];
 
-export const GENERAL_REPORT_NAME = 'report';
+export const LEGACY_GENERAL_REPORT_NAME = 'report';
+export const NEW_GENERAL_REPORT_NAME = 'responses';
```
mbanting: I missed this one. It's another small suggestion, but similar to my previous suggestion of continuing to use  using `reportHeader` instead of `newReportHeader`, I think it's ok to just use `GENERAL_REPORT_NAME`. Then during flag cleanup, only the removal of `LEGACY_GENERAL_REPORT_NAME` is needed.

mbanting (approved):
Great work @andrevitalb . I added one small suggestion I missed in my previous review. But pre-approving this already. Thank you!

PR #2003, opened 2025-01-03
fix: fixing crash on duplicated multi option saving on ifcl (M2-8323)

ramirlm:

### 📝 Description

🔗 [Jira Ticket M2-8323](https://mindlogger.atlassian.net/browse/M2-8323)

Changes include:

- Fix to the issue reported on the ticket when adding a condition with multi select after duplicating it

### 🪤 Peer Testing

- Log into the admin panel
- Go to Item Flow tab and add item flow Conditions:
Row1 = Option1
Row2 != Option2
- ALL → show Multiple selection item
- Save applet
- Start editing an applet
- Duplicate created activity
- Go ti Item Flow and set Row values
- Save applet 

    **Expected outcome:** The applet should not crash and should save the condition normally


src/modules/Builder/components/ConditionRow/Condition/SwitchCondition/SwitchCondition.tsx
```diff
@@ -126,8 +126,15 @@ export const SwitchCondition = ({
...
 
+        if (selectedItem.type === 'multiSelectRows') {
+          setValue(rowIndexName, e.target.value);
+
+          return;
+        }
+
         setValue(payloadName, payload);
       },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
felipeMetaLab, line 137: why is it necessary disabling eslint ?

ramirlm: Yes, the eslint config is a bit messed up as you can see below, so if I remove it it will add a warning, and If I fix it it will infinitely loop into this useEffect @felipeMetaLab

felipeMetaLab: I got it, but instead removing, is it possible fix the eslint rules ? 
if yes we could ask to create a ticket for it, what you think? 
is it complex ?

ramirlm: Yeah it's pretty complex since it is necessary to review all dependency arrays from the whole application :( 
Can you approve the PR for the moment @felipeMetaLab ? Thanks a lot! :)

felipeMetaLab (approved):

PR #2005, opened 2025-01-06
chore: Add Unit Tests for Configure Report to Select Scores or Raw Scores (M2-8036)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-8036](https://mindlogger.atlassian.net/browse/M2-8036)

This PR adds unit tests for the changes introduced in #1935 and #1962. As a precursor for implementing these, I did a couple quick type-related fixes.

Finally, I also took this opportunity to remove the husk pre-commit hook. I confirmed that it is not being used in any build pipelines

### 📸 Screenshots

N/A

### 🪤 Peer Testing

`npm run test`

### ✏️ Notes

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -119,10 +119,8 @@
         "eslint-plugin-react": "^7.31.10",
         "eslint-plugin-react-hooks": "^4.6.0",
         "eslint-plugin-unused-imports": "^3.0.0",
-        "husky": "^8.0.3",
         "jest-junit": "^16.0.0",
         "jest-mock-axios": "^4.7.3",
-        "lint-staged": "^14.0.1",
         "mock-local-storage": "^1.1.24",
         "prettier": "^3.0.3",
         "react-app-rewired": "^2.2.1",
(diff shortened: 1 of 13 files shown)


package-lock.json
farmerpaul, line 1: I'm guessing you didn't mean to remove this file…

src/modules/Dashboard/features/Applet/Schedule/ImportSchedulePopup/ImportSchedulePopup.utils.test.tsx
```diff
@@ -307,6 +307,8 @@ describe('prepareImportPayload', () => {
   test('returns prepared import payload with valid data', () => {
     const result = prepareImportPayload(mockUploadedEvents, mockAppletData, mockRespondentId);
 
+    const endOfCurrentYear = endOfYear(new Date()).getFullYear();
```
farmerpaul: I find the semantics here a bit confusing. `endOfCurrentYear` is simply just the _current year_, as far as I can deduce, not the date representing the end of the current year. Then below, this year is being used via string interpolation to manually generate the date representing the end of the current year. In the end, the call to `endOfYear` above isn't actually doing anything meaningful.

This would make more sense to me:

```tsx
const currentYear = new Date().getFullYear();

…

{
  endDate: `${currentYear}-12-31`,
  …
}
```

I'd either do that, or continue using `endOfYear` but make use of all the date information it returns to generate a formatted date.

src/shared/mock.ts
farmerpaul, line 1: 😍

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.test.tsx
```diff
@@ -39,43 +51,90 @@ const tableItems = [
...
+    });
+    mockUseParams.mockReturnValue({
+      appletId: mockedAppletId,
+      activityId: activity.id,
+    });
+    mockUseNavigate.mockClear();
+  });
+
   test('should render score', () => {
+    // @ts-expect-error Ignored temporarily
```
farmerpaul: Perhaps you're already addressing this, but just wanted to point out that these `@ts-expect-error` lines can be omitted if you just include:
```tsx
  isStaticActive: false,
```
in the `commonProps` definition above.

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.test.tsx
```diff
@@ -88,25 +147,487 @@ describe('ScoreContent', () => {
...
+        );
+        await userEvent.click(submitBtn);
+
+        const retryOkBtn = within(changeScoreIdPopup).getByTestId(
+          `${dataTestid}-change-score-id-popup-submit-button`,
+        );
+        expect(retryOkBtn.textContent).toBe('Ok');
+        await userEvent.click(retryOkBtn);
+      } catch {
+        // If it's not showing, that's fine
```
farmerpaul: Can you explain why this is fine? Is this unit test non-deterministic (can have multiple valid behaviours)? Or is it always expected to throw an error? If it's the latter, then I would opt to use `queryBy*` over `getBy*` to avoid throwing errors, allowing expected behaviour to be articulated a little more explicitly.

farmerpaul (approved):
Just a few minor comments and clarification questions, but looks good overall!

package-lock.json
farmerpaul: Oops, I automatically assumed the file was removed based on GitHub not rendering the diff. You didn't remove it, my bad! 😅

src/modules/Dashboard/features/Applet/Schedule/ImportSchedulePopup/ImportSchedulePopup.utils.test.tsx
sultanofcardio: This is fair. I did this at the last minute without thinking about it too much. I think I'll opt for the latter approach 🙏

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.test.tsx
sultanofcardio: Thanks. I planned to address these as part of a follow-up PR where I do more type fixes. At the time, I didn't want to dig into figuring out what's required to make it compile

src/modules/Builder/features/ActivitySettings/ScoresAndReports/ScoreContent/ScoreContent.test.tsx
sultanofcardio: It's somewhat non-deterministic. The change score ID popup is shown whenever a user tries to change the name of a score while the score's ID (which is generated based on its calculation type + name) is being referenced as a variable somewhere inside the score (e.g. in the message shown, conditional logic, etc).

When you run this test without this try-catch, it works only for the sum case because the score ID is already made based on sum, so there would be no potential change for which to prompt the user for confirmation.

I wrote the test like this because it depends on a mock that may change in a way that causes this popup to show (as it has since this test was written), but its visibility isn't important to the outcome of the test.

I can add this context to the test, or perhaps I can just override the mock locally to avoid this. Let me know what you think

farmerpaul: Thanks for the clarification! I'd be happy if you could just add this context in the form of a comment, sounds good to me.

sultanofcardio: As it turns out my mocking of some of the hooks used by the component was responsible for the change in the test outcome, which wasn't immediately clear either. So I opted to override the mock locally instead to avoid any confusion

PR #2007, opened 2025-01-08
fix: Adds text and message as possible flow sequences (M2-8286)

ramirlm:

### 📝 Description

🔗 [Jira Ticket M2-8286](https://mindlogger.atlassian.net/browse/M2-8286)

- As per discussion with the team, including also the possibility of choosing the sequence of the item flow to Text and Message items

### 🪤 Peer Testing

- Make sure the feature flag enableItemFlowExtendedItems is off.

- Log into the admin platform, create an applet and activity

- Include different types of items in the activity

- Create a new Flow for the activity

- Populate the trigger (1st box)

Expected outcome: Check whether the last box is populated with items other than legacy logic (Slider, Multiple Selection,  Single Selection, Text or Message )

### ✏️ Notes

- Based on the ticket https://mindlogger.atlassian.net/browse/M2-8248 status - Amplify should now be able to use DEV values, so it's possible to test this change on Amplify, changing the values on LD Dev.

### DIFF
--- a/src/modules/Builder/features/ActivityItemsFlow_old/ActivityItemsFlow.test.tsx
+++ b/src/modules/Builder/features/ActivityItemsFlow_old/ActivityItemsFlow.test.tsx
@@ -74,6 +74,13 @@ const mockedOrderedConditionNameItems = [
   mockedMultiActivityItem,
   mockedSliderActivityItem,
 ];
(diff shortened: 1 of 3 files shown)


andrevitalb (approved):
Looks good

PR #2008, opened 2025-01-08
feat: Add PDP Unit Tests (Part 2) (M2-8024)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8024](https://mindlogger.atlassian.net/browse/M2-8024)

This PR adds unit tests for:
- `AboutParticipant` screen component
- `ByParticipant` screen component

### 📸 Screenshots

CleanShot 2025-01-08 at 09 47 39

CleanShot 2025-01-08 at 09 47 59

### 🪤 Peer Testing

Unit tests should pass, so no manual testing should be needed.

### DIFF
--- a/src/shared/components/Svg/Svg.tsx
+++ b/src/shared/components/Svg/Svg.tsx
@@ -1,7 +1,7 @@
 import { SvgProps } from './Svg.types';
 
 export const Svg = ({ id, width = 24, height = 24, className, ...rest }: SvgProps) => (
-  <svg className={`svg-${id} ${className}`} width={width} height={height} {...rest}>
+  <svg className={`svg-${id} ${className || ''}`} width={width} height={height} {...rest}>
     <use xlinkHref={`#svg-${id}`} />
   </svg>
 );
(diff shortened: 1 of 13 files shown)


src/shared/mock.ts
sultanofcardio, line 1: Not a huge fan of index references across mocks, as the index of the referenced item can change. I think it's fine to leave these as-is though, just a heads up that I might change these in a future typing-related PR

src/shared/components/Svg/Svg.tsx
```diff
@@ -1,20 +1,7 @@
...
-}: SvgProps) => (
-  <svg
-    className={`svg-${id} ${className || ''}`}
-    width={width}
-    height={height}
-    fill={fill}
-    data-testid={dataTestid}
-  >
+export const Svg = ({ id, width = 24, height = 24, className, ...rest }: SvgProps) => (
+  <svg className={`svg-${id} ${className}`} width={width} height={height} {...rest}>
```
sultanofcardio: Won't this cause `undefined` to be appended to the end of all the classnames that don't provide an explicit one?
```suggestion
  <svg className={`svg-${id} ${className || ''}`} width={width} height={height} {...rest}>
```

sultanofcardio (approved):
One small comment but otherwise LGTM

src/shared/mock.ts
farmerpaul: Ah true… sounds good, thanks for the heads-up.

PR #2010, opened 2025-01-09
feature: Subscale fields renaming (M2-8458)

andrevitalb:

### 📝 Description

🔗 [Jira Ticket M2-8458](https://mindlogger.atlassian.net/browse/M2-8458)

This is a follow up on [M2-8268](https://mindlogger.atlassian.net/browse/M2-8268), containing the subscales changes.

All the renaming changes are behind the same feature flag that was introduced in #2002 (`enableDataExportRenaming`).

#### Current expected naming for subscales:
```plaintext
Final SubScale Score
Optional text for Final SubScale Score
{subscale name}
Optional text for {subscale name}
```

#### New expected naming for subscales:
```plaintext
activity_score
activity_score_lookup_text
subscale_name_{subscale_name}
subscale_lookup_text_{subscale name}
```

### 🪤 Peer Testing

In order to test functionality, set `enableDataExportRenaming` feature flag to `true`.

[M2-8268]: https://mindlogger.atlassian.net/browse/M2-8268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/shared/consts.tsx
```diff
@@ -348,9 +349,50 @@ export const reportHeader = [
...
+  'target_tag',
+  'source_id',
+  'source_secret_id',
+  'source_nickname',
+  'source_tag',
+  'source_relation',
+  'input_id',
+  'input_secret_id',
+  'input_nickname',
+  'userId',
```
gsch-cmi: is there a reason this can't be renamed 'user_id'? similarly with rawScore -> 'raw_score'?

seems like it is updated in the activity journey header, so it seems like it should be consistent across export tables.

src/shared/utils/exportData/getJourneyCSVObject.ts
```diff
@@ -217,56 +269,59 @@ export const getJourneyCSVObject = <T>({
...
-    timezone_offset: tzOffset,
-    legacy_user_id: legacyProfileId ?? '',
-  });
+  return getJourneyCSVReturn(
+    {
+      id: event.id,
+      activity_flow_submission_id: flowId ? submitId : '',
+      activity_scheduled_time: scheduledDatetime
+        ? convertDateStampToMs(scheduledDatetime)
+        : ActivityStatus.NotScheduled,
```
gsch-cmi: Rather than a string indication, which impacts the ability of analysis libraries like pandas or polars to infer the type correctly, could this be set to null or empty when the activity is not scheduled?

src/shared/utils/exportData/getSubscales.ts
```diff
@@ -217,24 +219,38 @@ export const getSubscales = (
...
+
   const parsedSubscales = subscaleSetting.subscales.reduce((acc: ParsedSubscale, item) => {
     const calculatedSubscale = calcScores(item, activityItems, subscalesObject, {});
-    acc[item.name] = calculatedSubscale[item.name].score;
-    if (calculatedSubscale?.[item.name]?.optionText) {
-      acc[`Optional text for ${item.name}`] = calculatedSubscale[item.name].optionText;
+    const cleanedName = cleanName(item.name);
+
+    if (enableDataExportRenaming) {
+      acc[`subscale_name_${cleanedName}`] = calculatedSubscale[item.name].score;
```
gsch-cmi, line 230: i'm not sure i'm understanding this correctly -- you're reducing over all of the subscales, but then accumulating into a dictionary, so presumably there is only one value collected per subscale over the entire list of activity items to be exported. is this called separately for each user to be included in the export, or how are different values for the subscale calculations stored?

gsch-cmi (commented):
Hi there, I'm a data engineer at CMI and have been working with the mindlogger exports a lot. Thanks so much for this work, this will be super helpful! I went through to understand the changes and left a few comments while I was going through with some questions I had. Thanks so much!

src/shared/utils/exportData/getSubscales.ts
andrevitalb: The functionality isn't changing for subscales. The way parsing happens is exactly the same. 

The only change here is a rename for the column header. Previously, it would be the straight value provided for `name`. After the update, naming will be changed to the same `name` value, with two changes:

- The `subscale_name_` prefix.
- Formatted by replacing all non-alphanumerical characters with `_`.

gsch-cmi: Right, I understand that, was just asking about the logic to help understand the flow

andrevitalb: Oh, my bad. A brief explanation:

`activityItems` contains a single user's responses to activity items.

This function would be called separately for each user's data during export, generating a new set of subscale scores for each user.

Each user gets their own row in the export, with their individual subscale calculations included in that row.

src/shared/consts.tsx
andrevitalb: A rename was considered, aligning both to snake case (`user_id`, `raw_score`), but current plans for those fields are to be deprecated, so the rename was not done.

src/shared/utils/exportData/getSubscales.test.ts
```diff
@@ -199,26 +205,39 @@ describe('getSubscales', () => {
...
-      ${itemsAndSubscalesWithAverageCalculation} | ${activityItems} | ${filledSubscaleScoresWithAverageCalculation} | ${'should return filled scores with average calculation: items and subscale'}
-      ${itemsOnly}                               | ${activityItems} | ${itemsOnlyfilledSubscaleScores}              | ${'should return filled scores: items only'}
-      ${itemsAndSubscales}                       | ${{}}            | ${{}}                                         | ${'should return empty object'}
-      ${{}}                                      | ${activityItems} | ${{}}                                         | ${'should return empty object'}
-      ${{}}                                      | ${{}}            | ${{}}                                         | ${'should return empty object'}
-      ${null}                                    | ${null}          | ${{}}                                         | ${'should return empty object'}
+      subscales            | activityItems    | expected                   | description
+      ${itemsAndSubscales} | ${activityItems} | ${newFilledSubscaleScores} | ${'should return filled scores: items and subscale'}
```
farmerpaul: I assume the legacy naming is due to be removed from the app at some later date, with the new naming being the replacement. I'm thinking it makes sense to ensure the new naming section has all the same tests as the legacy naming section does currently so we don't lose them when that happens. So something like this:

```ts
const newFilledSubscaleScoresWithAverageCalculation = {
  activity_score: 5,
  activity_score_lookup_text: 'Description #2 for range 4~20',
  'subscale_name_ss-1': 1.67,
};
const itemsOnlyNewFilledSubscaleScores = {
  activity_score: 5,
  activity_score_lookup_text: 'Description #2 for range 4~20',
  'subscale_name_ss-1': 5,
};

// …

      ${itemsAndSubscalesWithAverageCalculation} | ${activityItems} | ${newFilledSubscaleScoresWithAverageCalculation} | ${'should return filled scores with average calculation: items and subscale'}
      ${itemsOnly}                               | ${activityItems} | ${itemsOnlyNewFilledSubscaleScores}              | ${'should return filled scores: items only'}
      ${itemsAndSubscales}                       | ${{}}            | ${{}}                                            | ${'should return empty object'}
      ${{}}                                      | ${activityItems} | ${{}}                                            | ${'should return empty object'}
      ${{}}                                      | ${{}}            | ${{}}                                            | ${'should return empty object'}
      ${null}                                    | ${null}          | ${{}}                                            | ${'should return empty object'}
```

What do you think?

farmerpaul (approved):
Looks great @andrevitalb! I just had one minor comment about the unit tests, but otherwise good to go!

src/shared/utils/exportData/getSubscales.test.ts
andrevitalb: Totally makes sense! Updated :D

src/shared/utils/exportData/getJourneyCSVObject.ts
andrevitalb: Thanks for your comment. This will be considered as a user request. Further discussion will happen at a later point, which won't be comprised in this PR's scope 😄

PR #2015, opened 2025-01-17
fix: Navigate to activity-specific responses tab from data point (M2-8362)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8362](https://mindlogger.atlassian.net/browse/M2-8362)

After multi-informant was introduced last year, the DataViz **Responses** tab was redesigned to list responses only for the selected activity or flow, rather than all activities/flows that the subject has responses about. New routes were introduced to support this, and the old activity/flow-agnostic route was supposed to have been deprecated. However, the old route was still being used when clicking the **View response** link from the tooltip of a data point on the DataViz **Summary** tab.

This PR updates that navigate statement to respect the activity/flow ID parameter and use the respective new route, similarly as is done elsewhere in the app. The old route is deprecated with a comment (though not removed completely due to the slim possibility of bookmarks referencing this URL).

While I was making this change, I also fixed some incorrect parsing of dates happening on the Responses tab, which was causing an off-by-one error on available dates as well as the preselected date in the datepicker depending on the time zone/time of day.

### 📸 Screenshots

| <video src="https://github.com/user-attachments/assets/20455167-b29a-4f0d-910f-f364623280a8"></video> | <video src="https://github.com/user-attachments/assets/0c60a37c-2e07-4d8d-b733-ef0c338c01e7"></video> |

### 🪤 Peer Testing

- Navigate to PDP for a subject that that has existing submissions on an activity or flow.
- Click **View Data** for that subject.
- Hover over a data point in the Activity Completed chart to reveal the tooltip, then click **View response**.
    **Expected outcome:** The left sidebar shows responses only the the selected activity/flow (which can be confirmed if the subheading is **Select response** rather than **Select activity and response**).


src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -143,7 +143,8 @@ export const RespondentDataReview = () => {
     const datesResult = data?.result;
     if (!datesResult) return;
 
-    const submitDates = datesResult.dates.map((date: string) => new Date(date));
+    // Map each date string to a Date object, appending 'T00:00:00' to remove timezone offset
+    const submitDates = datesResult.dates.map((date: string) => new Date(`${date}T00:00:00`));
```
sultanofcardio, line 147: @farmerpaul do you have a demo of the issue this fixes? I'm not sure what it is

farmerpaul: I may have to wait until Monday morning to demonstrate it, as it's not happening right now presumably because I'm outside of the time window required to reproduce based on my time zone's difference with UTC. Might work if I set my machine to a different time zone, but my half day's over now so I'll just leave it till Monday.

For context, Ramir originally came up with the [fix](https://github.com/ChildMindInstitute/mindlogger-admin/pull/1987/files#diff-9c32f7ebd71024c91037fdb69acfb323fd532c988821129aa024cf6b7a6ec64aR151), which I confirmed was a partial solution for his task while reviewing his PR; I just copied his fix into mine and applied the same technique to the passed date parameter, which required the same adjustment.

farmerpaul: I'm able to reproduce it again this morning. Here's the incorrect behaviour (note how no initial response is loaded when accessing the Responses tab, and the preselected date and available dates in the datepicker are offset by 1 day):


https://github.com/user-attachments/assets/ff77f546-8737-4b0a-acb2-042ad4dab0b1

This is how it's supposed to work, and what this PR fixes:


https://github.com/user-attachments/assets/f81a5198-bbff-48f5-af9a-ce6b87c465a4

sultanofcardio: Looks good, thanks

sultanofcardio (approved):
Lgtm

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -232,13 +233,15 @@ export const RespondentDataReview = () => {
     if (lastActivityCompleted === undefined) return;
 
     const lastActivityCompletedDate = lastActivityCompleted && new Date(lastActivityCompleted);
-    const selectedDateInParam =
-      selectedDateParam && isValid(new Date(selectedDateParam)) && new Date(selectedDateParam);
 
-    if (selectedDateInParam) {
-      handleSetInitialDate(selectedDateInParam);
+    if (selectedDateParam) {
+      // Map each date string to a Date object, appending 'T00:00:00' to remove timezone offset
```
sultanofcardio:
```suggestion
      // Map the date string to a Date object, appending 'T00:00:00' to remove timezone offset
```

PR #2020, opened 2025-01-24
fix: update Time and TimeRange condition time parsing (M2-8306)

ChaconC:
### 📝 Description

Update Time parsing to fix an application crash when adding Time and TimeRange conditions

🔗 [Jira Ticket M2-8306](https://mindlogger.atlassian.net/browse/M2-8306)

### 📸 Screenshots

| CleanShot 2025-01-24 at 11 51 53 | CleanShot 2025-01-24 at 11 50 19 |

### 🪤 Peer Testing

- Create an Activity with 2+ items, including Time and/or TimeRange items
- Add Conditional Logic based on Time or TimeRange
- Select a Time
- Click Save
- Item should save correctly

### ✏️ Notes

The `enableItemFlowItemsG2` and `enableItemFlowExtendedItems`  feature flags need to be enabled in order to test Time/TimeRange conditions

### DIFF
--- a/src/modules/Builder/components/ConditionRow/ConditionRow.utils.tsx
+++ b/src/modules/Builder/components/ConditionRow/ConditionRow.utils.tsx
@@ -362,7 +362,11 @@ export const getPayload = ({
           DEFAULT_PAYLOAD_MAX_VALUE,
       };
     default:
-      return {};
+      return formTimeType
+        ? {
+            fieldName: formTimeType,
+          }
+        : {};
   }
 };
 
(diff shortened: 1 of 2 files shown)


farmerpaul (approved):
Though I'm not super familiar with the code, peer testing passed, so happy to approve.

As noted in my DM to you, I was not able to see the conditional logic work when I actually performed the activity (the target item was skipped even when the conditional logic was met), but as you pointed out, that's to be investigated in [M2-8302](https://mindlogger.atlassian.net/browse/M2-8302).

PR #2023, opened 2025-01-29
fix: update minDate conditional logic (M2-8317, M2-8275, M2-8032)

ChaconC:
### 📝 Description

Fix `minDate` condition when selecting `BETWEEN_DATES` or `OUTSIDE_OF_DATES` conditional logic by using the correct value and addressing a date parsing issue.

🔗 [Jira Ticket M2-8317](https://mindlogger.atlassian.net/browse/M2-8317)
🔗 [Jira Ticket M2-8275](https://mindlogger.atlassian.net/browse/M2-8275)
🔗 [Jira Ticket M2-8032](https://mindlogger.atlassian.net/browse/M2-8032)

Changes include:

- Use `minDateValue` instead of `minValue` when setting up `DatePicker`
- Update date parsing to remove timezone offset effect
- Additionally prevents setting a `minDate` greater than `maxDate`

### Case 1 (M2-8317):
**Before:** It's not possible to select a date prior to "today" in the second DatePicker when selecting Outside/Between conditional logic

https://github.com/user-attachments/assets/e15f37f9-4cd6-402a-a335-e116e55cea62

**After:** It's now possible to select a date prior to "today" but greater than the first date when selecting Outside/Between conditional logic

 https://github.com/user-attachments/assets/329c79e3-6b61-40b2-896c-41252a453236

### Case 2 (M2-8275):
Date parsing error after saving 

**Before:** 
- After selecting a date and pressing "Save", the date would change depending on the timezone (29 -> 28 in the video.)
- Clicking "Save" a second time would display an error due to a date parsing issue

https://github.com/user-attachments/assets/cf67f1ec-bc8a-45ab-8b96-3ecdd4d03706

**After:**
- Selecting a date and clicking "Save" results in the date not changing
- Date parsing error no longer appears, clicking Save a second time correctly reports no changes

https://github.com/user-attachments/assets/ae237a87-3e55-4b72-bf20-89678f288b09

### 🪤 Peer Testing

- Create an applet with 2 or more items, first item should be "Date" type
- Create conditional logic using the date item in the top row
- Select different values between Greater Than, Less Than, Outside or Between
  - For Outside/Between dates condition, it should be possible to select a date before "today" but greater than the first date.
- Applet should save correctly

### ✏️ Notes

- The `enableItemFlowItemsG2` and `enableItemFlowExtendedItems` feature flags need to be enabled in order to test **Date** conditional logic.

- ~Web app conditional logic for Date/Time is currently not working as expected, will be addressed by M2 8302~
- Web app conditional logic can be tested with this PR/Amplify environment https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/573

### DIFF
--- a/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
+++ b/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
@@ -44,6 +44,7 @@ import {
   getTextBetweenBrackets,
   INTERVAL_SYMBOL,
   isSystemItem,
+  parseDateToMidnightUTC,
   Path,
   pluck,
 } from 'shared/utils';
@@ -788,7 +789,7 @@ const formatTime = (hours: number, minutes: number) => {
 };
 
 const formatDate = (dateValue: string) => {
-  const date = dateValue ? new Date(`${dateValue}T00:00:00`) : undefined;
+  const date = dateValue ? parseDateToMidnightUTC(dateValue) : undefined;
 
   return date;
 };
--- a/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
+++ b/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
@@ -33,6 +33,8 @@ import {
   TimeRangeIntervalValueCondition,
   TimeSingleValueCondition,
   TimeRangeSingleValueCondition,
+  DateSingleValueCondition,
+  DateRangeValueCondition,
 } from 'shared/state';
 import {
   createArray,
@@ -96,6 +98,8 @@ import {
 import {
   ALLOWED_TYPES_IN_VARIABLES,
   CONDITION_TYPES_TO_HAVE_OPTION_ID,
+  DATE_INTERVAL_CONDITION_TYPES,
+  DATE_SINGLE_CONDITION_TYPES,
   defaultFlankerBtnObj,
   ordinalStrings,
   SAMPLE_SIZE,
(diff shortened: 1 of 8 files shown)


mbanting (approved):
Works as expected. Code looks good. Thanks for fixing this!

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -783,8 +787,31 @@ const formatTime = (hours: number, minutes: number) => {
   return `${formattedHours}:${formattedMinutes}`;
 };
 
+const formatDate = (dateValue: string) => {
+  const date = dateValue ? new Date(`${dateValue}T00:00:00`) : undefined;
+
+  return date;
+};
```
farmerpaul: We do this in a couple places in the app already:
- https://github.com/ChildMindInstitute/mindlogger-admin/blob/f8d4c66b64650ecca284423efbd3b46e761a2984/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx#L147
- https://github.com/ChildMindInstitute/mindlogger-admin/blob/f8d4c66b64650ecca284423efbd3b46e761a2984/src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx#L239

As well as in the `CalendarEvents.utils.test.ts` [unit tests](https://github.com/ChildMindInstitute/mindlogger-admin/blob/b69abd8b45abc0ee5d9b55869f3c6ba4551e90df/src/modules/Dashboard/state/CalendarEvents/CalendarEvents.utils.test.ts#L294-L307).

So I'd propose we create a shared utility – maybe call it `normalizeDateString` – move it to `src/shared/utils`, and make use of it in these other places, too. Also add a JSDoc explaining the expected string format: `YYYY-MM-DD`. What do you think?

farmerpaul (approved):
Tested and everything looks fixed, nice work. Just one suggestion, but preapproving.

src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
ChaconC: Nice callout @farmerpaul. I did something similar in the web-repo so it also makes sense to add it here, I’ll update this PR to include the changes

farmerpaul: Oh, just reviewing your other PR now, where it looks like you've actually [done just this](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/573/files#diff-7206aa05b0ccb436c6f374b7ff8cdab398b192e1140f6e57d5a2e43f59255faa). I'm okay with the name you've chosen in the web app, `parseDateToMidnightUTC ` – can we incorporate the same shared utility in the Admin App, with the added validation check? Then you can copy the unit tests over for free. 🙂

ChaconC: Done, only updated the tests where there was string interpolation and not where `T00:00:00` is being added as part of a fully formed date.

PR #2027, opened 2025-02-12
feat: New score calculation behaviour for skipped items (M2-8684)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8684](https://mindlogger.atlassian.net/browse/M2-8684)

With the Report Server now interpreting skipped answers as `null` rather than `0` (with the `enable-subscale-null-when-skipped` feature flag enabled), it's important for Dataviz and Data Export to align with this new interpretation as well – specifically around scoring calculations.

This task is to bring the Admin Panel inline with the changes to the Report Server to ensure that, when the feature flag is enabled, score calculations based on missing responses:

-   if **all items** associated with a score calculation (either subscale score, or activity total score) were skipped:
    -   do not have a corresponding data point in Dataviz
    -   do not have a value in the corresponding column of the exported data
-   if **any item** associated with an **average** score calculation (either subscale score, or activity total score) is skipped, it is **excluded** from the average score calculation

#### Preconditions:

-   Applet containing activity with **two skippable items** that have scoring enabled, with each one configured to be able to produce a score of `0` or `10` based on the response
-   Activity is configured with a subscale calculated as an **average** of both items, with lookup table that has these raw score ⇒ T-score mappings: `0 ⇒ 0`, `5 ⇒ 10`, `10 ⇒ 20`
-   Subscale has **Calculate total score** enabled, using **Average of Item Scores**:
    

#### Testing procedure:

**New behaviour:** _Skipped items are excluded from scores_
1. Add this to line 26 of `useFeatureFlags.ts` (feature flag **enabled**):
    ```ts
    features.enableSubscaleNullWhenSkipped = true;
    ```
2. As an eligible subject, perform an assessment of the activity in the Web App, skipping all items.
3. In the Admin App in another tab, open the PDP for that subject and click View Data to access Dataviz.
4. Adjust the date to restrict results to today's date, and inspect the subscale score chart.
    **Expected outcome:** No subscale or `activity_score` data points for the most recent submission should be on the chart.
5. Perform another assessment of the activity in the Web App, this time **choosing a response for the first item** that scores `10`, and **skipping the second item**.
6. Refresh Dataviz, restrict results to today's date, and inspect the subscale score chart.
    **Expected outcome:** There should be 2 data points for the most recent submission:
      - for the subscale, the data point should be `20` (the average of all responses' raw scores = single score of `10` = average score of `10` ⇒ T-score of `20`)
      - for the `activity_score`, the data point should be `10` (the average of all responses' raw scores = single score of `10` = average score of `10`)
7. Export the data from the last 24 hours and inspect the CSV file.
    **Expected outcome:** Find each group of rows representing **each submission**, and look at the **topmost row** of each group:
      - For the **1st submission**: the `activity_score` column should be **empty**, and the `subscale_name_[your subscale name]__Average` column should be **empty**
      - For the **2nd submission**: the `activity_score` column should contain `10` and the `subscale_name_[your subscale name]__Average` column should contain `20`

**Confirm legacy behaviour:** _Skipped items have a score of `0`_

1. Edit line 26 of `useFeatureFlags.ts` (feature flag **disabled**):
    ```ts
    features.enableSubscaleNullWhenSkipped = false;
    ```
2. Refresh Dataviz, restrict results to today's date, and inspect the subscale score chart.
    **Expected outcome:**
      - There should be 2 data points for the 1st submission, both with a value of `0` (skipped scores default to `0`)
      - There should be 2 data points for the 2nd submission:
        - for the subscale, the data point should be `10` (the average of all responses' raw scores = [`10`, `0`] = average score of `5` ⇒ T-score of `10`)
        - for the `activity_score`, the data point should be `5` (the average of all responses' raw scores = [`10`, `0`] = average score of `5`)
3. Export the data from the last 24 hours and inspect the CSV file.
    **Expected outcome:** Find each group of rows representing **each submission**, and look at the **topmost row** of each group:
      - For the **1st submission**: the `activity_score` column should contain `0`, and the `subscale_name_[your subscale name]__Average` column should contain `0`
      - For the **2nd submission**: the `activity_score` column should contain `5` and the `subscale_name_[your subscale name]__Average` column should contain `10`

### DIFF
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/Subscales/Subscales.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/Subscales/Subscales.tsx
@@ -32,10 +32,8 @@ export const Subscales = ({
   flowResponsesIndex,
 }: SubscalesProps) => {
   const { currentActivityCompletionData } = useContext(ReportContext);
-  const {
-    featureFlags: { enableDataExportRenaming },
-  } = useFeatureFlags();
-  const finalSubscale = enableDataExportRenaming ? FinalSubscale : LegacyFinalSubscale;
+  const { featureFlags: flags } = useFeatureFlags();
+  const finalSubscale = flags.enableDataExportRenaming ? FinalSubscale : LegacyFinalSubscale;
 
   const { finalScores, latestFinalScore, allSubscalesScores, allSubscalesToRender } = useMemo(
     () =>
@@ -55,9 +53,7 @@ export const Subscales = ({
           const calculatedTotalScore =
             item?.subscaleSetting?.calculateTotalScore &&
             activityItems &&
-            calcTotalScore(item.subscaleSetting, activityItems, enableDataExportRenaming)?.[
-              finalSubscale.Key
-            ];
+            calcTotalScore(item.subscaleSetting, activityItems, flags)?.[finalSubscale.Key];
 
           if (calculatedTotalScore) {
             acc.latestFinalScore = calculatedTotalScore?.score;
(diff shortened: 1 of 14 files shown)


sultanofcardio (approved):
Works as described, though I did note two things:
- The bar chart looks the same when there is a score of 0 and a skipped score (no bar), so we may not be able to tell them apart
- The bar chart is present, but empty when there are all skipped questions. However, the line chart does not show up when there are all skipped questions

farmerpaul:
Thanks @sultanofcardio for pointing those things out! I'm not sure how to deal with the ambiguity mentioned in your first bullet. I just think that a bar chart is inappropriate if it's important to represent `0` clearly in Dataviz. So far CAMHI has not complained about that. 🤷🏻‍♂️ 

To your second bullet: the impetus for all this work is really just to ensure the reporting conditional logic works correctly, and in turn I had to update Dataviz to align with the reported scores. But there's no requirement to hide either reports or score charts if all answers were skipped (and is probably an edge case anyway). If that comes along later, we can address it then.

PR #2036, opened 2025-02-19
feat: filter skipped responses in dataviz (M2-8669)

ChaconC:

### 📝 Description

Adds a toggle to the More Filters section in DataViz which allows hiding skipped answers from data tables. Enabling this option persists the setting per applet in session storage.

🔗 [Jira Ticket M2-8669](https://mindlogger.atlassian.net/browse/M2-8669)

This setting will only filter answers for item types using DataTables:

- Short Text
- Paragraph Text
- Date
- Time Range

### 📸 Screenshots

* New toggle under "More filters"

CleanShot 2025-02-19 at 12 11 59

* Data Tables

| Hide Skipped **Disabled**                      | Hide Skipped **Enabled**                                 |
| CleanShot 2025-02-19 at 12 12 11 | CleanShot 2025-02-19 at 12 12 19 |

* Data Table is omitted if there are no responses

| Hide Skipped **Disabled**                      | Hide Skipped **Enabled**                                 |
| CleanShot 2025-02-19 at 12 22 31 | CleanShot 2025-02-19 at 12 22 36 |

### 🪤 Peer Testing

- Create 2 applets with 1 or more compatible items (Short Text, Paragraph Text, Date, Time Range)
  - At least 1 item should have be skippable, or have conditional logic which causes the item to be skipped
CleanShot 2025-02-19 at 12 48 49
 
- Submit 1 or more responses for each applet, making sure to skip answering the test item
- Navigate to Participants > View Data
- Click "More Filters" to show the "Hide Skipped" toggle and enable the setting
- Empty answers should be removed from the data tables, including item counts and pagination
  - ❗  Leave the setting **enabled**
- Navigate back to the participant details
- Reopen DataViz, "Hide Skipped" should default to enabled
- Navigate to a different applet and open DataViz
  - ❗   Hide Skipped should be **disabled** (default)
- Logging out will clear the session storage resetting all applets back to default

### DIFF
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
@@ -21,3 +21,14 @@ export const StyledMoreFilters = styled(Button)`
   height: 5.5rem;
   margin-left: ${theme.spacing(1.2)};
 `;
+
+export const StyledCheckboxTitle = styled(StyledBodyMedium)`
+  display: flex;
+  align-items: center;
+
+  svg {
+    height: 1.6rem;
+    margin-left: ${theme.spacing(0.4)};
+    margin-top: ${theme.spacing(0.4)};
+  }
+`;
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
@@ -28,7 +28,6 @@ export const StyledCheckboxTitle = styled(StyledBodyMedium)`
 
   svg {
     height: 1.6rem;
-    margin-left: ${theme.spacing(0.4)};
-    margin-top: ${theme.spacing(0.4)};
+    display: block;
   }
 `;
--- a/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.tsx
+++ b/src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.tsx
@@ -236,7 +236,7 @@ export const ReportFilters = ({
                 data-testid={`${dataTestid}-versions`}
               />
             </Box>
-            <Box sx={{ width: '24rem', ml: theme.spacing(2.4) }}>
+            <Box sx={{ whiteSpace: 'nowrap', ml: 2.4 }}>
               <CheckboxController
                 name="hideSkipped"
                 control={control}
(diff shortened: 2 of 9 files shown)


src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.tsx
```diff
@@ -214,6 +236,25 @@ export const ReportFilters = ({
                 data-testid={`${dataTestid}-versions`}
               />
             </Box>
+            <Box sx={{ width: '24rem', ml: theme.spacing(2.4) }}>
```
farmerpaul: Nit: Just wondering if it's needed to prescribe a static width here. Some localized strings may not be suited to 240 pixels, and/or the parent flexbox may not always respect this value based on viewport constraints. If the idea is to prevent word wrapping, I'd do the following instead (also you can use the shorthand instead of calling `theme.spacing` here):

```suggestion
            <Box sx={{ whiteSpace: 'nowrap', ml: 2.4 }}>
```

src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.tsx
```diff
@@ -214,6 +236,25 @@ export const ReportFilters = ({
...
+                    {t('hideSkipped')}
+                    <Tooltip tooltipTitle={t('hideSkippedTooltip')}>
+                      <span>
+                        <StyledTitleTooltipIcon id="more-info-outlined" />
+                      </span>
+                    </Tooltip>
+                  </StyledCheckboxTitle>
+                }
+                data-testid={`${dataTestid}-hide-skipped`}
+                sx={{ mr: theme.spacing(0) }}
```
farmerpaul:
```suggestion
                sx={{ mr: 0 }}
```

src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
```diff
@@ -21,3 +21,14 @@ export const StyledMoreFilters = styled(Button)`
...
 `;
+
+export const StyledCheckboxTitle = styled(StyledBodyMedium)`
+  display: flex;
+  align-items: center;
+
+  svg {
+    height: 1.6rem;
+    margin-left: ${theme.spacing(0.4)};
+    margin-top: ${theme.spacing(0.4)};
```
farmerpaul: Nit: the misalignment can be better corrected by the display mode than by adding magic margin values:
```suggestion
    display: block;
```

farmerpaul (approved):
Just a couple styling nits, but otherwise good to go!

src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.styles.ts
ChaconC: The `margin-left` was because the icon is slightly too far from the label, but overall I like your suggestion better, thanks!

src/modules/Dashboard/features/RespondentData/RespondentDataSummary/Report/ReportFilters/ReportFilters.tsx
ChaconC: No reason at all, was basing it off the other components in this, but happy to accept this 👍

PR #2038, opened 2025-02-20
fix: update date parsing from backend (M2-8720)

ChaconC:
### 📝 Description

Changed how dates are parsed from the backend. 

🔗 [Jira Ticket M2-8720](https://mindlogger.atlassian.net/browse/M2-8720)

### 📸 Screenshots

| CleanShot 2025-02-20 at 12 47 48 | CleanShot 2025-02-20 at 12 47 04 |

### 🪤 Peer Testing

- Create an activity with 2 items, Date + other item types.
- Add conditional logic using the Date as condition
- Select a date
- **Date should be the same** after closing the Date Picker
- Save applet
- **Date should remain the same** as previously selected
- Refreshing the page **should show the same date**

### DIFF
--- a/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
+++ b/src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
@@ -788,7 +788,7 @@ const formatTime = (hours: number, minutes: number) => {
   return `${formattedHours}:${formattedMinutes}`;
 };
 
-const formatDate = (dateValue: string) => {
+const parseDate = (dateValue: string) => {
   const date = dateValue ? parseISO(dateValue) : undefined;
 
   return date;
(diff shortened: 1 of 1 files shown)


src/modules/Builder/pages/BuilderApplet/BuilderApplet.utils.tsx
```diff
@@ -789,7 +789,7 @@ const formatTime = (hours: number, minutes: number) => {
 };
 
 const formatDate = (dateValue: string) => {
```
farmerpaul: Nit: I question the naming of this function – it's doing precisely the opposite of "formatting" a date, isn't it? Taking a date that's in a particular expected format, and parsing it into an object…

```suggestion
const parseDate = (dateValue: string) => {
```

farmerpaul (approved):
Looks good, just wanted to flag one dumb nit in the existing code. 🙊

PR #2041, opened 2025-02-26
fix: Support possible empty nickname for Team Members (M2-8757)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8757](https://mindlogger.atlassian.net/browse/M2-8757)

It's possible to add a Team Member that lacks a nickname field by adding them as a Full Account first. Normally, the BE auto-populates the subject record's `nickname` with `[first name] [last name]`, but this procedure bypasses that.

This is a problem because the nickname field is the only thing shown in participant dropdowns (used by Take Now and Assign Activity) for Team Members, and if that field is empty (along with missing the Team tag, which is a related bug that will be taken care of in a separate BE ticket), those rows can end up being empty.

### 📸 Screenshots

(FYI, the participants listed are indeed team members despite their "Full" account names.)

| CleanShot 2025-02-26 at 17 55 17@2x | CleanShot 2025-02-26 at 17 55 43@2x |

### 🪤 Peer Testing

1.  Create a new applet.
2.  From **Applet > Participants**, click **Add Participant** > **Full Account** > fill in only required fields using a known accessible email address. Leave Nickname and Tag blank, and submit the form.
3.  Accept the invitation from the email so that the Full Account is now added to the applet.
4.  From **Applet > Team**, click **Add Team Member** > fill in required fields using the **same email address** as in step 1.
5.  Accept the invitation from the email so that the account has not been converted into a Team Member.
6.  On the **Applet > Participants** tab, observe the **Nickname** column for the row associated with the new Team Member account.
    -   **Expected outcome:** Nickname should be `[first name] [last name]`. (Ignore the fact that it's missing the Team tag; this task is not intended to fix that separate issue.)
7.  From **Applet > Activities**, open the **Take Now** modal for any activity, and observe the contents of the dropdowns.
    -   **Expected outcome:** There should be no blank rows in the dropdowns; the new Team Member should be listed as `[first name] [last name]`. (Ignore the fact that it's missing the Team tag; this task is not intended to fix that separate issue.)

### DIFF
--- a/src/modules/Dashboard/components/ParticipantDropdown/ParticipantDropdown.test.tsx
+++ b/src/modules/Dashboard/components/ParticipantDropdown/ParticipantDropdown.test.tsx
@@ -1,8 +1,13 @@
 import { renderWithProviders } from 'shared/utils/renderWithProviders';
 import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
+import {
+  mockedFullParticipant1WithDataAccess,
+  mockedOwnerParticipantWithDataAccess,
+} from 'shared/mock';
 
 import { ParticipantDropdown } from './ParticipantDropdown';
 import { ParticipantDropdownOption } from './ParticipantDropdown.types';
+import { participantToOption } from './ParticipantDropdown.utils';
 
 jest.mock('shared/hooks/useFeatureFlags');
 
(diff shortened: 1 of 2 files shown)


sultanofcardio (approved):
Nice quick fix here @farmerpaul

PR #2048, opened 2025-03-10
fix: Error on Dataviz first load (M2-8779)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8779](https://mindlogger.atlassian.net/browse/M2-8779)

The Dashboard > Applet screen was deleting the Redux-stored applet data after it was unmounted, causing moments during navigation to subscreens when the required encryption data was not available to be able to decrypt answers. This resulted in an error when navigating to Dataviz for the first time, but only when there are subscales and there is reasonable network latency.

It appears that applet data was being removed in the `AppletMultiInformant` screen as a manual approach to invalidating cache, ensuring it's refreshed in case `appletId` changes. However, the better approach is to include `appletId` as a dependency on the fetching callback, which is possible by passing it as an optional dep to the `usePermissions` hook.

Also, updated the `calcScores` function to handle the possibility that the answer data is unavailable to prevent displaying a JS error in the React UI.

### 📸 Screenshots

| Before                       | After                                 |
| <video src="https://github.com/user-attachments/assets/c6cb9b39-9572-4df4-bd13-f86ca5401a93"></video> | <video src="https://github.com/user-attachments/assets/64eca661-ca4c-410d-b129-dcbe21649469"></video> |

### 🪤 Peer Testing

**Preconditions**

1.  Create an activity with an item that has scoring enabled.
2.  Add a subscale calculated from that item (no lookup table needed).

**Steps to Reproduce the Issue**

1.  In the Web App, submit an activity assessment.
2.  Start a new login session in the Admin App.
3.  Open DevTools (command+option+I), navigate to the **Network** tab, choose **Slow 4G** from the throttling dropdown. (**Fast 4G** often works as well, but less consistently.)
    
4.  Navigate to the Applet Overview screen, and click the most recent submission in the list.
    **Expected outcome:** Dataviz should render correctly.

### DIFF
--- a/src/shared/utils/exportData/getSubscales.ts
+++ b/src/shared/utils/exportData/getSubscales.ts
@@ -104,34 +104,36 @@ export const calcScores = <T>(
 
     let value: number | null = null;
 
-    if (typedOptions && 'options' in typedOptions && typedOptions.options.length) {
-      // Single & Multiple Select
-      const scoresObject = typedOptions.options.reduce((acc: ScoresObject, item) => {
-        if (item.value !== undefined && item.score !== undefined) {
-          acc[item.value as keyof ScoresObject] = item.score;
+    if (typedOptions) {
+      if ('options' in typedOptions && typedOptions.options.length) {
+        // Single & Multiple Select
+        const scoresObject = typedOptions.options.reduce((acc: ScoresObject, item) => {
+          if (item.value !== undefined && item.score !== undefined) {
+            acc[item.value as keyof ScoresObject] = item.score;
+          }
+
+          return acc;
+        }, {});
+
+        if (Array.isArray(answer?.value)) {
+          value =
+            answer?.value.reduce((result: null | number, val) => {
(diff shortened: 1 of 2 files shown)


src/shared/utils/exportData/getSubscales.ts
```diff
@@ -104,7 +104,7 @@ export const calcScores = <T>(
 
     let value: number | null = null;
 
-    if ('options' in typedOptions && typedOptions.options.length) {
+    if (typedOptions && 'options' in typedOptions && typedOptions.options.length) {
```
andrevitalb: Wouldn't a general

```ts
if (typedOptions) {
	// ...
}
```

wrapping around these two be a bit cleaner? (Avoiding checking two times for the same value)

andrevitalb (approved):
Nice and clean with the dependency array solution 🤓

sultanofcardio (approved):

PR #2049, opened 2025-03-13
feat: item_type column addition to export file (M2-8276)

andrevitalb:

### 📝 Description

🔗 [Jira Ticket M2-8276](https://mindlogger.atlassian.net/browse/M2-8276)

This PR adds a new column to both data export files (`responses.csv` and `activity_user_journey.csv`) called `item_type` which will be used to, at a quick glance, determine the type of field the answer value belongs to.

This maps the pre-existing `reponseType` value that is returned from the data export endpoint from the BE.

### 🪤 Peer Testing

Currently, there is support for a feature flag (`enableDataExportRenaming`) targeted towards previously implemented data export changes. Support for this new feature was included for both the new implementation and the existing legacy code.

When enabling/disabling this feature flag, the `item_type` column should be present on both versions of both files, with the same values in each case.

The only difference between new & legacy versions of this specific feature is the order/positioning of the column. The positioning for the column would be as such:

| -      | `responses.csv`/`report.csv`                                                    | `activity_user_journey.csv`                                                     |
| ------ | ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| Legacy | <ul><li>`version`</li><li>`item_type`</li><li>`rawScore`</li></ul>              | <ul><li>`version`</li><li>`item_type`</li></ul>                                 |
| New    | <ul><li>`item_response_status`</li><li>`item_type`</li><li>`rawScore`</li></ul> | <ul><li>`item_response`</li><li>`item_type`</li><li>`press_next_time`</li></ul> | 

This feature flag can be locally enabled/disabled for testing within `src/shared/hooks/useFeatureFlags.ts`


farmerpaul (approved):
Great job! Thanks for accommodating my request in the BE PR and using the existing `ResponseType` enum, it's so much simpler!

It also occurred to me that the data team is already using the values from this enum verbatim in Mixpanel event tracking, so I think it only makes sense to continue to stay consistent in all facets of the app.

One tip: if the requirements for the [Jira task](https://mindlogger.atlassian.net/browse/M2-8276) have changed, be sure to keep the task definition updated. I'm not sure if the attached XLS file is still relevant to the task? Also, it appears the AC for the task still has the templated content in it (that should be removed), and the bullet that was added specifically for the task is not clear that the new column should be included in _both_ exported files. That should to be refined before moving the task to `To Be Tested` to minimize confusion with QA – feel free to make the changes yourself.

mbanting (approved):

PR #2052, opened 2025-03-13
feat: add percentage option to subscale calculation (M2-8433)

ChaconC:

### 📝 Description

Adds Percentage as a subscale calculation option.

🔗 [Jira Ticket M2-8433](https://mindlogger.atlassian.net/browse/M2-8433)

### 🪤 Peer Testing

Test in conjunction with [Backend PR #1783](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1783)

> [!NOTE]
> Applet needs to have Report Server configured 

In the admin app:
- Go to Applet Settings > Activity Settings
- Add scored items (Slider, Single Select, Multi Select)
  - Enable Score using the item settings menu under "Scores and Alerts"
  - Add a Slider item and enable score
  - Add a Multi Select item and enable score
- Add Subscale configuration selecting "Percentage" in Subscale scoring
  CleanShot 2025-03-13 at 12 02 55
- Add a Lookup table to the Subscale
- Enable "Calculate total score" and select "Percentage of Item Scores"
- Add a "Score" under "Scores & Reports" and use the generated subscale
  - Score calculation type should appear as "Percentage"
  - Make sure to enable "Generate Report"
- Save applet 

**Expected behavior**: Admin is able to save/submit the applet with Percentage subscale score calculation type

Complete the activity using the web or mobile app and generate at least 1 submission.

Admin:
- Navigate to DataViz for the participant used for submission
- DataViz should reflect the score as a percentage
  - **Note**: if the value doesn't match any subscale value/ranges, the raw score will be displayed (70 in my screenshot)
  CleanShot 2025-03-13 at 12 12 03
- Export Data 
  - The exported values should match the DataViz results
- Download Latest Report
  - The PDF values should match the DataViz results 

|Score out of range|Score in range|
|-|-|
|CleanShot 2025-03-13 at 12 14 33|CleanShot 2025-03-13 at 12 17 25|
|CleanShot 2025-03-13 at 12 12 03|CleanShot 2025-03-13 at 12 17 33|

### Lookup table:
```
score,rawScore,age,sex,optionalText,severity
10,0,15,M,score-0,
20,1~20,15,M,score-1,Minimal
30,21~30,15,M,score-2,Mild
40,31~40,15,M,score-3,Moderate
50,41~50,15,M,score-4,Severe
```

### Item Configuration Example:
|Item|Screenshot|Notes|
|-|-|-|
|**Single Select**|CleanShot 2025-03-13 at 12 24 34|Score is calculated by using the maximum possible score option (40)|
|**Multi Select**|CleanShot 2025-03-13 at 12 22 03|Score is calculated by adding all possible options (60)|
|**Slider**|CleanShot 2025-03-13 at 12 25 19|Score is calculated by using the maximum possible score option (13)|

### Subscale configuration example:
|Item|Screenshot|
|-|-|
|**Multiple Items**|CleanShot 2025-03-13 at 12 26 23|
|**Single Item**|CleanShot 2025-03-13 at 12 26 54|

Example scoring: Score is calculated by adding the maximum score for each item (40 + 60), then adding each answer 
 and calculating the percentage in `getSubscaleScore`. If there is a lookup table attached to the subscale, it will use this value to find a matching range or return then `rawScore`.

### ✏️ Notes

The Report Server already [handles Percentage calculation](https://github.com/ChildMindInstitute/mindlogger-report-server/blob/e8cf505ebe4663fd8f60d1e759b97653450eb2b3/src/models/activity.ts#L75), so the score calculation logic is implemented following that as guidance.


mbanting (approved):
Thanks @ChaconC. code looks good. Tested successfully with BE PR. Only thing I'm not sure about is what's expected for range of scores for lookup scores with percentage scoring type.

farmerpaul (changes requested):
No comments on your code changes, all of it looks good. I did notice the **Possible Range of Scores** calculation wasn't taking into account the lookup table data when "Percentage" was picked, the way the "Sum" and "Average" calculations currently do.

I whipped together a fix for that, but please confirm. In `ScoreContent.utils.ts`:

```ts
    case CalculationType.Percentage: {
      const minScore = isNaN(lookupTableMinScore)
        ? totalMinScore
        : Math.min(totalMinScore, lookupTableMinScore);

      const maxScore = isNaN(lookupTableMaxScore)
        ? totalMaxScore
        : Math.max(totalMaxScore, lookupTableMaxScore);

      return { minScore: maxScore ? (minScore / maxScore) * 100 : 0, maxScore: 100 };
    }
```

farmerpaul (commented):
I'm getting some discrepancies between the Report Server and Dataviz when I configure an activity with a subscale that includes a nested subscale score as ones of its items. Here's the subscale configuration:




In the Report Server, both subscales report the same value, which in this case makes sense because the two subscales are calculated from the same activity items, and the only difference is that the one called "Percent 2" also contains the nested "Regular Percent" subscale score as one of its items (keeping the percentage the same):

CleanShot 2025-03-18 at 15 24 18@2x

But Dataviz shows something different. The most recent submission has a score of **843.75** for the "Percent 2" subscale (and an even more extreme value of 1350 for a previous submission as you can see there):

CleanShot 2025-03-18 at 15 24 02@2x

Seems there's some disagreement between how the Report Server interprets nested subscales vs. Dataviz.

ChaconC:
Thanks @farmerpaul, I was able to replicate your exact results. However, there's 2 issues here:
1. DataViz wasn't taking into account the result from nested subscales, I have added a fix for this case. This should apply to the data you already collected.

|Before|After|
|-|-|
|CleanShot 2025-03-18 at 20 18 12|CleanShot 2025-03-18 at 20 18 44|

3. Report Server doesn't take nested subscales into account, there's an open ticket for this issue. M2-8124 (thanks for finding it!). To test this I created a new activity using Sum operations and the same values you shared via slack. DataViz now shows 8 and 16 correctly, but the PDF shows 8 in both cases.

|DataViz|Report PDF|
|-|-|
|CleanShot 2025-03-18 at 20 24 53|CleanShot 2025-03-18 at 20 25 04|

### Subscale configuration
CleanShot 2025-03-18 at 20 28 00

src/shared/utils/exportData/getSubscales.ts
```diff
@@ -88,6 +88,7 @@ export const calcScores = <T>(
 
       if (typeof calculatedNestedSubscale?.score === 'number') {
         result[item.name] = calculatedNestedSubscale;
+        maxScore += calculatedNestedSubscale.score;
```
farmerpaul: I don't see how the solution is quite this simple, and my own findings confirm my suspicion. Let me demonstrate with a simple example (with **no lookup tables** set up for now):

- The activity has 3 items whose sum of scores can range from 0 to 8.
- "Sum" is a subscale that calculates the sum of those three items
    - Maximum "Sum" score is 8
- "Nested Sum" is a subscale that includes the sum of the three items AND the "Sum" nested subscale
    - Maximum "Nested Sum" score is 16
- "Percentage" is a subscale that calculates the percentage based on the three items
    - For example, when "Sum" is 6: this should be 6 ÷ 8 = 75%
- "Nested Percentage" is a subscale that calculates the percentage based on the three items AND the "Percentage":
    - For example, when "Sum" is 6: the percentage score of the three items becomes 75% out of a possible maximum of 100%. Meanwhile, the "Percentage" _nested subscale_ is also 75% out of a possible maximum of 100%. Thus we arrive at 75% + 75% (150%) out of a possible maximum of 100% + 100% (200%) which brings us back to 75% for the "Nested Percentage".

However, this is what I see in Dataviz. The first submission is when I selected responses to make a "Sum" score of 6, and the second submission is when I selected responses to make a "Sum" score of 8. The second submission is fine, but in the first submission the "Nested Percentage" score is showing a score of 97.59% rather than 75%:

CleanShot 2025-03-19 at 09 36 51

Here are some issues that I see:

- `maxScore += calculatedNestedSubscale.score` is just adding the score that's _empirically derived from current responses_ to the maximum score – not the _theoretical maximum possible_ nested subscale score. I think that's part of the issue. I believe `maxScore` may need to be returned as a separate property from this recursive function.
- Care should be taken when mixing different kinds of nested subscale calculations (sums, averages, percentages), depending on the approach. For instance, to me it makes little  sense to add 75% (percentage-based nested score) + 6 (sum-based nested score) = 81 (kind of meaningless?). Or maybe that's fine? I'm not sure. Also not sure if this is even happening or would happen, but just something that occurred to me that I thought worth mentioning.
- I guess the task says that we don't have to pay too much attention to lookup tables for percentage, but I think we still have to handle it in some way. The simplest way to evaluate the maximum possible score when a lookup table is involved would be to get the maximum value of the `score` column, and using that as the `maxScore`. (This doesn't take into account the possibility that some rows may not be applicable to the gender/age of the subject, but unclear if doing so is even desirable.)

farmerpaul (commented):
Thanks for looking into it, but I'm still getting weird results with nested subscales. I posted my thoughts; it just seems a little non-trivial to fix it – and why I thought maybe it's worth splitting this off into a separate task.

src/shared/utils/exportData/getSubscales.ts
ChaconC: After talking with Jody about this, I've made a change to prevent adding nested subscales to Percentage type calculations:

https://github.com/user-attachments/assets/1c499f04-24da-4308-b34f-b74465449dbf

In the proposed use case it doesn't make sense to include nested subscales for the reasons you highlighted, and now the UI properly reflects this. 

CleanShot 2025-03-19 at 23 23 58

farmerpaul: Fantastic. Nice work!

farmerpaul (approved):
🙌🏻

PR #2053, opened 2025-03-18
feat: Dataviz filtering changes (M2-8221)

andrevitalb:

### 📝 Description

🔗 [Jira Ticket M2-8221](https://mindlogger.atlassian.net/browse/M2-8221)

This PR propagates the changes introduced in [this BE PR](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1726), which added an extra filter to the answers `/dates` endpoint (`activityOrFlowId`) as part of a fix to only show correct dates in the data visualization date selector calendar.

This is now automatically added to the dates call and is determined by the route parameter (activity or activity flow ID).

While adding these changes, tests targetted towards data visualization that were still using legacy routes (multi-informant agnostic), which are now deprecated, were spotted. The work to update these tests as a whole will be done in a separate ticket. Within this PR's scope, some minor tweaking was done to the existing tests to address these new changes and the base groundwork for the test updates was laid out, as well as the tests for this specific implementation.

### ✏️ Notes

Fear not! This isn't a huge PR. Most of the line changes are from moving/renaming mock variables & files (`/src/shared/mock.ts` -> `/src/shared/mock/index.ts` mainly)


src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -235,7 +238,7 @@ export const RespondentDataReview = () => {
     const lastActivityCompletedDate = lastActivityCompleted && new Date(lastActivityCompleted);
 
     if (selectedDateParam) {
-      const selectedDate = parseDateToMidnightUTC(selectedDateParam);
+      const selectedDate = parseDateToMidnightLocal(selectedDateParam);
```
farmerpaul: By changing these three instances of `parseDateToMidnightUTC` to `parseDateToMidnightLocal`, the `parseDateToMidnightUTC` utility is now only being used in unit tests and not in the actual app.

And its uses in `CalendarEvents.utils.test.ts` seem to be wrong, because all the other date parsing instances in that `getEventEndDateTime` unit test do not actually use the `Z` UTC signifier. That is, that unit test should probably also be using `parseDateToMidnightLocal`.

Which means maybe `parseDateToMidnightUTC` shouldn't even exist in the app, if it's not in fact the right function to use when parsing dates when interacting with the API. Given all this, I think we should probably just remove it (and the `parseDateToMidnightUTC` unit tests) – what do you think?

src/shared/mock/index.ts
```diff
@@ -17,8 +17,16 @@ import {
...
+import { page } from 'resources';
 
+import {
+  defaultMultiSelectionConfig,
+  defaultSingleSelectionConfig,
+  defaultSliderConfig,
+  defaultTextConfig,
+  defaultTimeConfig,
+} from '../../modules/Builder/features/ActivityItems/ItemConfiguration/OptionalItemsAndSettings/OptionalItemsAndSettings.const';
+import { LookupTableDataItem } from '../../modules/Builder/features/ActivitySettings/SubscalesConfiguration/LookupTable';
```
farmerpaul: Nit: sometimes relative imports sneak in where they're not supposed to, it's hard to get VS Code to behave consistently!

```suggestion
} from 'modules/Builder/features/ActivityItems/ItemConfiguration/OptionalItemsAndSettings/OptionalItemsAndSettings.const';
import { LookupTableDataItem } from 'modules/Builder/features/ActivitySettings/SubscalesConfiguration/LookupTable';
```

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -144,11 +144,13 @@ export const RespondentDataReview = () => {
     const datesResult = data?.result;
     if (!datesResult) return;
 
-    const submitDates = datesResult.dates.map(parseDateToMidnightUTC);
+    const submitDates = datesResult.dates.map(parseDateToMidnightLocal);
     setResponseDates(submitDates);
   });
 
   const handleGetSubmitDates = (date: Date) => {
+    const activityOrFlowId = activityId || activityFlowId;
```
farmerpaul: Nit: the suggested change below is not going to affect behaviour of this screen, for now (or maybe ever). However, checking for the presence of a flow ID before activity ID is better aligned with the hypothetical possibility that you could be in a context where you're looking at an activity flow _and_ an activity within that flow. Our routes don't support that yet and we haven't built that out in the context of Dataviz – but one day, someone might. So it's safer to check if there's a flow present before assuming we're only looking at an activity.  (This order of checks is done more commonly in the backend and respondent apps, but I figured it wouldn't hurt to do the same here.)

```suggestion
    const activityOrFlowId = activityFlowId || activityId;
```

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -235,7 +238,7 @@ export const RespondentDataReview = () => {
     const lastActivityCompletedDate = lastActivityCompleted && new Date(lastActivityCompleted);
 
     if (selectedDateParam) {
-      const selectedDate = parseDateToMidnightUTC(selectedDateParam);
+      const selectedDate = parseDateToMidnightLocal(selectedDateParam);
       if (isValid(selectedDate)) {
         handleSetInitialDate(selectedDate);
```
farmerpaul: I'm placing this comment here, but it actually applies to lines 249-253 below:

```tsx
    if (lastActivityCompletedDate) {
      handleSetInitialDate(lastActivityCompletedDate);
      return;
    }
```

This is still a relic from before we redesigned Dataviz to be activity/flow-specific. What this code appears to do is choose a default date for the datepicker based on the most recent submission of _any activity/flow – not necessarily the contextual one being viewed_.

Which means that if your applet has multiple activities/flows – and you've only submitted answers for one of those activities/flows most recently, but not the others on that submission date – then the page shows no responses by default when you get here from one of those other activities/flows. Furthermore, the datepicker is highlighting an unselectable date by default, which isn't right.

I think there's gotta be a fairly minor refactor you can do here that gives preference to the result returned by `getAppletSubmitDateList`, picking the most recent one from that array. Do you mind looking into this for this task? It would be a really nice bit of polish. 🙏🏻

farmerpaul (commented):
This is great work @andrevitalb. It's _such_ a relief to use Dataviz and actually not be so completely mystified as to which dates in the datepicker actually have responses, feels so much better!

I mentioned a couple tiny nits, and pointed out a now-unused utility function. I also made a suggestion for a last bit of polish that I hope you'll agree is in scope, but let me know if you feel otherwise.

Again, great work! 💪🏻

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
andrevitalb: Yep, totally agree with you on that one 🤓. Adding changes.

src/shared/mock/index.ts
andrevitalb: Updating 🤓

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
andrevitalb: I had actually a couple refactors, including this one that might've been pertinent to include in this PR. The two main reasons I hadn't included them were:

1. They required extra alignments on the product side to verify desired/ideal behavior for some parts.
2. I wasn't able to finish those before my OOO and ideally didn't want to block this feature from being released due to those changes holding it off.

I'll go ahead and this small one and align with the team for the rest of minor QOL improvements like this.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
andrevitalb: Updated 🤓

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
```diff
@@ -246,6 +246,12 @@ export const RespondentDataReview = () => {
       }
     }
 
+    if (responseDates?.length) {
+      handleSetInitialDate(responseDates[responseDates.length - 1]);
+
+      return;
+    }
+
```
farmerpaul, line 266: This doesn't seem to do anything when I try to test this out. I don't think `responseDates` is actually initialized (it stays `undefined`) until after this effect has already been run, because `handleGetSubmitDates` only gets called in the `handleSetInitialDate` function, which is what this effect is designed to call. That's why I thought this change would need a little bit more of a refactor.

Or did you find it was working for you? Maybe there's just something quirky about my local data…

farmerpaul (changes requested):
Thanks for taking care of these changes @andrevitalb, in your off time no less! The last change it's doing the trick for me, curious what your own tests showed. If it's too much of a refactor feel free to split it off into another ticket.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
andrevitalb: For some reason I was inconsistently getting the expected behavior with those changes.

Did a more complete overhaul on how the initial date is set. Behavior now works as it should. Updating tests accordingly now.

farmerpaul: Hey @andrevitalb, I'm testing your latest version, and now I'm not seeing any default date being selected, for any subject or any activity/flow I open the Responses tab for:

https://github.com/user-attachments/assets/32426576-90e0-4ff1-90c2-f1307829830b

I understand your code change, it seems like it should work. I'm not totally comfortable with the additional `// eslint-disable-next-line react-hooks/exhaustive-deps` line (we're trying to avoid those whenever possible), but I get that the excluded dependencies are probably okay to exclude? It might not hurt to align the dependencies with the linter, which I realize will entail creating a `useCallback`. Though I'm not sure if that will fix what I'm seeing.

I wonder if there's some race condition that's causing it to work fine for you, but not for me?

farmerpaul: Oh I do see the latest responses appearing for an activity that has **recent responses**. In that example I recorded above, I was testing with an older applet whose most recent responses are at least a couple months old (no more recent than January). So something about when the last responses were submitted seems to be playing a role in this bug. Otherwise, it seems to be working!

andrevitalb: ~Hmm, I know what's happening, and has to do with how the `selectedDateParam` is set.~

~When the `selectedDateParam` is set to a date that is not included in the `responseDates` array, the initial date is set to the current one, hence, not showing any results by default.~

Let me have a look at this one.

andrevitalb: So, that was causing a different issue. 

This issue is still present. Will keep trying to find what's causing this one

andrevitalb: @farmerpaul I've found the cause for this behavior. The thing is, when we're fetching the submit dates for any given activity flow/activity, the `fromDate` and `toDate` params are required for said endpoint (`/dates`).

There's currently no way for us to be able to fetch the dates without those parameters or to keep track of the current activity flow/activity latest submission date. We're only able to do this for the whole applet-wide submissions (using `lastActivityCompleted`).

With latest changes, one of two cases can happen:

1. If the applet has submissions (`lastActivityCompleted` is set), the answer dates for the same month as `lastActivityCompleted` are fetched from the API.
	1. If at least one date is returned from that API call (activity flow/activity in context has answers that month), that one is set as initial date.
	2. If no answers for context activity flow/activity are present for that month, no dates are rendered.
2. If applet doesn't have submissions (`lastActivityCompleted` is not set), answer dates for current month are set as parameters to fetch dates, activities and flows, which will just end up rendering as no answers.

This behavior isn't ideal as there's currently no way of manually updating the selected date range to fetch answers from (similar to what happens for summary), hence, reducing the options to properly fetch answers for a given activity to manually accessing a desired submission, and only then, be able to see the submissions for that same month. The only other option to get results from a month outside of the previously mentioned ones is manually setting the `selectedDate` query param in the URL, which of course, is not ideal

farmerpaul: Great investigation, interesting how it's working. Seeing as how long this is taking, my feeling now is what you've done is currently acceptable for now, and we should create a separate ticket for the remaining errant behaviour, which can be prioritized accordingly. It's already a whole lot better than what it was! I can create that separate ticket if you like.

farmerpaul (approved):
Thanks for pushing through with all my nitpicks and extra requests, and for giving such detailed explanations – so helpful. I think this is good to ~~merge~~ QA. 💪🏻

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.tsx
farmerpaul: Created [M2-8967](https://mindlogger.atlassian.net/browse/M2-8967) to eventually fix this bug.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.test.tsx
```diff
@@ -373,224 +131,514 @@ describe('RespondentDataReview', () => {
...
+
+    //     const feedbackMenuClose = screen.getByTestId(`${dataTestid}-feedback-menu-close`);
+    //     expect(feedbackMenuClose).toBeInTheDocument();
+
+    //     await userEvent.click(feedbackMenuClose);
+
+    //     expect(feedbackMenu).toHaveStyle({ width: 0 });
+    //   },
+    //   JEST_TEST_TIMEOUT,
+    // );
```
farmerpaul: Was this commented code meant to be included?

farmerpaul (approved):
Seems to be working well now, thanks for making all those changes. I just pointed out a big block of commented unit test code.

src/modules/Dashboard/features/RespondentData/RespondentDataReview/RespondentDataReview.test.tsx
andrevitalb: Updated the tests and splitted this one out 🤓

PR #2055, opened 2025-03-20
fix: Clear previous applet data when creating new applet (M2-8920)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-8920](https://mindlogger.atlassian.net/browse/M2-8920)

This PR intercepts the stored applet data on the applet builder screen, and returns an empty object if the user navigated there to create a new applet.

### 📸 Screenshots

**Before**

https://github.com/user-attachments/assets/92fa2508-e55d-496c-838c-d9da6bba9eec

**After**

https://github.com/user-attachments/assets/03525378-4633-460c-9c0e-05aff585cdcc

### 🪤 Peer Testing

1. Log in to the admin panel
2. Without clicking anything else, create an applet and save it
3. Navigate to the applet overview screen
4. Navigate back to the dashboard
5. Click Add applet, then click New
6. Confirm that all fields are empty
7. Fill in the required minimum fields and save the applet
8. Confirm that the applet can be saved

### ✏️ Notes

1. This seems to have existed prior to #2048. That PR just made it more obvious. For example, you could previously produce this using the following steps:
    1. Navigate to the Dashboard
    2. Click on an applet to view the Applet overview screen
    3. Click on a submission to go to the Dataviz summary screen
    4. Navigate back to the Dashboard
    5. Click add applet
    6. Click new
    
    This PR should fix it everywhere

### DIFF
--- a/src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
+++ b/src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
@@ -31,7 +31,11 @@ export const BuilderApplet = () => {
   const hiddenHeader = !!params.activityId || !!params.activityFlowId;
   const dispatch = useAppDispatch();
   const { appletId } = useParams();
+  const removeAppletData = useRemoveAppletData();
   const isNewApplet = useCheckIfNewApplet();
+  if (isNewApplet) {
+    removeAppletData();
+  }
   const redirectedFromBuilder = forbiddenState.useData()?.redirectedFromBuilder ?? {};
   const { result: appletData } = applet.useAppletData() ?? {};
   const { getAppletWithItems } = applet.thunk;
@@ -44,7 +48,6 @@ export const BuilderApplet = () => {
     appletResponseType === 'applet/getAppletWithItems' &&
     !isNewApplet;
   const { ownerId } = workspaces.useData() || {};
-  const removeAppletData = useRemoveAppletData();
   const [isAppletInitialized, setAppletInitialized] = useState(false);
   const { data: dataFromLibrary, isFromLibrary } = location.state ?? {};
   const hasLibraryData = isFromLibrary && !!dataFromLibrary;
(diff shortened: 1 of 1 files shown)


src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
```diff
@@ -31,7 +31,11 @@ export const BuilderApplet = () => {
   const hiddenHeader = !!params.activityId || !!params.activityFlowId;
   const dispatch = useAppDispatch();
   const { appletId } = useParams();
+  const removeAppletData = useRemoveAppletData();
   const isNewApplet = useCheckIfNewApplet();
+  if (isNewApplet) {
+    removeAppletData();
```
farmerpaul: I'm always wary of changing reactive state in the render thread, especially as this doesn't need to re-execute on every render (it only should execute on first render or if `isNewApplet` changes). Did you try this as an effect, or run into an issue with that approach?

farmerpaul (commented):
Thanks for taking this on @sultanofcardio. Just one question…

src/modules/Builder/pages/BuilderApplet/BuilderApplet.tsx
sultanofcardio: You're right, and I thought about that, but I didn't try it because I didn't want to have cases where the data shows for a split second before re-rendering. But I forgot about useLayoutEffect which could probably achieve the same thing so I'll update this to use that if it works

farmerpaul: I had the same thought about a flicker issue – was wondering if that could be remediated by a loading state, or intercepting what's assigned to `appletData` in case `isNewApplet` is true…

sultanofcardio: Intercepting the applet data seems like the simplest option. Thanks for the suggestion

farmerpaul (approved):
Looks good, thanks @sultanofcardio!

PR #2057, opened 2025-03-26
Release/2025.03.5 [main]

mbanting:
Merge `release/2025.03.5` to `main`


farmerpaul:
@mbanting I think if we're going to deploy #2048, we should also include #2055, the fix for [M2-8920](https://mindlogger.atlassian.net/browse/M2-8920), a bug that became reliably reproducible after #2048 was merged.

[M2-8920]: https://mindlogger.atlassian.net/browse/M2-8920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

sultanofcardio:
@farmerpaul M2-8920 was tested only in the amplify environment, but not in UAT. So I don't think we should ship it to production until QA gets a chance to test it there unless we clear it with them first

rcmerlo (approved):

mbanting:
@farmerpaul @sultanofcardio #2055 was included and tested on UAT as part of `2025.03.5-rc2`
Thanks for raising this.

PR #2058, opened 2025-03-27
feat: Add EHR banners (M2-8823)

farmerpaul:

### 📝 Description

🔗 [Jira Ticket M2-8823](https://mindlogger.atlassian.net/browse/M2-8823)

Add EHR "available" and "active" Banners to the main Dashboard screen, above the tab bar, depending on feature flag and workspace permissions. Banner is shown on the main Applets, Respondents, and Managers screens, but not shown when navigating to a deeper level.

A few small app-wide improvements were made to support this change:
- Strongly type feature flags similarly as how it's been done in a couple of other other repos, to support the new string union-typed `enable-ehr-health-data` flag.
- Track whether feature flags have been loaded in `workspaces` global state, required to prevent flickering of banners during login & when switching workspaces. (Previously, `LeftBar` was tracking this locally, but never actually using it for anything. It also tracked a local copy of returned feature flags for no apparent reason, so I removed that as well.)
- Extend the MUI theme palette with an `infoAlt` entry, for purple banner state.
- Add `ThemeProvider` to our `render*WithProviders` utilities to keep Jest happy when using palette extensions.
- Add a `reason` argument to the `onClose` callback of `Banner` components to distinguish why a banner was closed (in this case, whether it was manually dismissed).

### 📸 Screenshots

CleanShot 2025-03-27 at 08 24 30@2x

CleanShot 2025-03-27 at 08 24 13@2x

### 🪤 Peer Testing

> [!NOTE]
> In DEV, the `enable-ehr-health-data` feature flag is set to `available` by default. It can be set to `active` on a per-workspace basis. For testing, I recommend asking @mbanting to set the `enable-ehr-health-data` feature flag to `active` for a specific user ID in your local dev environment to be able to test both the `active` and `available` states.

**Preconditions:**
- Invite the user whose workspace you requested to have the `enable-ehr-health-data` feature flag set to `active` (call it Workspace A, owned by User A) to be a **Manager** in an applet in another workspace (Workspace B, owned by User B), for which the `enable-ehr-health-data` feature flag is set to `available` (the default).
- Invite User B to be a **Coordinator** in an applet owned by User A.

**Testing as User A:**
1. **Log in** to the admin app as **User A**.
    **Expected outcome:** [EHR "active" banner](https://www.figma.com/design/HYigOkc6fI5sxOwhmJxlVI/EHR-2025?node-id=4294-17797&t=jwZGuHfoaeOMhl6d-4) is visible.
2. Navigate to the **Respondents** or **Managers** tab.
    **Expected outcome:** EHR "active" banner remains visible.
3. **Click an applet** to navigate to its Overview screen, or any of its other sub-screens
    **Expected outcome:** EHR "active" banner is hidden.
4. Switch to **Workspace B**.
    **Expected outcome:** [EHR "available" banner](https://www.figma.com/design/HYigOkc6fI5sxOwhmJxlVI/EHR-2025?node-id=4294-17648&t=jwZGuHfoaeOMhl6d-4) is shown.
5. Return to **Workspace A**.
    **Expected outcome:** EHR "active" banner is visible again.
6. **Dismiss** the "active" banner.
    **Expected outcome:** EHR "active" banner is hidden with a smooth exit transition.
5. Return to **Workspace B**.
    **Expected outcome:** EHR "available" banner is visible again.
6. **Dismiss** the "available" banner.
    **Expected outcome:** EHR "available" banner is hidden with a smooth exit transition.
5. Return to **Workspace A**.
    **Expected outcome:** EHR "active" banner remains hidden.
5. Return to **Workspace B**.
    **Expected outcome:** EHR "available" banner remains hidden.
5. **Log out** and **log back in**.
    **Expected outcome:** EHR "active" banner remains hidden.
5. Return to **Workspace B**.
    **Expected outcome:** EHR "available" banner remains hidden.

**Testing as User B:**
1. **Log in** to the admin app as **User B**.
    **Expected outcome:** EHR "available" banner is visible.
2. Navigate to **Workspace A**.
    **Expected outcome:** No banner is shown (those with Coordinator role are not permitted to access EHR features).
3. **Log out** and **log back in**.
    **Expected outcome:** EHR "available" banner is visible.
6. **Dismiss** the "available" banner.
    **Expected outcome:** EHR "available" banner is hidden with a smooth exit transition.
3. **Log out** and **log back in**.
    **Expected outcome:** EHR "available" banner remains hidden.


src/shared/components/Banners/EHRBanners/EHRBanners.const.ts
```diff
@@ -0,0 +1,3 @@
+// TODO: Replace with actual links before merging PR
```
farmerpaul: Note to @farmerpaul to update this file before QA/merging.

src/modules/Dashboard/pages/Main/Main.tsx
mbanting, line 1: In general this works. It seems to deviate from the existing banner patterns that abstract the banner behaviour from the parent component. Eg

- Hooks like `useSessionBanner` and `useModalBanners`
- The global `Banners` component handling banner state changes on the redux store

Do you see an opportunity to leverage any of these existing solutions or if this behaviour is different enough, create something similar that could handle banners like the EHR banner? If there's a quick win here, great, if not we can consider it outside the scope of this ticket and keep it in mind for the future.

src/shared/utils/storage.ts
```diff
@@ -5,6 +5,8 @@ export const enum LocalStorageKeys {
   LibraryPreparedData = 'libraryPreparedData',
   IsFromLibrary = 'isFromLibrary',
   LibraryUrl = 'libraryPath',
+  EHRBannerAvailableDismissed = 'EHRBannerAvailableDismissed',
+  EHRBannerActiveDismissed = 'EHRBannerActiveDismissed',
```
mbanting: Looks like this was intended to be used in `Applets.tsx` but is no longer referenced and has been replaced by a unique key generated by `getDismissedKey`. Can this be removed?

mbanting (commented):
As always great job @farmerpaul .
Left a couple of comments but going ahead and approving already. Thank you!

src/modules/Dashboard/pages/Main/Main.tsx
farmerpaul: Good callout. I'm going to try to make this a little more generic and separate it into its own component.

farmerpaul: I opted to just move it to a separate component, `EHRBanners`, but left it with very EHR-specific logic, and this is why:

It would be a bit of work to make it more generic, while continuing to account for some very finicky flickering issues I had encountered during implementation. My solution to one of the flickering issues ended up relying on having a single state variable `ehrBanner` that could either show one banner, or the other, or neither. Also, the collapse animation has to run only when the banner was manually closed, but not when switching between workspaces/banner states.

There's probably a way to make it more generic (like `useModalBanners`) while accounting for these nuances, but I don't feel the effort is justified at this stage. At least for now it's separated from the screen component.

mbanting (approved):
Thanks for incorporating the suggestions!

src/shared/components/Banners/Banner/Banner.tsx
```diff
@@ -24,7 +24,7 @@ export const Banner = ({
   useEffect(() => {
     if (!duration || !onClose || isHovering || !isWindowFocused) return;
 
-    const timeoutId = setTimeout(onClose, duration);
+    const timeoutId = setTimeout(() => onClose?.('timeout'), duration);
```
andrevitalb, line 27: Nice and simple. Love this!

src/shared/components/Banners/EHRBanners/EHRBanners.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const handleOnExited = () => {
+    if (isCollapsing) {
+      setEhrBanner(null);
+      setIsCollapsing(false);
+    }
+  };
+
+  useEffect(() => {
+    // We have to prevent the banner from showing when feature flags or roles are still being
+    // loaded to prevent flickering of underisable banner states.
```
andrevitalb: Small typo 🤓
```suggestion
    // loaded to prevent flickering of undesirable banner states.
```

src/shared/components/Banners/EHRBanners/EHRBanners.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+      renderWithProviders(<EHRBanners />, {
+        preloadedState: {
+          ...getPreloadedState(Roles.Manager),
+          ...mockAuthState,
+        },
+      });
+
+      expect(screen.getByTestId('ehr-banner-available')).toBeInTheDocument();
+      expect(screen.queryByTestId('ehr-banner-active')).not.toBeInTheDocument();
```
andrevitalb: Nit: There were some recent issues with test ids being out of sync. Would it be wise/cleaner to keep these two declared as `const` at the top of the tests? We can minimize places these need to be replaced in case we need to in the future

andrevitalb (approved):
Great work as always! Spotted one typo and left a small nit comment, but overall, amazing work 🤓

Loving those changes to feature flag handling.

src/shared/components/Banners/EHRBanners/EHRBanners.test.tsx
farmerpaul: Good callout, made that change!

PR #2069, opened 2025-04-16
Release/2025.04.2 [main]

mbanting:
Merge `release/2025.04.2` to `main`


farmerpaul (approved):
Good to go! (Assuming the feature flag is set to `unavailable` for all workspaces on PROD 😅)

src/shared/components/Banners/EHRBanners/EHRBanners.const.ts
```diff
@@ -0,0 +1,4 @@
+// TODO: Replace with actual links when microsite is ready
+// https://mindlogger.atlassian.net/browse/M2-8823
```
andrevitalb, line 2: While working on M2-8927 I spotted a small issue with this one. I believe it might be better to point this comment to [M2-8965](https://mindlogger.atlassian.net/browse/M2-8965)

andrevitalb (approved):
All looking good to 🚢. Left a small note on a comment

src/shared/components/Banners/EHRBanners/EHRBanners.const.ts
farmerpaul: Good catch - but probably doesn't make a difference for production

mbanting: @andrevitalb thanks! If you don't mind, please update in one of your active feature branches so that it finds it way back to dev via that merge instead of this release branch. Thank you!

PR #2072, opened 2025-05-01
feat: Export Modal Everywhere (M2-8584) (M2-9089) (M2-8938)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-8584](https://mindlogger.atlassian.net/browse/M2-8584)
🔗 [Jira Ticket M2-9089](https://mindlogger.atlassian.net/browse/M2-9089)
🔗 [Jira Ticket M2-8938](https://mindlogger.atlassian.net/browse/M2-8938)

This PR introduces the export modal in all the places where it is possible to perform a data export. This includes the following places:
1. Dashboard respondents tab (`/dashboard/respondents`) using the applet context menu
2. Applet page top-level export button (e.g. on the overview page at `/dashboard/{applet_id}/overview`)
3. Applet page activities tab (`/dashboard/{applet_id}/activities`) using the activity/flow context menu
4. Applet page participants tab (`/dashboard/{applet_id}/participants`) using the participant table row context menu
5. Participant detail page assignments tab (`/dashboard/{applet_id}/participants/{subject_id}`) using the activity/flow/participant context menu
6. DataViz Summary tab (`/dashboard/{applet_id}/participants/{subject_id}/activities/{activity_id}/summary`) export button
7. Applet Settings (`/dashboard/{applet_id}/settings`) export data button
8. Anywhere else I might have forgotten

It also introduces the concept of supplementary exported files, optionally controlled by feature flags.

Finally, this PR includes two bug fixes:
- Passing the missing filters on the dataviz screen, which now allows the applet to properly filter exported data by the activity/flow and participant that we are viewing
- Properly determining the "from date" when exporting using the "All time" setting, as in some cases this was not being properly determined and resulting in missing data

#### More on component reorganisation

Previously the export process was controlled by two components:
- `DataExportPopup` - Does not contain a modal with date options, and is used in the places where the modal is not shown
- `ExportDataSetting` - Contains the actual modal with the date options (all time, last 24h, etc), and is used in the places where that modal is shown (usually within the context of a single applet). Has `DataExportPopup` as a nested component.

The `DataExportPopup` component is the component that is mostly used, which is why data exports do not show a modal in most places. It contains another modal with an optional table of applets from which to select (in case multiple applets were applicable, such as on the Dashboard respondents tab).

In order to support this behaviour in places where the modal is not shown, I initially thought it would be a simple case of switching components. However, the date picker on the date options modal (shown by `ExportDataSettings`) needs the applet to be selected in order to populate the "from date" field with the applet creation date. Thus it became necessary to create a new component (the `ChooseAppletPopup` component) to handle picking the applet before the `ExportDataSetting` component is shown. This abstract the behaviour that used to be handled by `DataExportPopup`, so now all that component does is collect the applet password (if necessary) and performs the data export.

As a result of this change, the `ExportDataSetting` component is expected to be used everywhere data exports are done, and `DataExportPopup` should not be used directly.

#### Adding more supplementary files

Adding more checkboxes is now a 3 step process:

Update the list of supplementary files in `ExportDataSetting.types.ts`

https://github.com/ChildMindInstitute/mindlogger-admin/blob/a3e0bb5bf5deb1b89c2c191af86096656f668f74/src/shared/features/AppletSettings/ExportDataSetting/ExportDataSetting.types.ts#L14-L20

Add a corresponding entry in the translation files with the copy that shows up in the modal

https://github.com/ChildMindInstitute/mindlogger-admin/blob/b9653f9c9bce42800959555a5ab33aaada11aa80/src/resources/app-en.json#L468-L471

image

Handle the supplementary files value in the `DataExportPopup` component. For example:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/a3e0bb5bf5deb1b89c2c191af86096656f668f74/src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old.tsx#L189-L196

### 📸 Screenshots

Here's a before and after example from the dashboard "Respondents" tab

#### Before

https://github.com/user-attachments/assets/f61ab366-2593-4e7b-bf82-1977a8a3638a

#### After

https://github.com/user-attachments/assets/660d42bb-39e7-4aa6-b074-76958233b0bf

#### Modal Presentation

Use the list above to perform a data export and confirm the following in each place:
- The modal with the various date options is shown in each place
- The choose dates options works
- The applet name is populated in the date options modal

#### Supplementary files

While the date options modal is shown, confirm that the supplementary files checkboxes are shown when the `enableEmaExtraFiles` feature flags is true.
Confirm that the activity flow history checkbox is not shown in the modal from the context menu for activities

https://github.com/user-attachments/assets/e0f77a67-c9d6-4c74-afad-36e3db4669e6

#### Dataviz Export

- Perform a data export on the Dataviz summary screen in an applet with lots of responses across activities/flows
- Confirm that the data exported is specific to the activity/flow and participant being viewed

#### Incorrect from date

You will need two applets with different creation dates for this test

- Log out, then back into the admin panel
- Select the first applet, then open the network tab of dev tools
- Do an "All time" data export
- In developer tools network tab, inspect the ‘data’ request and confirm the presence of a `fromDate` property in the payload. Note its value
- Go back to the dashboard and select the second applet
- Do an "All time" data export of this second applet
- In developer tools network tab, inspect the second ‘data’ request and confirm that the `fromDate` property in the payload is not the same as before.

### ✏️ Notes

There are a couple broken tests, but I've opened this to get some eyes on it in the meantime


src/modules/Dashboard/features/Respondents/Popups/ChooseAppletPopup/ChooseAppletPopup.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { ChosenAppletData } from 'modules/Dashboard/features/index';
+
+/**
+ * A modal component to allow the user to choose from one of several applets a respondent has
+ * access to
+ * @param tableRows
+ * @param popupVisible
+ * @param handleClose
+ * @param dataTestid
+ * @constructor
```
farmerpaul: These props seem to be out of sync; I don't usually bother with this level of JSdoc detail anymore since they're so subject to change, and IDEs infer them correctly already. Maybe a relic of autocomplete?

src/modules/Dashboard/features/Respondents/Popups/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from './ScheduleSetupPopup';
 export * from './ViewParticipantPopup';
 export * from './EditRespondentPopup';
-export * from './DataExportPopup';
+export * from '../../../../../shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup';
```
farmerpaul: :sus-eyes:

src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportWorkersManager/DataExportWorkersManager.ts
```diff
@@ -7,12 +7,11 @@ import { ItemResponseType } from 'shared/consts';
...
-import DecryptionWorker from '../DataExportWorker/DataExportWorker.worker';
-import { IdleWorker } from '../DataExportPopup.types';
-import { NUM_WORKERS } from '../DataExportWorker/DataExportWorker.const';
-import { WorkerOnMessageEvent } from '../DataExportWorker/DataExportWorker.types';
-import { getExportDataSuffix } from '../DataExportPopup.utils';
+import DecryptionWorker from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportWorker/DataExportWorker.worker';
+import { IdleWorker } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup.types';
+import { NUM_WORKERS } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportWorker/DataExportWorker.const';
+import { WorkerOnMessageEvent } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportWorker/DataExportWorker.types';
+import { getExportDataSuffix } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup.utils';
```
farmerpaul, line 14: There's a lot of these long import lines happening between different modules now… Seems a tad cumbersome, but I can live with it so long as it's intentional.

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
```diff
@@ -386,8 +386,14 @@ export const useAssignmentsTab = ({
...
-      {showExportData && (
-        <DataExportPopup
+      {appletData && (
+        <ExportDataSetting
+          isExportSettingsOpen={showExportData}
+          onExportSettingsClose={() => setShowExportData(false)}
+          onDataExportPopupClose={() => {
+            setSelectedActivityOrFlow(undefined);
+            setSelectedTargetSubjectId(undefined);
+          }}
```
farmerpaul, line 396: I'm curious why in your PR notes you make an exception that on the PDP, we shouldn't expect the "Include activity flow history" checkbox to be hidden for activities. Can't we still do that here with the below change? Maybe I'm missing something.

```suggestion
          }}
          supportedSupplementaryFiles={
            selectedActivityOrFlow?.isFlow ? undefined : ['scheduleHistory']
          }
```

farmerpaul: Similarly, there doesn't seem to be any distinction between whether the currently viewed data is for an activity or for a flow in DataViz – `RespondentDataHeader.tsx`:

https://github.com/ChildMindInstitute/mindlogger-admin/blob/f4e44f36a064dea4725e2ebb846625a4f5d31635/src/modules/Dashboard/features/RespondentData/RespondentDataHeader/RespondentDataHeader.tsx#L184-L187

Could we not add this line there?
```tsx
              supportedSupplementaryFiles={activityFlowId ? undefined : ['scheduleHistory']}
```

I suppose then the `filters` prop should be here, too, right? Unless, again, I'm missing something!

farmerpaul (commented):
Great work as always @sultanofcardio. Just had a couple comments.

src/modules/Dashboard/features/Respondents/Popups/ChooseAppletPopup/ChooseAppletPopup.tsx
sultanofcardio: Indeed. These are usually for param specific docs, but I should remove them here

src/modules/Dashboard/features/Respondents/Popups/index.ts
sultanofcardio: 🤦‍♂️😅

src/modules/Dashboard/features/Participant/Assignments/AssignmentsTab/AssignmentsTab.hooks.tsx
sultanofcardio: Yes, this is correct. Based on how I defined the type, the values in this array can't be dynamic. For some reason I then ended up conflating this with dynamically passing the array, but that's totally fine because the values are still static.

Thanks for pointing this out, this was a pretty silly miss 😅

About the missing filters on Dataviz, I guess we could add those in. I think that may fix [M2-9089](https://mindlogger.atlassian.net/browse/M2-9089) as well

farmerpaul (approved):
Peer testing for all 3 tickets passing, nice work!

src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup.tsx
```diff
@@ -0,0 +1,21 @@
+import { useFeatureFlags } from 'shared/hooks/useFeatureFlags';
+import { DataExportPopup as DataExportPopupOld } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old';
+import { DataExportPopup as DataExportPopupNew } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_new';
+import { DataExportPopupProps } from 'shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup.types';
```
andrevitalb: I think for these imports, we could do `'./...'`, since these routes are a tad too long IMO

src/shared/types/generics.ts
```diff
@@ -11,3 +11,26 @@ export type AtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyo
...
+ * const validTuple2: MyTuple = ['foo'];
+ * const validTuple3: MyTuple = ['bar'];
+ *
+ * // Not accepted
+ * const invalidTuple: MyTuple = ['foo', 'foo']; // Error: Duplicate element 'foo'
+ * const invalidTuple2: MyTuple = ['foo', 'bar', 'baz']; // Error: 'baz' is not part of the union
+ * const invalidTuple3: MyTuple = []; // Error: Empty tuple
+ * ```
+ */
+export type UniqueTuple<
```
andrevitalb, line 33: I like this one 🤓

andrevitalb (approved):
Great work @sultanofcardio. Peer tested the 3 tickets & everything is working great.

src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup.tsx
sultanofcardio: That's cool with me. I almost never look at these tbh

PR #2077, opened 2025-05-07
fix: EMA Fast Follows (M2-9162) (M2-9158)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-9162](https://mindlogger.atlassian.net/browse/M2-9162)
🔗 [Jira Ticket M2-9158](https://mindlogger.atlassian.net/browse/M2-9158)

This PR fixes a couple EMA issues. 

For M2-9162, the schedule history file currently does not accurately reflect the behavior of the MindLogger application where participants revert to the default schedule after their individual schedule has been deleted. Specifically this ticket describes the scenario where there are multiple versions of the individual schedule before it is deleted.

For M2-9158, this PR fixes the integration of the new RTK Query method of fetching respondents. The request was not sending an explicit *limit* query parameter, and it seems the API has a default limit of 10. This in turn caused the method that calculates the total number of pages to fail because it assumes that the limit is set to 250 (The ceiling of 10/250 = 1 page). Therefore all applets would export schedule history for no more than 10 respondents. My fix for this makes the assumption of 250 explicit to the caller by providing a parameter for it

### 📸 Screenshots

N/A

#### M2-9162

> [!WARNING]
> Not yet dev testable, unless you already have an applet that fit the criteria below

Testing this scenario completely requires the passage of several days. This is because we need to perform the following steps all on different days:
- Create applet, invite participant, create individual schedule
- Update individual schedule
- Delete individual schedule
- Perform applet export

You can look at the code and the test and convince yourself that it works, or you can wait for me to update the applet seed script to support this scenario.

#### M2-9158

1. Create an applet with many participants (more than 20)
2. Perform an applet data export
3. Inspect the schedule history file and confirm the presence of every participant at least once

### ✏️ Notes

N/A


src/shared/utils/exportData/exporters/ScheduleHistoryExporter.ts
```diff
@@ -130,7 +131,7 @@ export class ScheduleHistoryExporter extends DataExporter<
       const data = await store
         .dispatch(
           apiDashboardSlice.endpoints.getWorkspaceRespondents.initiate({
-            params: { ownerId: this.ownerId, appletId, page },
+            params: { ownerId: this.ownerId, appletId, page, limit: MAX_LIMIT },
```
mbanting:
```suggestion
            // TODO: Paginate to get all data when endpoint supports pagination
            params: { ownerId: this.ownerId, appletId, page, limit: MAX_LIMIT },
```

mbanting (approved):
Code looks good. Just one comment to add a TODO.
I didn't test myself but unit tests seems to cover the scenario.

src/shared/utils/exportData/exporters/ScheduleHistoryExporter.ts
sultanofcardio: @mbanting I was mistaken here. The API endpoint does return paging info, but my code wasn't properly handling it. `requestAllPagesConcurrently` implicit depends on the `limit` query parameter being set to 250 (the value of `DEFAULT_ROWS_PER_PAGE`) in order to calculate the number of pages. I've since made a couple changes:
- Make the dependency on the limit query parameter explicit by adding a parameter for it to `requestAllPagesConcurrently`
- Set the limit query parameter to `DEFAULT_ROWS_PER_PAGE` for the respondents endpoint

These changes should remove the need for this comment

PR #2081, opened 2025-05-14
fix: EMA Fast Follows 2 (M2-9106) (M2-9108)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-9106](https://mindlogger.atlassian.net/browse/M2-9106)
🔗 [Jira Ticket M2-9108](https://mindlogger.atlassian.net/browse/M2-9108)

This PR fixes two EMA bugs, as described below

#### M2-9106: In the export the new version is linked for the participant not only the individual but also the default schedule

In the specific scenario where the following steps occur on the same day:
- An applet is created (version 1.1.0)
- A participant is added
- The participant is given an individual schedule
- The applet version is increased (version 2.0.0)
- The schedule history is exported for the participant

The default schedule event for the activity in the applet should not be present in the schedule history export file for the participant with the individual schedule on the new applet version. Therefore the schedule history file should contain the following entries for this participant:
- Default schedule event (applet version 1.1.0)
- Individual schedule event (applet version 1.1.0)
- Individual schedule event (applet version 2.0.0)

This is because the default schedule did not apply when the applet version was increased (as the participant already had an individual schedule at that time), and so it should not be present in the schedule history export file.

#### M2-9108: The whole schedule_history associated with the removed participant is also removed

This issue is mostly fixed with https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1848 (and requires it to be tested), but this PR adds a few things:

- Specifies to the API that soft-deleted participants should be included in the workspace respondents endpoint
- Prevents schedule history from being generated past the date when the participant is removed

### 📸 Screenshots

N/A

#### M2-9106

- Create an applet
- Add a participant
- Create an individual schedule for the participant
- Rename the applet (to increase the version)
- Export the applet data
- Open the schedule_history export file
- Confirm that the default schedule event does not show up for the increased applet version (you can confirm which one is the default schedule event by looking in the `events` table filtered by your `applet_id`)

#### M2-9108

- Create an applet
- Add a participant
- Remove the participant without removing their data
- Export the applet data
- Open the schedule_history export file
- Confirm that the data for the removed participant is still present

### ✏️ Notes

N/A


src/shared/utils/exportData/exporters/ScheduleHistoryExporter.ts
```diff
@@ -514,7 +531,96 @@ export class ScheduleHistoryExporter extends DataExporter<
...
+
+          for (let i = sortedByLinkDate.length - 1; i >= 0; i--) {
+            const schedule = sortedByLinkDate[i];
+            if (schedule.userId && indexOfLastIndividualScheduleLinkedToday === -1) {
+              indexOfLastIndividualScheduleLinkedToday = i;
+            } else if (!schedule.userId && indexOfLastIndividualScheduleLinkedToday !== -1) {
+              indexOfLastDefaultScheduleLinkedTodayBeforeIndividual = i;
+              break;
+            }
+          }
```
andrevitalb, line 583: This can be done a bit cleaner with a `reduce`: 
```suggestion
          const { lastIndividualIndex, lastDefaultBeforeIndividualIndex } = sortedByLinkDate.reduce(
            (acc, schedule, index) => {
              if (schedule.userId) {
                return { ...acc, lastIndividualIndex: index };
              } else if (
                !schedule.userId &&
                acc.lastIndividualIndex !== -1 &&
                acc.lastDefaultBeforeIndividualIndex === -1
              ) {
                return { ...acc, lastDefaultBeforeIndividualIndex: index };
              }

              return acc;
            },
            { lastIndividualIndex: -1, lastDefaultBeforeIndividualIndex: -1 },
          )
```

andrevitalb (approved):
Code is looks great @sultanofcardio. Peer testing now 🤓, but pre-aproving

src/shared/utils/exportData/exporters/ScheduleHistoryExporter.ts
sultanofcardio: Thanks for the suggestion @andrevitalb, but I don't think there's a reliable way to make this work with a forward reduce. The loop method is meant to run backwards so it's easy to find the last index of an individual event, and the index of the default event that came before it can only be set after its index has been found. With the reduce method, the index of the last individual schedule event will eventually settle on the correct one, but the default schedule event index probably won't. This is because it needs to know ahead of time what the index is for the individual schedule event, but that index keeps changing.

I think if I was to rewrite it as a reduce, I'd do it like this

```typescript
const { lastIndividualIndex, lastDefaultBeforeIndividualIndex } = [...sortedByLinkDate]
            .reverse()
            .reduce(
              (acc, schedule, index) => {
                if (schedule.userId && acc.lastIndividualIndex === -1) {
                  return { ...acc, lastIndividualIndex: sortedByLinkDate.length - 1 - index };
                } else if (
                  !schedule.userId &&
                  acc.lastIndividualIndex !== -1 &&
                  acc.lastDefaultBeforeIndividualIndex === -1
                ) {
                  return {
                    ...acc,
                    lastDefaultBeforeIndividualIndex: sortedByLinkDate.length - 1 - index,
                  };
                }

                return acc;
              },
              { lastIndividualIndex: -1, lastDefaultBeforeIndividualIndex: -1 },
            );
```

But I'm not sure we gain much readability from this, and we lose the ability to terminate the loop early when the two indexes have been found

andrevitalb:
Peer testing successful 🤓

src/shared/utils/exportData/exporters/ScheduleHistoryExporter.ts
andrevitalb: You're right. Seems like the way to go for now functionality-wise

PR #2084, opened 2025-05-23
feat: Rebrand banner (M2-9069)

andrevitalb:

### 📝 Description

🔗 [Jira Ticket M2-9069](https://mindlogger.atlassian.net/browse/M2-9069)

This PR introduces a brand new Rebrand banner to the application as a first notice for the upcoming branding changes.

### 📸 Screenshots

image

### 🪤 Peer Testing

Testing is pretty simple.

The banner should be visible in two places:
- Auth-related screens (`/auth`) (no authentication needed)
- Top level dashboard pages (`/dashboard/applets`, `/dashboard/managers`, `/dashboard/respondents`) (authentication needed)

The dismissed state is tracked separately for auth & non-auth routes.

Dismissing the banner in a non-auth route should keep the state persistent throughout the use of these routes, without login/logout sessions disrupting it.

On the other hand, dismissing in an auth route should keep the state persistent for the logged in user throughout login sessions.


src/resources/app-en.json
```diff
@@ -1262,6 +1262,7 @@
   "quote": "Quote",
   "random": "Random",
   "rangeOfScores": "Possible Range of Scores",
+  "rebrandBanner": "<0>We are rebranding!</0> <1>Design updates are on the way — same great app, fresh new look. Curious?</1>",
```
farmerpaul: The Jira task's AC states doesn't include "Curious?" in the rebrand banner, it just ends at "fresh new look."

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+const DISPLAY_ROUTES = [
+  '/auth',
+  '/dashboard/applets',
+  '/dashboard/managers',
+  '/dashboard/respondents',
+];
+
+export const RebrandBanner = (props: BannerProps) => {
+  const { ownerId } = workspaces.useData() ?? {};
```
farmerpaul: I don't think there's we need take into account the workspace for this banner. The main reason the current workspace was important for the EHR banners was because there was workspace-specific permissions to use the EHR feature. In this case, we only care whether the user has seen that there are are going to be brand updates. Which workspace they happen to be in shouldn't affect the flag that they've seen the banner.

But I'm open to hearing another argument for this – lemme know if you feel otherwise!

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+            severity={undefined}
+            data-testid="rebrand-banner"
+            onClose={(reason) => {
+              if (reason === 'manual') {
+                handleDismiss();
+              }
+            }}
+            icon={<StyledImg src={curiousIcon} />}
+            sx={{
+              backgroundColor: '#000',
```
farmerpaul: Nit: Super minor shade difference in Figma…
```suggestion
              backgroundColor: '#0B0907',
```

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+              color: variables.palette.surface,
+              '& .MuiAlert-message': {
+                maxWidth: theme.spacing(90),
+              },
+              '& .MuiAlert-icon': {
+                opacity: 1,
```
farmerpaul: 🦅 Good eye. This icon in fact should never have been `opacity: 0.9`, which just happens to be the default styling from Material UI. Could you remove this override here, and define it in `theme.tsx` instead? 🙏🏻 

https://github.com/ChildMindInstitute/mindlogger-admin/blob/388ed5c9e5910397e9e388972885bb368340578d/src/shared/styles/theme.tsx#L781-L783

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+              },
+              '& .MuiButton-root': {
+                color: variables.palette.surface,
+              },
+            }}
+            {...props}
+          >
+            <Trans i18nKey="rebrandBannerNoLink">
```
farmerpaul: The translation key is different, maybe you meant to update it here, or in the language files?
```suggestion
            <Trans i18nKey="rebrandBanner">
```

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+              if (reason === 'manual') {
+                handleDismiss();
+              }
+            }}
+            icon={<StyledImg src={curiousIcon} />}
+            sx={{
+              backgroundColor: '#000',
+              color: variables.palette.surface,
+              '& .MuiAlert-message': {
+                maxWidth: theme.spacing(90),
```
farmerpaul: I'm guessing this was added to keep the content all on one line. This works for English, but doesn't quite do the same for the French version (with the current text). Maybe we should just remove `maxWidth` altogether?
```suggestion
                maxWidth: 'none',
```

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+      (location.pathname.startsWith(route) && route.includes('/auth')),
+  );
+
+  const handleDismiss = () => {
+    // For auth screen or when user is not logged in yet
+    if (!userId) {
+      localStorage.setItem(GLOBAL_DISMISSED_KEY, 'true');
+    } else {
+      // Set user-level dismissal
+      localStorage.setItem(getDismissedKey(userId), 'true');
```
farmerpaul, line 62: It's tricky to to perfectly align with this AC:

> - If the user dismisses the banner, its dismissed state should be remembered for the duration of the session and across login sessions

One way would be to be a little more aggressive with setting dismissal state, that is, we dismiss the global auth screen banner even when the user is logged in. Else it's a bit annoying when the banner reappears after you've already dismissed it once while being logged in. It seems like the app has "forgotten" your choice.

What do you think?

```suggestion
    // Always dismiss global auth screen banner to prevent banner from reappearing
    // after having already dismissed it during an active user session.
    localStorage.setItem(GLOBAL_DISMISSED_KEY, 'true');

    if (userId) {
      // Set user-level dismissal
      localStorage.setItem(getDismissedKey(userId), 'true');
```

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+      setIsRebrandBannerActive(!userDismissed);
+    }
+  }, [userId, ownerId]);
+
+  const isDisplayRoute = DISPLAY_ROUTES.some(
+    (route) =>
+      location.pathname === route ||
+      (location.pathname.startsWith(route) && route.includes('/auth')),
+  );
```
farmerpaul: Nit: I think something like this would make the logic a little easier to follow, but also NBD. 😉

```tsx

const DISPLAY_ROUTES = [
  /^\/auth(?:\/|$)/,
  /^\/dashboard\/(applets|managers|respondents)(?:\/|$)/,
];

…


  const isDisplayRoute = DISPLAY_ROUTES.some((route) => route.test(location.pathname));
```

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
```diff
@@ -0,0 +1,141 @@
...
+      if (ownerId) {
+        localStorage.setItem(getDismissedKey(userId, ownerId), 'true');
+      }
+    }
+
+    setIsCollapsing(true);
+  };
+
+  // Don't render if not on a display route or if no user info is available on non-auth routes
+  const shouldRender = isDisplayRoute && (location.pathname.startsWith('/auth') || userId);
```
farmerpaul, line 69: Doesn't `isDisplayRoute` already look after the `/auth` path checks, or do we have to check it twice?

farmerpaul (commented):
Great work @andrevitalb! 🎉 Just had a few minor notes, and would like your feedback on a couple things. 🙏🏻

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
farmerpaul: On second thought, I'm not so sure my idea is the best. 🙃 The more I think about it the more it feels a bit arbitrary… and the AC doesn't clarify or say to do this explicitly. Maybe just leave it the way you have it, treating the "global" login screen key acting like its own "user" independent of other users.

src/resources/app-en.json
andrevitalb: Will align on this on the correct source of truth for this, since the Figma assets do include it.

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Yep, was planning on going with dynamic assignment of translations. This was left over. Nice spot!

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Updating 🤓

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Nice spot! Hadn't noticed looking at it in my monitor. Got a get a more color-accurate one 😅

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Well, this was actually a hard-coded value in the Figma file. But you're right, it doesn't fully wrap in French. I believe it's safe to go with `'none'` in this case to keep it as consistent as possible.

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Hmm, yeah, I think I went a tad overkill with this. Removing it.

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Yep, looks cleaner. Thanks 🤓

src/shared/components/Banners/RebrandBanner/RebrandBanner.tsx
andrevitalb: Yes it does, but this check is intended for something else. This should be interpreted as:
- On auth routes: show the banner regardless of user login status
- On other display routes (dashboard): only show if user is logged in (userId exists)

src/resources/app-en.json
farmerpaul: "Curious?" should not be a part of it for now – this part of the AC was updated by Katy and overrides what is in Figma currently. Katy created 3 follow-up tasks to included the "Curious? Click to learn more." part that includes the link. (It didn't make sense to include the word "Curious?" until there's a link that users can click to satisfy their curiosity.)

andrevitalb: Yep, updated. I also updated the follow-up tickets to make note of the "Curious?" text

farmerpaul (approved):

PR #2098, opened 2025-06-25
fix: Error state on Item Card (M2-9441)

andrevitalb:

### 📝 Description

🔗 [Jira Ticket M2-9441](https://mindlogger.atlassian.net/browse/M2-9441)

This PR fixes an issue where after clicking "Save & Publish" on the applet editor screen, editing items, if an item had errors, the corresponding card wouldn't reflect this in its styling

### 🪤 Peer Testing

1. Login to the admin panel.
2. Start editing any applet.
3. Edit the applet's items.
4. Perform any update that will result in error (e.g. Add an empty item).
5. Switch to a valid item.
6. Click Save & Publish.
7. **Expected:** A modal specifying the error should show up & the faulty item's card on the left bar should be marked as with error (background color should be `error_container` (`error[90]`))

### DIFF
--- a/src/modules/Builder/features/ActivityItems/LeftBar/Item/Item.styles.ts
+++ b/src/modules/Builder/features/ActivityItems/LeftBar/Item/Item.styles.ts
@@ -37,8 +37,8 @@ export const StyledItem = styled(StyledFlexTopCenter, shouldForwardProp)`
         stroke: ${variables.palette.on_surface_variant};
       }
 
-      ${isActive && `background-color: ${variables.palette.secondary_container};`}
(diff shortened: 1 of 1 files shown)


farmerpaul (approved):
I don't quite understand why this fixes the issue, but it does.

andrevitalb:
@farmerpaul Had do to some digging to figure out why. Turns out in styled-components, each CSS property declaration in the template string needs to be terminated with a semicolon, but also each line of the template literal expression needs to be terminated with a semicolon as part of the JavaScript syntax.

When moving the semicolon inside, it becomes part of the CSS string, but the overall template literal expression doesn't have a terminating semicolon, breaking the styled-component.

PR #2167, opened 2025-11-19
perf: Replace JSZip with fflate for export speedup (M2-9993)

sricharan-varanasi:
### 📝 Description

Export operations (CSV, media files, logs) were taking 60+ seconds after Vite migration due to JSZip's poor polyfill performance. Replaced JSZip with fflate for speedup.

🔗 [Jira Ticket M2-9993](https://mindlogger.atlassian.net/browse/M2-9993)

**Root Cause:** JSZip's `generateAsync()` relies on Node.js streams that are slower when transformed by Vite's esbuild polyfills (stream-browserify).

**Solution:** fflate uses synchronous operations with zero Node.js dependencies, designed for modern bundlers.

Changes include:

- Added fflate ^0.8.2 (jszip kept for safe rollback)
- Replaced JSZip with fflate in `exportCsvZip.ts`, `exportZip.ts`, `logger.ts`
- Export times: 68s → 0.15s (trails), 68s → 2.8s (drawing with 828 files)

### 🪤 Peer Testing

**Test export functionality:**

1. Navigate to Dashboard and select any applet with response data
2. Click **Export** button for responses (drawing, trails, stability tracker, flanker)

   **Expected outcome:** Exports complete in under a minute (previously hung for 60+ seconds)

3. Export activity with media files (images/SVGs)

   **Expected outcome:** Media zip downloads quickly without hanging

### ✏️ Notes

- JSZip dependency kept in package.json for safe rollback after production verification
- Can be removed in future PR once confirmed stable
- No compression used (level: 0) for maximum speed - file sizes acceptable for browser downloads

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
         "dompurify": "^3.3.0",
         "dotenv": "^16.5.0",
         "downloadjs": "^1.4.7",
+        "fflate": "^0.8.2",
         "file-saver": "^2.0.5",
         "history": "^5.3.0",
         "i18next": "^23.5.1",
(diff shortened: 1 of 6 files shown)


aweiland (approved):
Testing locally against the offending applet and it performed as expected if not better.

PR #2169, opened 2025-11-24
fix(M2-9714): Correct last 24 hours download time range

maxto024:
### Problem

  Users reported that newly completed activities were not appearing in exported data files when downloading responses from the admin panel when selecting "Last 24 hours" as the date range.

 ###  Root Cause Analysis

  The investigation revealed a timezone handling issue specifically for the "Last 24 hours" option:
  - Backend Limitation: The backend is not timezone-aware and expects UTC timestamps
  - Frontend Issue: "Last 24 hours" option was sending local time using DateFormats.shortISO format (yyyy-MM-dd'T'HH:mm:ss)
  - Result: Recent activities were being filtered out incorrectly due to timezone mismatch between frontend (local time) and backend (expecting UTC)

  Solution Implemented

  Created UTC conversion specifically for "Last 24 hours" option
  - Added formatDateAsUTC helper function that converts local dates to UTC while maintaining the same format
  - Applied UTC conversion ONLY to the "Last 24 hours" date range
  - Other date ranges (Last week, Last month, Custom) remain unchanged
  - Maintains DateFormats.shortISO format (yyyy-MM-dd'T'HH:mm:ss) for backend compatibility

  Implementation details:
 ```
 // Helper function to convert to UTC (workaround for non-timezone-aware backend)
  const formatDateAsUTC = (date: Date): string => {
    const utcDate = new Date(
      date.getUTCFullYear(),
      date.getUTCMonth(),
      date.getUTCDate(),
      date.getUTCHours(),
      date.getUTCMinutes(),
      date.getUTCSeconds()
    );
    return format(utcDate, DateFormats.shortISO);
  };
```

  ### Testing Considerations

  - Test "Last 24 hours" exports across different timezones
  - Verify activities from exactly 24 hours ago are included
  - Confirm activities created near midnight are properly included
  - Validate other date range options still work as before

  ### Impact

  This targeted fix ensures activities from the last 24 hours are properly included in exports by:
  - Converting "Last 24 hours" timestamps to UTC to match backend expectations
  - Maintaining existing behavior for all other date range options
  - Properly handling recent activities regardless of user timezone

  Note: This is a workaround. The proper long-term solution would be for the backend to handle timezone-aware timestamps.

[M2-9714]: https://mindlogger.atlassian.net/browse/M2-9714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old.tsx
+++ b/src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old.tsx
@@ -1,3 +1,4 @@
+import { format } from 'date-fns';
 import get from 'lodash.get';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
@@ -10,6 +11,7 @@ import { ChosenAppletData } from 'modules/Dashboard/features/Respondents/Respond
 import { useDecryptedActivityData } from 'modules/Dashboard/hooks';
 import { Modal } from 'shared/components/Modal';
 import { EnterAppletPassword } from 'shared/components/Password';
+import { DateFormats } from 'shared/consts';
 import {
   ExportDataExported,
   ExportDataFormValues,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old.tsx [hunk 1]
```diff
@@ -118,16 +118,16 @@ export const DataExportPopup = ({
...
-        let fromDate = format(formFromDate, DateFormats.shortISO);
-        let toDate = format(formToDate, DateFormats.shortISO);
+        let fromDate = format(formFromDate, DateFormats.shortISOWithTimezone);
+        let toDate = format(formToDate, DateFormats.shortISOWithTimezone);
 
         // Update the time for last 24 hours submissions
         if (dateType === ExportDateType.Last24h) {
           const currentTime = new Date();
-          const oneDayAgo = new Date(currentTime);
-          oneDayAgo.setHours(currentTime.getHours() - 24);
```
jodybrookover, line 128: This code around one day ago calculation was working previously and I think it's more clear as it was.

jodybrookover (dismissed):

src/shared/features/AppletSettings/ExportDataSetting/Popups/DataExportPopup/DataExportPopup_old.tsx
maxto024: reverted

aweiland (approved):

PR #2173, opened 2025-12-03
fix: Rerender to trigger Google Translate when dropdown changes (M2-10076)

divbzero:

### 📝 Description

🔗 [Jira Ticket M2-10076](https://mindlogger.atlassian.net/browse/M2-10076)

Google Translate mutates DOM and interferes with React. This resulted in the selected dropdown value not updating in the UI even though the underlying `selectedValue` variable was updating in JavaScript.

Changes include:
- Rerender dropdown value when `selectedValue` changes so Google Translate will re-translate.

### 🪤 Peer Testing

Test steps:
1. Use Google Chrome and enable Google Translate. (On macOS, this could require updating the macOS language to something other than English in macOS Settings.)
2. Edit an Item Flow in an applet.
3. Change the value of a dropdown.

Behavior before fix:
- The old dropdown value remains in the UI.

Behavior after fix:
- The new dropdown value in the UI briefly displays in English (e.g., “is equal to”) before Google Translate re-translates it (e.g., “é igual a”).

### DIFF
--- a/src/shared/components/FormComponents/SelectController/SelectController.tsx
+++ b/src/shared/components/FormComponents/SelectController/SelectController.tsx
@@ -187,11 +187,12 @@ export const SelectController = <T extends FieldValues>({
     selectedValue?: string,
     error?: FieldError,
   ) => {
-      // HACK: Rerender when selectedValue changes to avoid interference from Google Translate (M2-10076)
-      const [forceRerenderKey, setForceRerenderKey] = useState(1);
-      useEffect(() => setForceRerenderKey(-forceRerenderKey), [selectedValue]);
+    // HACK: Rerender when selectedValue changes to avoid interference from Google Translate (M2-10076)
+    const [forceRerenderKey, setForceRerenderKey] = useState(1);
+    useEffect(() => setForceRerenderKey(-forceRerenderKey), [selectedValue]);
 
-      return (<Box sx={{ position: 'relative', width: '100%', ...sx }} className={className || ''}>
+    return (
+      <Box sx={{ position: 'relative', width: '100%', ...sx }} className={className || ''}>
         {placeholder && !selectedValue && (
           <>
             <StyledPlaceholderMask>{placeholder}</StyledPlaceholderMask>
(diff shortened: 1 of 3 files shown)


src/shared/components/FormComponents/SelectController/SelectController.tsx
```diff
@@ -186,62 +186,71 @@ export const SelectController = <T extends FieldValues>({
...
-  ) => (
-    <Box sx={{ position: 'relative', width: '100%', ...sx }} className={className || ''}>
-      {placeholder && !selectedValue && (
-        <>
-          <StyledPlaceholderMask>{placeholder}</StyledPlaceholderMask>
-          <StyledPlaceholder className="controller-placeholder">{placeholder}</StyledPlaceholder>
-        </>
-      )}
+  ) => {
+    /* eslint-disable react-hooks/rules-of-hooks */
```
maxto024: If its not complaining in linting, can we remove this `eslint-disable react-hooks/rules-of-hooks *`

maxto024 (approved):

src/shared/components/FormComponents/SelectController/SelectController.tsx
divbzero: Unfortunately, this indeed is complaining in linting:

    192:53  error  React Hook "useState" is called in function "renderSelect" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"   react-hooks/rules-of-hooks
    193:5   error  React Hook "useEffect" is called in function "renderSelect" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks

sricharan-varanasi (approved):

src/shared/components/FormComponents/SelectController/SelectController.tsx
divbzero: I’ve simplified the fix to avoid the ESLint errors.

PR #2176, opened 2025-12-04
fix: Search Item Type in Google translated language (M2-10095)

divbzero:

### 📝 Description

🔗 [Jira Ticket M2-10095](https://mindlogger.atlassian.net/browse/M2-10095)

We filter **Item Type** options according to the original English values, not the values translated by Google Translate.

Changes include:
- Try fetching translated value from the DOM
- Filter **Item Type** options by translated value if available
- Switch from `.toLowerCase()` to `.toLocaleLowerCase()` to handle non-English lowercase (e.g., normalizing `İ` to `i`)

### 🪤 Peer Testing

Test steps:
1. Use Google Chrome and enable Google Translate. (On macOS, this could require updating the macOS language to something other than English in macOS Settings.)
2. Create a new **Item**.
3. Click the **Item Type** dropdown.
4. Type something to search for item types in the dropdown.

Behavior before fix:
- The item types filter for English search terms, but not for search terms in the translated language.

Behavior after fix:
- The item types filter for both English search terms and search terms in the translated language.


divbzero:
Won’t fix as discussed in [Jira Ticket M2-10095](https://mindlogger.atlassian.net/browse/M2-10095).

PR #2192, opened 2026-01-09
  fix(MFA): route-based navigation and backend-driven error handling [M2-10263, M2-10198]

maxto024:
### Summary
- Implement route-based MFA navigation for proper browser back/forward support
- Remove all client-side MFA sessions and attempt tracking
- Backend API responses are now the single source of truth for session status and attempts
- Separate error states for TOTP and Recovery to prevent error bleeding between forms

Jira Tickets: [M2-10198](https://mindlogger.atlassian.net/browse/M2-10198) | [M2-10263](https://mindlogger.atlassian.net/browse/M2-10263)

### Changes

**Route-based Navigation (M2-10263)**
- New routes: `/auth/verify-mfa`, `/auth/verify-recovery`
- New page components with route guards
- Browser back button works with single click (no blinking)
- Session cleanup on login mount prevents stale access

**Backend-driven Errors (M2-10198)**
- `MFASession` now only contains `token` and `sessionId`
- `attemptsRemaining` comes from backend API error metadata
- `isSessionExpired` set by backend response (no client timer)
- Separate `totpVerification` and `recoveryVerification` Redux states
- Errors cleared when switching forms (UI click or browser back/forward)

[M2-10198]: https://mindlogger.atlassian.net/browse/M2-10198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-10263]: https://mindlogger.atlassian.net/browse/M2-10263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/Auth/features/Login/MFAForm/useMFAVerification.ts
```diff
@@ -93,43 +81,32 @@ export const useMFAVerification = (type: MFAVerificationType) => {
...
+    [dispatch, navigate, type, isSubmitting, isSessionExpired],
   );
 
-  // Clear error when user types
+  // Clear error when user types - dispatches correct action based on type
   const clearError = useCallback(() => {
     // Don't clear session-expired errors
-    if (mfaVerification.isSessionExpired || mfaVerification.displayError === 'mfaSessionExpired') {
+    if (isSessionExpired || verification.displayError === 'mfaSessionExpired') {
       return;
```
sricharan-varanasi, line 95: Quick clarification needed:

Could you help me understand why we need both conditions in this check? 
I see that `isSessionExpired` gets set whenever we have the session expired error,
so I'm wondering if the string comparison is providing additional safety that I'm missing?

If they're always in sync, perhaps we could simplify to just checking the flag?
Just want to make sure I understand the intent here.

sricharan-varanasi (commented):
Looks good overall!

src/modules/Auth/state/Auth.reducer.ts
```diff
@@ -134,94 +104,96 @@ export const extraReducers = (builder: ActionReducerMapBuilder<AuthSchema>): voi
...
-    // CHECK FOR EXPIRY FROM BACKEND FIRST
-    if (isSessionExpiredError(errorMessage)) {
-      applySessionExpired(state, errorMessage);
+    const { translationKey, isSessionExpired, attemptsRemaining } =
+      action.payload as MfaErrorResult;
 
-      return; // No further processing
+    state.recoveryVerification.attemptsRemaining = attemptsRemaining;
+
+    if (isSessionExpired) {
```
sricharan-varanasi, line 192: Since the backend now tracks attempts for both ,we can update recovery codes to show the same warnings

src/modules/Auth/features/Login/MFAForm/useMFAVerification.ts
maxto024: Great question! You're right that they're essentially always in sync, when the backend returns a session-expired error, the reducer sets both:

  1. state.isSessionExpired = true
  2. state.[totp/recovery]`Verification.displayError = 'mfaSessionExpired'`

  The dual check is defensive programming - a belt-and-suspenders approach. In theory, we could simplify to just:

```
  if (isSessionExpired) {
    return;
  }
```

  However, I kept both conditions because:

  1. Historical context: The original code checked mfaVerification.isSessionExpired (which was on the verification object). After splitting into separate totpVerification/recoveryVerification states, I moved isSessionExpired to the root auth state, but kept the string check as a fallback.
  2. Edge case safety: If somehow displayError gets set to 'mfaSessionExpired' through an unexpected code path without isSessionExpired being true, we'd still prevent clearing.

src/modules/Auth/state/Auth.reducer.ts
maxto024: Done! Added the same attempts warning logic for recovery codes:

  Auth.reducer.ts (lines 197-202):
```
  if (attemptsRemaining !== null && attemptsRemaining <= 3) {
    state.recoveryVerification.displayError = `${translationKey}|${attemptsRemaining}`;
  } else {
    state.recoveryVerification.displayError = translationKey;
  }
```

  RecoveryCodeForm.tsx - Added `getErrorMessage()` to parse the key|attempts format (same as MFAForm.tsx).

  Recovery codes now show "Invalid recovery code. X attempts remaining"

src/modules/Auth/features/Login/MFAForm/useMFAVerification.ts
sricharan-varanasi: Thanks for the clarification

src/modules/Auth/pages/MFARecovery/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  // Prevent flash of content when no session
+  if (!mfaSession) {
+    return null;
+  }
+
+  const handleSwitchToTOTP = () => {
+    // Clear recovery error when switching to TOTP
+    dispatch(auth.actions.clearRecoveryError());
+    navigate(-1); // Use history back for clean navigation
```
sricharan-varanasi, line 37: we can use Use `navigate(page.mfaVerify)` instead of '-1' to have the same coding pattern as in MFAVerify/index.tsx

sricharan-varanasi (commented):
Can we add basic tests for as part of another ticket on top of this:
    - Route guard behavior (redirects when no session)
    - Error clearing on mount
    - Navigation handlers

sricharan-varanasi (approved):
LGTM! Just a small thing to address before merging. Thank you

src/modules/Auth/state/Auth.reducer.ts
```diff
@@ -134,94 +104,101 @@ export const extraReducers = (builder: ActionReducerMapBuilder<AuthSchema>): voi
...
   });
 
   builder.addCase(verifyMFATOTP.fulfilled, (state, action) => {
     // Set user data and clear MFA session
     createAuthFulfilledData(state, action.meta.requestId, { user: action.payload?.result.user });
     state.mfaSession = undefined;
-    state.mfaVerification = { status: 'idle' };
+    state.totpVerification = { status: 'idle' };
+    state.recoveryVerification = { status: 'idle' };
+    state.isSessionExpired = false;
```
sricharan-varanasi, line 117: I see the fulfilled handler depends on `createAuthFulfilledData` checking `authentication.status === 'loading'`, but in this MFA flow that status isn’t set prior to fulfillment. I see signIn.fulfilled setting `authentication.status = 'idle'`, and the `verifyMFATOTP.fulfilled` path later relies on the loading state and matching requestId, which won’t align. This mismatch causes `isAuthorized` not to be updated in the fulfilled step.

sricharan-varanasi (changes requested):

src/modules/Auth/pages/MFARecovery/index.tsx
maxto024: this is acting like back to not adding navigation to new routes.

src/modules/Auth/state/Auth.reducer.ts
maxto024: thanks for suggestion. 
here what i did   
 **Fix Applied:**                                                                                                                                                                                                                                       
  1. Added `createAuthPendingData(state.authentication, action.meta.requestId)` to both `verifyMFATOTP.pending `and `verifyMFARecoveryCode.pending `handlers                                                                                               
  2. Added` state.authentication.status = 'idle' `reset in rejected handlers to support retry scenarios (when first TOTP attempt fails, second attempt can properly update the requestId)

maxto024:
> Can we add basic tests for as part of another ticket on top of this: - Route guard behavior (redirects when no session) - Error clearing on mount - Navigation handlerers
we can do that latter seperate pr probably much cleaner

sricharan-varanasi (approved):
Thanks for the fix, LGTM!

PR #2202, opened 2026-03-06
fix: Shorten Unity export file path (M2-10494)

divbzero:

### 📝 Description

🔗 [Jira Ticket M2-10494](https://mindlogger.atlassian.net/browse/M2-10494)

Changes include:

- Shorten folder name in Unity export file path to use answer ID only
- Update the corresponding test


sricharan-varanasi (approved):
Looks good!

PR #2208, opened 2026-04-23
fix: Rename MERIT task performance (M2-10625)

divbzero:
🔗 [Jira Ticket M2-10625](https://mindlogger.atlassian.net/browse/M2-10625)

Changes include:

- Extract `OpenModalButton` as a separate `UnityFileButton` component
- `useMemo` to avoid building `AppletSchema` on every keystroke
- `useMemo` to avoid duplicate LaunchDarkly requests
- `useMemo` and `React.memo` to avoid `LinkedTabs` rerenders
- `useMemo` and `React.memo` to avoid `ToggleItemContainer` rerenders
- `useMemo` and `React.memo` to avoid `UnityFileModal` rerenders
- `useImmediateValidation` for `NameDescription` fields

These changes reduce the time needed to rerender `BuilderApplet`. A remaining optimization we could make is breaking `BuilderApplet` into separate components we don’t rerender the whole thing on each keystroke. This is a bigger refactor that we should tackle only if needed.

### DIFF
--- a/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/AudioPlayer/AudioPlayer.tsx
+++ b/src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/AudioPlayer/AudioPlayer.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { MediaType } from 'modules/Builder/components/MediaUploader';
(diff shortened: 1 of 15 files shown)


divbzero:
@natalia-muzyka noticed that we weren’t triggering validation as we typed for `NameDescription` fields. Fixed with commit ecd052a.

src/modules/Builder/features/ActivityItems/ItemConfiguration/InputTypeItems/AudioPlayer/AudioPlayer.tsx
```diff
@@ -27,14 +27,17 @@ export const AudioPlayer = ({ name }: AudioPlayerProps) => {
     setMedia({ ...media, name: getMediaName(url) });
   }, [url]);
 
+  const headerContentProps = useMemo(() => ({ media }), [media]);
```
adeiji, line 30: Nice architectural decision here!

adeiji (approved):
Everything looks excellent!

PR #2231, opened 2026-06-17
fix: Assign drawer drops search-selected participant (M2-10902)

sricharan-varanasi:
- [ ] OSS packages added to Curious [open source credit page](https://mindlogger.atlassian.net/wiki/spaces/MINDLOGGER1/pages/263127186/Admin+Panel+Applet+Builder+Library)

### 📝 Description

🔗 [Jira Ticket M2-10902](https://mindlogger.atlassian.net/browse/M2-10902)

The Assign drawer loads the participant list once when it opens. If we search for a participant who isn't in that list (e.g. one who accepted their invite after the drawer loaded), selecting them worked, but the **Assign to** field then went blank and showed a false "Add Respondent" warning. The assignment was saving correctly , it was only the display that was wrong.

This happened because the field's displayed value was always rebuilt from that initial list, which didn't include search-only participants.

Changes include:

- Remember participants picked from search in the Assign drawer, and use that list (plus the original one) to resolve what's shown in the dropdowns.
- Apply the same on the review step so the participant shows there too.
- No changes to search itself or the shared participant dropdown hook (avoids re-renders that affect the Take Now modal).

### 🪤 Peer Testing

Preconditions: an activity/flow with Auto-assign off, on the Applet > Participant page.

- Add a new full participant and accept the invite from the email. **Do not refresh the admin page.**
- Switch to **Activities**, click **Assign** → **Assign to**, and note the new participant isn't in the list.
- Search for them, then select them.

    **Expected outcome:** The participant stays shown in the **Assign to** field.

- Pick a user in **Who is activity about**.

    **Expected outcome:** The **Assign to** field still shows the selected participant - no "Add Respondent" warning. Completing the assignment works as normal.

### ✏️ Notes

Scoped to the display issue when searching/selecting a participant. It does not make a just-accepted participant appear in the list automatically - we still search for them (or refresh).

### DIFF
--- a/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
+++ b/src/modules/Dashboard/components/ActivityAssignDrawer/ActivityAssignDrawer.tsx
@@ -1,6 +1,7 @@
 import { useTranslation } from 'react-i18next';
 import { Box, Button, Drawer, Fade, IconButton } from '@mui/material';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import unionBy from 'lodash/unionBy';
 import { Controller, SubmitHandler, useForm, useWatch } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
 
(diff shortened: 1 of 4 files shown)


divbzero (approved):
LGTM. Thanks @sricharan-varanasi!

PR #2237, opened 2026-07-09
fix: No Export dropdown if audit log disabled (M2-10978)

divbzero:

### 📝 Description

🔗 [Jira Ticket M2-10978](https://mindlogger.atlassian.net/browse/M2-10978)

Changes include:

- Display export modal immediately instead of opening dropdown with a single **Responses** option.

### DIFF
--- a/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
+++ b/src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
@@ -33,9 +33,9 @@ export const HeaderOptions = () => {
   const exportButtonRef = useRef<HTMLButtonElement>(null);
 
   const handleOpenExportMenu = () => {
-    const fullAccess = checkIfFullAccess(roles);
+    const canExportAuditLogs = checkIfFullAccess(roles) && featureFlags.enableAuditLogs;
 
-    if (fullAccess) {
+    if (canExportAuditLogs) {
       setShowExportMenu(true);
     } else {
       setIsResponseDataExportOpen(true);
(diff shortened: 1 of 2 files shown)


src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
```diff
@@ -67,27 +67,20 @@ export const HeaderOptions = () => {
...
-        action: handleOpenAuditLogs,
-        title: t('dataExport.auditLogs.menuCaption'),
-        'data-testid': 'header-option-audit-logs-button',
-      });
-    }
-
-    return actions;
-  };
+  const getExportActions = () => [
+    {
```
sricharan-varanasi, line 71: minor:can we keep the flag check inside getExportActions() itself (e.g. featureFlags.enableAuditLogs && {...} filtered with .filter(Boolean)) so the item reacts to flag changes on every render, not just at the moment the menu was opened.

sricharan-varanasi (approved):
Looks good @divbzero . Approving with a minor suggestion.

src/modules/Dashboard/components/HeaderOptions/HeaderOptions.tsx
divbzero: @sricharan-varanasi I think we already check the flag when the user clicks the export button by running the check in `handleOpenExportMenu` above, so there wouldn’t be a benefit to running the check on every render.

sricharan-varanasi: Makes sense, Chris. Thank you.

PR #2238, opened 2026-07-13
fix: resolve dependabot alerts

sricharan-varanasi:

🔗 [Jira Ticket M2-10974](https://mindlogger.atlassian.net/browse/M2-10974)

### 📝 Description

Resolves the open Dependabot alerts on `develop` (20 alerts → 1). No application code changed - only `package.json` and `yarn.lock`.

Changes include:

- **vite** `7.3.2` → `7.3.6`
- **form-data** → `3.0.5` (jsdom) / `4.0.6` (axios) - scoped, since jsdom 16 needs 3.x
- **linkify-it** → `5.0.2`, **markdown-it** → `14.3.0`
- **dompurify** → `3.4.12`, **js-yaml** → `4.3.0`
- **@babel/core** → `7.29.7`, **esbuild** → `0.27.2`
- **ws** → `7.5.11`

### 🪤 Peer Testing

**Requires `yarn install`**

- **Open an encrypted applet and enter the applet password.**

    **Expected outcome:** Responses decrypt and are readable. (Vite polyfills `Buffer`/`crypto` - highest-risk area.)

- **Log in and load the dashboard.**

    **Expected outcome:** App loads, API calls hit the correct environment, icons/SVGs render.

- **Upload an image / audio / video to an activity item.**

    **Expected outcome:** Upload succeeds and the media renders back.

- **Export data / download a report.**

    **Expected outcome:** File downloads and opens without corruption.

### ✏️ Notes

- **elliptic** (low, transitive via `vite-plugin-node-polyfills`) is the one remaining alert - no upstream patch exists yet, so it needs a "no fix available" dismissal rather than a bump.


divbzero (approved):
LGTM! Thank you @sricharan-varanasi!

package.json
```diff
@@ -183,19 +183,27 @@
     "@testing-library/user-event": "14.6.1",
     "@types/react": "18.3.26",
     "@types/react-dom": "18.3.7",
+    "@babel/core": "^7.29.6",
     "@tootallnate/once": ">=3.0.1",
     "aria-query": "5.1.3",
+    "axios/form-data": "^4.0.6",
     "defu": ">=6.1.5",
+    "dompurify": "^3.4.11",
+    "esbuild": "0.27.2",
```
divbzero: Would this pin `esbuild` to this specific version? Would we want `>=0.27.2` or `^0.27.2` instead?

sricharan-varanasi: a regression was introduced in `0.27.3`, Chris. so ^0.27.2 would resolve up to `0.27.7`, which is vulnerable. 0.27.2 was the last safe version in the 0.27.x line. Since the vite bump to 7.3.6 accepts `esbuild ^0.27.0 || ^0.28.0`, I've switched it to `^0.28.1` instead .

PR #2243, opened 2026-07-21
Release 2026.07.2

divbzero:
2026.07.2


adeiji (commented):
Looks like there's some issues with the automations here as well, but other then that everything looks good.

aweiland (approved):

PR #2245, opened 2026-07-29
fix: Display correct participant answer from alert (M2-10904)

divbzero:

### 📝 Description

🔗 [Jira Ticket M2-10904](https://mindlogger.atlassian.net/browse/M2-10904)

Changes include:

- Pass `selectedDate` and `answerId` as query params when navigating from alert
- Reselect answer from the route when different answer is already selected
- Remount responses when the participant or activity changes

### 🪤 Peer Testing

Given an activity with responses from two participants:

1. Open panel to see alerts.

2. Click Participant 1 alert → **Take me to the response data**.

   **Expected outcome:** Participant 1 responses are displayed.

3. Open the right panel to see alerts.

4. Click the Participant 2 alert → **Take me to the response data**.

   **Expected outcome:** Participant 2 responses are displayed.

The correct participant answer should be displayed even if two participants answered on the same date.

### DIFF
--- a/src/shared/layouts/BaseLayout/components/TopBar/Notifications/Notification/Notification.tsx
+++ b/src/shared/layouts/BaseLayout/components/TopBar/Notifications/Notification/Notification.tsx
@@ -1,6 +1,6 @@
 import { Box } from '@mui/material';
 import { useTranslation } from 'react-i18next';
-import { generatePath, useNavigate } from 'react-router-dom';
+import { createSearchParams, generatePath, useNavigate } from 'react-router-dom';
 
 import { useAppDispatch } from 'redux/store';
 import { page } from 'resources';
@@ -44,6 +44,7 @@ export const Notification = ({
   type,
   activityId,
   answerId,
+  createdAt,
 }: NotificationProps) => {
   const { t } = useTranslation('app');
   const dispatch = useAppDispatch();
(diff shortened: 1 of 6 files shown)


src/shared/layouts/BaseLayout/components/TopBar/Notifications/Notification/Notification.tsx
```diff
@@ -74,14 +75,17 @@ export const Notification = ({
...
+    navigate({
+      pathname: generatePath(page.appletParticipantActivityDetailsDataReview, {
         appletId,
         subjectId,
         activityId,
-        answerId,
       }),
-    );
+      search: createSearchParams({
+        selectedDate: createdAt.substring(0, 10), // ISO 8601 date in UTC
```
sricharan-varanasi, line 85: Just had a question here, @divbzero 

`selectedDate` comes from the server-stamped `alert.createdAt`, while `/answers/{appletId}/dates` uses the app-provided `AnswerSchema.created_at`. For offline responses, these can differ - for example, answered Monday and synced Thursday-so the alert date may not exist in the returned list. Could we fall back to the latest available date when there’s no match?

divbzero: Thanks @sricharan-varanasi. Addressed this gap in commit 3cfdd64.

sricharan-varanasi (approved):
LGTM, Chris

PR #2246, opened 2026-07-31
feat: keep session alive on Admin ( M2 - 11001 )

sricharan-varanasi:

### 📝 Description

🔗 [Jira Ticket M2-11001](https://mindlogger.atlassian.net/browse/M2-11001)

Admin only refreshes its access token reactively today, when a request happens to fail with 401. Once the backend switches web/admin to short-lived tokens, an active user who isn't generating API calls (e.g. typing in a long Builder form) can silently lose their session. There's also no inactivity timeout at all right now.

This adds a background engine, behind a new `enableSessionKeepAlive` flag (default off):

- While the user is active, it refreshes the token shortly before it expires, so an active session never dies just because it wasn't making API calls.
- If the user goes idle (`REACT_APP_IDLE_TIMEOUT_MIN`, default 30 min), it logs them out through the existing soft-lock flow - email pre-filled, returned to their page after re-login.
- Both timers read the token's real expiry at decision time rather than assuming a fixed interval, so this stays correct regardless of what lifetime the backend issues.

With the flag off, nothing changes - no timers, no listeners, no extra requests.

**Needs the backend token-rotation work merged to test end-to-end** (short-lived, rotating tokens for web/admin). Until that lands, keep this flag off in any shared environment - a backgrounded duplicate tab can otherwise log out a tab you're actively using; that fix is a separate follow-up ticket.

### 🪤 Peer Testing

**Requires setting `enableSessionKeepAlive` to `true` locally** (LaunchDarkly default is `false`)

- Set short local values, e.g. `REACT_APP_IDLE_TIMEOUT_MIN=3`, and shorten the backend's web/admin token lifetimes.

  **Expected outcome:** staying active keeps you logged in past the token's normal lifetime - periodic refresh calls in the network tab, no 401s.

- Log in, then leave the tab untouched past the idle timeout.

  **Expected outcome:** automatic logout, landing on the login page with your email pre-filled.

- Turn the flag back off and repeat.

  **Expected outcome:** no refresh calls, no logout - identical to today's behavior.

### ✏️ Notes

Duplicate-tab sync and the warning-before-logout modal are tracked as separate follow-up tickets.


src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
```diff
@@ -0,0 +1,20 @@
...
+export const DEFAULT_IDLE_TIMEOUT_MIN = 30;
+
+// Headroom before token expiry. Overridden by REACT_APP_REFRESH_LEAD_SEC.
+export const DEFAULT_REFRESH_LEAD_SEC = 90;
+
+// mousemove fires continuously, so writes are throttled well below the timeout's precision.
+export const ACTIVITY_THROTTLE_MS = 5 * MS_IN_SEC;
+
+export const ACTIVITY_EVENTS = [
+  'mousemove',
```
divbzero, line 14: `'pointermove'` might capture more input methods than `mousemove`.

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
```diff
@@ -0,0 +1,20 @@
...
+  'pointerdown',
+  'keydown',
+  'scroll',
+  'wheel',
+  'touchstart',
```
divbzero, line 19: `'touchstart'` might be redundant, should be captured as `'pointerdown'` already.

divbzero (commented):
@sricharan-varanasi Did we rule out switching from `sessionStorage` to `localStorage`? That's the only major question I have before we move forward. The code logic related to session timeout looks good.

divbzero (approved):

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.ts
```diff
@@ -0,0 +1,81 @@
...
+      hasEnded = true;
+      logoutRef.current({ shouldSoftLock: true });
+    };
+
+    const schedule = () => {
+      if (hasEnded) return;
+      clearTimeout(refreshTimer);
+      clearTimeout(logoutTimer);
+
+      const idleDeadline = (getLastActivityAt() ?? Date.now()) + idleTimeoutMs;
```
divbzero: Are we setting the right default if `getLastActivityAt()` returns `null`? I think we should change this to something like:

```ts
const lastActivityAt = getLastActivityAt();
if (lastActivityAt === null) return endSession();
const idleDeadline = lastActivityAt + idleTimeoutMs;
```

divbzero:
Superseded by #2251.

PR #2251, opened 2026-08-12
feat: Share one session across tabs (M2-11052,M2-11002)

sricharan-varanasi:
### 📝 Description

🔗 [Jira Ticket M2-11052](https://mindlogger.atlassian.net/browse/M2-11052)
🔗 [Jira Ticket M2-11002](https://mindlogger.atlassian.net/browse/M2-11002)

#2249 moved tokens into one shared, encrypted slot. But `react-secure-storage` reads from a snapshot taken when the page loads, so a tab still can't see what another tab wrote.

That leaves three problems:

- Tab A rotates the token, tab B keeps using the old one, and the retry revokes the family for both
- Tab A logs out, tab B keeps showing a signed-in UI over a dead session
- A tab sitting on the login page doesn't notice a sign-in elsewhere

Tabs now talk over a `BroadcastChannel`.

Changes include:

- Rotated tokens, logout, and session hand-off sync across tabs
- A tab on the login page joins a session started elsewhere, or reloads into one if no other tab is left
- A frozen tab catches up on focus, or tears down if the session ended while it slept
- An idle logout no longer gets caught by the builder's unsaved-changes prompt

Changes include:

- A modal at 25 minutes idle, counting down the last 5 before logout
- Stay logged in keeps the session; Log out ends it immediately. Closing the modal keeps the session
- Activity pauses while the modal is open, so moving the mouse toward the buttons cannot answer it for you
- The countdown length is configurable via REACT_APP_IDLE_WARNING_MIN, capped at half the idle timeout
- Answering in one tab closes the modal in the others

The main bug - a background tab logging out the tab you're actually using - is fixed by one line. The idle clock is shared now, so the timer just re-reads it instead of firing blind.

### 🪤 Peer Testing

Needs `enableSessionKeepAlive` on locally, `REACT_APP_IDLE_TIMEOUT_MIN=3`, `REACT_APP_REFRESH_LEAD_SEC=20`.

Use `chrome://discards` (or `edge://discards`) → Freeze for the frozen-tab steps.

- **Two tabs. Work in tab A past the idle timeout, leave tab B in the background**

    **Expected outcome:** neither logs out

- **Two tabs. Wait for a rotation, then click something in tab B**

    **Expected outcome:** one refresh call across the browser, and tab B doesn't 401

- **Log out in tab A**

    **Expected outcome:** tab B goes to the login page straight away

- **Edit something in the builder, then idle out**

    **Expected outcome:** soft-lock login, no unsaved-changes modal

- **Edit something in the builder, then log out from the account menu**

    **Expected outcome:** the save/discard prompt still appears

- **Freeze a tab, rotate a few times elsewhere, then focus it**

    **Expected outcome:** it catches up, no 401

- **Freeze tab B, log out in tab A, focus tab B**

    **Expected outcome:** tab B tears down on focus

- **Leave tab A on the login page, sign in on tab B**

    **Expected outcome:** tab A moves to the dashboard. Close tab B first and tab A reloads once instead

- **Flag off**

    **Expected outcome:** no channel traffic, no idle logout, same as #2249 alone

### ✏️ Notes

- Base is `session-storage-local` (#2249), which sits on `session-keep-alive` (#2246). Merge in order.
- Supersedes #2247. Five of its nine commits are cherry-picked; the rest no longer apply.
- Two fixes ship unflagged: the idle-timer fix and the failed-refresh logout. Neither is about tab sync.

### Functionality

### Testing

- [x] Test suite passes

### Security & Data Privacy

- [x] New dependencies are well maintained, have significant justification for being added to the project, and are documented in the Curious open source credit page - none added

### Logging/Monitoring

- [ ] Logging is implemented for this change such that you could troubleshoot this feature in production — none added
- [ ] The change/feature is able to be monitored in production — none added

### Performance

### Readability

### Change Safety

- [x] ~~Destructive database migrations are rolled out in stages. For example, renaming a column means adding a new column and migrating the existing data to that columns in one deployment. Then monitoring to ensure that field isn't used, and finally removing that old column in a separate deployment.~~


src/shared/hooks/useSessionKeepAlive/clearStaleSession.ts
```diff
@@ -0,0 +1,18 @@
...
+export const clearStaleSession = () => {
+  const lastActivityAt = getLastActivityAt();
+  // No clock means the session predates this check. Leave it to the usual 401 path.
+  if (!lastActivityAt) return;
+
+  const { idleTimeoutMs } = resolveSessionConfig();
+  if (Date.now() - lastActivityAt < idleTimeoutMs) return;
+
+  authStorage.clear();
+  clearSessionState();
```
divbzero, line 17: [**Important**] Also clear `sessionStorage` like we do in `Auth.reducer.ts`?

```ts
sessionStorage.clear();
```

src/shared/hooks/useSessionKeepAlive/clearStaleSession.ts
```diff
@@ -0,0 +1,18 @@
...
+};
```
divbzero: Also invalidate tokens on the server like we do in `useLogout.ts`?

```ts
await deleteAccessTokenApi();
```

```ts
await deleteRefreshTokenApi();
```

src/shared/hooks/useSessionKeepAlive/clearStaleSession.ts
```diff
@@ -0,0 +1,18 @@
...
+// A session outlives its tab now, so closing the browser no longer ends one. Without this, a
+// session left idle too long comes back on the next visit and only fails once it makes a request.
+export const clearStaleSession = () => {
+  const lastActivityAt = getLastActivityAt();
+  // No clock means the session predates this check. Leave it to the usual 401 path.
+  if (!lastActivityAt) return;
+
+  const { idleTimeoutMs } = resolveSessionConfig();
+  if (Date.now() - lastActivityAt < idleTimeoutMs) return;
+
```
divbzero, line 15: [**Important**] Also invalidate tokens on the server like we do in `useLogout.ts`?

```ts
await deleteAccessTokenApi();
```

```ts
await deleteRefreshTokenApi();
```

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
```diff
@@ -0,0 +1,20 @@
...
+export const DEFAULT_IDLE_TIMEOUT_MIN = 30;
+
+// Headroom before token expiry. Overridden by REACT_APP_REFRESH_LEAD_SEC.
+export const DEFAULT_REFRESH_LEAD_SEC = 90;
+
+// mousemove fires continuously, so writes are throttled well below the timeout's precision.
+export const ACTIVITY_THROTTLE_MS = 5 * MS_IN_SEC;
+
+export const ACTIVITY_EVENTS = [
+  'mousemove',
```
divbzero, line 20: `s/mousemove/pointermove/`? `pointermove` should capture movement from mouse, touch, or pen.

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
```diff
@@ -0,0 +1,20 @@
...
+  'pointerdown',
+  'keydown',
+  'scroll',
+  'wheel',
+  'touchstart',
```
divbzero, line 25: Remove `touchstart`? `touchstart` should be captured by `pointerdown`.

src/shared/api/api.utils.ts
```diff
@@ -26,45 +28,78 @@ export const getRefreshTokenData = (config: InternalAxiosRequestConfig) => {
...
+    data: { result },
+  } = await signInRefreshTokenApi({
+    refreshToken: authStorage.getRefreshToken(),
+  });
+  const { accessToken, refreshToken } = result ?? {};
+
+  if (!accessToken || !refreshToken) {
+    throw new Error('Access token refresh failed.');
+  }
+
```
divbzero, line 42: [**Important**] Do we need a check that the user hasn’t logged out while awaiting the refreshed token?

divbzero (commented):
@sricharan-varanasi This is big set of changes and looks really good overall. Just a few comments: 3 that I think could tighten the security, and 2 `ACTIVITY_EVENTS` comments that could slightly reduce how often the event handler fires.

src/resources/app-en.json
```diff
@@ -1513,6 +1514,8 @@
   "sept": "September",
   "serverConfiguration": "Server Configuration",
   "serverStatusConfigured": "Server Status: Connected.",
+  "sessionTimeoutDescription": "You haven't done anything in this window for awhile. If you don't make a choice, we'll log you out in {{countdown}}.",
```
divbzero, line 1517: `s/awhile/a while/` — I think this was a typo in the design.

src/shared/hooks/useSessionKeepAlive/SessionTimeoutModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+      onSubmit={onStayLoggedIn}
+      title={t('sessionTimeoutTitle')}
+      buttonText={t('stayLoggedIn')}
+      hasSecondBtn
+      secondBtnText={t('logOut')}
+      onSecondBtnSubmit={onLogOut}
+      data-testid="session-timeout-modal"
+    >
+      <StyledModalWrapper>
+        {t('sessionTimeoutDescription', { countdown: formatCountdown(msRemaining) })}
```
divbzero, line 31: :+1: countdown formatting looks good

.env.example
```diff
@@ -11,4 +11,12 @@ REACT_APP_MIXPANEL_FORCE_ENABLE=false
...
 REACT_APP_DD_CLIENT_TOKEN=
 REACT_APP_DD_VERSION=local
-REACT_APP_DD_TRACING_URLS=http://localhost:3000
\ No newline at end of file
+REACT_APP_DD_TRACING_URLS=http://localhost:3000
+
+# Session keep-alive
+REACT_APP_IDLE_TIMEOUT_MIN=30
+REACT_APP_REFRESH_LEAD_SEC=90
+REACT_APP_IDLE_WARNING_MIN=5
```
divbzero, line 19: @sricharan-varanasi Will we need to set these environment variables during deployment?

divbzero (commented):
The latest 9 commits look solid :+1: just one typo from the design. Thank you @sricharan-varanasi!

src/shared/hooks/useSessionKeepAlive/clearStaleSession.ts
sricharan-varanasi: Thank you, @divbzero . Fixed in 7c31d7956 . Added `sessionStorage.clear()` after the token clear, matching the teardown in `Auth.reducer.ts`.

src/shared/hooks/useSessionKeepAlive/clearStaleSession.ts
sricharan-varanasi: Fixed in 46a02ca62 . It now calls `deleteRefreshTokenApi()` before clearing, guarded on the
refresh token's own `exp` so it only fires while the server would still accept it.

On today's config it never fires: the idle timeout and the web/admin refresh token are both
30 min, so the token has already expired by the time a session goes stale. It engages only if
those two ever stop lining up. `logout2` revokes the whole family, so the access token goes
with it and one call is enough.

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
sricharan-varanasi: Fixed in  46a6f63f6 . Switched to `pointermove`, which covers mouse, touch and pen in one.

src/shared/hooks/useSessionKeepAlive/useSessionKeepAlive.const.ts
sricharan-varanasi: Fixed in 46a6f63f6 . Dropped both `touchstart` and `mousedown` for `pointerdown`.

src/shared/api/api.utils.ts
sricharan-varanasi: Thank you @divbzero . Fixed this gap in 3713f859c . The refresh now checks the store still
holds a refresh token before saving the new pair, and throws if it doesn't.

src/shared/hooks/useSessionKeepAlive/SessionTimeoutModal.tsx
sricharan-varanasi: Thank you @divbzero

src/resources/app-en.json
sricharan-varanasi: Fixed in cbef683d1 . Corrected in both `app-en.json` .

.env.example
sricharan-varanasi: Yeah @divbzero . The code uses these as defaults, but we could config them on ECS.

src/modules/Auth/state/Auth.reducer.ts
```diff
@@ -34,6 +40,11 @@ export const reducers = {
   endSoftLock: (state: AuthSchema): void => {
     delete state.softLockData;
   },
+  // Set when another tab answers with a live session. Signing in from here is blocked while it
+  // holds, so it must outlive the banner, which the user can dismiss.
+  markSessionElsewhere: (state: AuthSchema): void => {
+    state.hasSessionElsewhere = true;
+  },
```
divbzero: @sricharan-varanasi This appears to be a one-way operation. `markSessionElsewhere()` sets `state.hasSessionElsewhere = true` but we don’t at any point set `state.hasSessionElsewhere = false`. To sign in again, a user needs to do a hard refresh of the tab.

I think that makes sense. But just want to call that out to make sure that’s what we intend?

src/shared/hooks/useSessionKeepAlive/useSessionAdoption.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const handleVisibilityChange = () => {
+      if (document.visibilityState !== 'visible') return;
+
+      publishSessionMessage({ type: 'SESSION_REQUEST' });
+      clearTimeout(fallbackTimer);
+      fallbackTimer = setTimeout(raiseBannerFromClock, SESSION_REQUEST_WINDOW_MS);
+    };
+
+    document.addEventListener('visibilitychange', handleVisibilityChange);
```
divbzero: @sricharan-varanasi We only send out the `SESSION_REQUEST` on `visibilitychange`. Should we also send it out at least once on mount? This would handle the case where the tab has gone through `leaveEndedSession()` which ends with `window.location.reload()`.

We could insert here something like:

```ts
if (document.visibilityState === 'visible') {
  publishSessionMessage({ type: 'SESSION_REQUEST' });
  clearTimeout(fallbackTimer);
  fallbackTimer = setTimeout(raiseBannerFromClock, SESSION_REQUEST_WINDOW_MS);}
}
```

divbzero (commented):
Looks good overall! Just a couple things to check. Thank you @sricharan-varanasi.

src/shared/hooks/useSessionKeepAlive/useSessionAdoption.ts
divbzero: Addressed in commit 213f5b3.

src/shared/hooks/useSessionKeepAlive/useSessionAdoption.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (hasRaised) return;
+      hasRaised = true;
+
+      dispatch(auth.actions.markSessionElsewhere());
+      dispatch(banners.actions.addBanner({ key: 'SessionElsewhereBanner' }));
+    };
+
+    // Nobody answered, so no live tab is left to hand the session over — the last one was closed.
+    // The activity clock is the only witness left that there is anything to go back to.
+    const raiseBannerFromClock = () => {
```
divbzero: @sricharan-varanasi What happens if we can fetch `lastActivityAt` from `sessionStorage` but we can no longer fetch `accessToken` or `refreshToken` from `authStorage`? I think we might fall into a situation where  we `raiseBanner()` and `markSessionElsewhere()`, but there isn’t a session we can access and the user cannot login again.

I think we need to either add a condition here:

```ts
if (!authStorage.getRefreshToken()) return;
```

Or, alternatively, clear `lastActivityAt` on load in `index.ts`.

divbzero (commented):
Hi @sricharan-varanasi just one more thing to check for this PR. Thank you!

src/modules/Auth/state/Auth.reducer.ts
sricharan-varanasi: Yes, Chris. The block has to outlive the banner since the banner can be dismissed, and a reload clears it.

src/shared/hooks/useSessionKeepAlive/useSessionAdoption.ts
sricharan-varanasi: Thank you for catching this @divbzero . Fixed in 3cede8cba

divbzero (approved):
@sricharan-varanasi I think this one is good to go too! 🤞

divbzero:
@sricharan-varanasi The most recent commit looks good too ✅ thank you.

PR #2256, opened 2026-08-21
feat: Session timeout modal (M2-11002)

sricharan-varanasi:
### 📝 Description

🔗 [Jira Ticket M2-11002](https://mindlogger.atlassian.net/browse/M2-11002)

Warns the user before an idle logout instead of ending the session with no notice.

Changes include:

- A modal at 25 minutes idle, counting down the last 5 before logout
- **Stay logged in** keeps the session; **Log out** ends it immediately. Closing the modal keeps the session
- Activity pauses while the modal is open, so moving the mouse toward the buttons cannot answer it for you
- The countdown length is configurable via `REACT_APP_IDLE_WARNING_MIN`, capped at half the idle timeout
- Answering in one tab closes the modal in the others

### 📸 Screenshots

|                                        |                                       |

### 🪤 Peer Testing

Use **PR #2257** - it shortens the timings to 3 min idle / 1 min warning so this is observable. On this PR the modal only appears after 25 real minutes.

- **Sign in, then leave the machine completely alone for 2 minutes.**

    **Expected outcome:** the modal appears, counting down from 1:00.

- **Move the mouse without clicking anything.**

    **Expected outcome:** the countdown keeps running. Activity does not dismiss it.

- **Click Stay logged in.**

    **Expected outcome:** the modal closes and you stay signed in past the original 3-minute deadline.

- **Trigger it again, then click Log out.**

    **Expected outcome:** logged out immediately, landing on a plain login page.

- **Trigger it again and let it run to 0:00.**

    **Expected outcome:** logged out with the soft lock - email pre-filled and a banner explaining why.

- **Open two tabs, let both show the modal, click Stay logged in in one.**

    **Expected outcome:** the other tab's modal closes within a second.

### ✏️ Notes

- Base is `session-flag-removal` (#2255), which sits on `session-sync-local` (#2251). Both need to merge first.
- The buttons follow the repo's standard modal pairing (filled primary, text secondary) rather than the two text buttons in Figma, and the title uses Moderat like every other admin modal rather than the Affix in the design. Raised with design.

#### Functionality

#### Testing

- [x] Test suite passes

#### Security & Data Privacy

- [x] New dependencies are well maintained, have significant justification for being added to the project, and are documented in the Curious [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

#### Logging/Monitoring

- [ ] Logging is implemented for this change such that you could troubleshoot this feature in production
- [ ] The change/feature is able to be monitored in production

#### Performance

#### Readability

#### Change Safety

- [x] ~~Destructive database migrations are rolled out in stages. For example, renaming a column means adding a new column and migrating the existing data to that columns in one deployment. Then monitoring to ensure that field isn’t used, and finally removing that old column in a separate deployment.~~


divbzero:
Changes pushed to #2251.

